PR #254, opened 2015-01-14
2.11.5

adriaanm:

PR #255, opened 2015-01-14
Remove trailing comma, appease Maruku?

adriaanm:


adriaanm:
Follow up to #254

adriaanm:
No, that wasn't it -- my local version of maruku renders this all just fine.

PR #262, opened 2015-02-02
Fix the Forking url and grammatical consistency

JustinPihony:
Review by @dickwall


heathermiller:
Thanks for the fix-ups!

PR #265, opened 2015-02-04
Switched out logos, tweaked margin-top for logo on homepage

rasummer:
@heathermiller I've switched out the updated logo. I don't think all of these logos are being used in the site, but I made replacements regardless. Also because the font is thiner on the newer logo, I made the logos used on the homepage and secondary pages a little taller to compensate. Screenshots attached.

screenshot-localhost 4000 2015-02-04 12-31-28

screenshot-localhost 4000 2015-02-04 12-31-12

screenshot-localhost 4000 2015-02-04 12-31-00


rasummer:
@heathermiller I'll move onto the docs site next. Are there any other sites that require logo updates?

heathermiller:
Thanks @rasummer!

PR #269, opened 2015-02-11
Upcoming events from Underscore

noelwelsh:
This adds our upcoming events.

BTW, I noticed that http://scala-lang.org/training/ is not rendering correctly. All the links there are currently broken, pointing to http://scala-lang.org/training/test

Rendering worked fine on my local fork of the repo.

### DIFF
--- a/training/_posts/2015-02-24-essential-scala-online.html
+++ b/training/_posts/2015-02-24-essential-scala-online.html
@@ -1,0 +1,9 @@
+---
+title: Essential Scala Online
+description: A studio-style online course for professional developers who want a thorough introduction to Scala.
+link-out: http://underscore.io/events/2015-02-24-essential-scala-online.html
+where: Online
+when: 24 February 2015
+trainers: Noel Welsh and Dave Gurnell
+organizer: Underscore
+---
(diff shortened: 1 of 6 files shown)


heathermiller:
LGTM!

heathermiller:
Thanks. And yes, we know about the links going to the wrong place – that's a problem on Typesafe's end. @rasummer, do you know yet who's responsible for the Typesafe training feeds?

noelwelsh:
Thanks Heather.

rasummer:
@heathermiller sorry this slipped through the cracks on my end. 

You're pulling from this JSON feed correct: http://www.typesafe.com/how/training.json

If so it looks like this feed does not list the URL for the details of each training, which I assume you want so that you can send visitors off to these pages.

heathermiller:
No worries! Yes, that's correct. I just don't know who's the one to ping to get the URLs back. (No idea who's responsible for the Typesafe feeds)

rasummer:
Not sure who owns this now, working on that.

rasummer:
@heathermiller 

The URL for each training session has been added to the feed. http://typesafe.com/subscription/training.json

Please also note that the feed was also redirecting from http://typesafe.com/how/training.json  to http://typesafe.com/subscription/training.json , so probably good to update this in the code. 

Although the URL's are now in the feed, it doesn't look like it's fixed the issue completely. We needed to add the URLs into the 'sessions' node of the feed. 

Hope this get's you what you need.

Rich

rasummer:
Also I noticed that some of the trainers show up as 'null'. I'll find out why we aren't entering trainers for these particular sessions (BoldRadius), likely because we don't know whom will be instructing yet.

PR #273, opened 2015-02-15
Corrected ScalaDays events (was listed under training)

andy1138:
The ScalaDays events were listed under 'training' 
Also removed 2 training courses that were listed under events (one with incorrect format so no showing on site)


heathermiller:
Thanks for the fix! (This has been driving me a bit nuts! I noticed it but never had a minute to shuffle things around) Thanks again!

andy1138:
Me also, I just broke into rage first :-)

PR #277, opened 2015-02-18
Add IDEs and Build Tools page

dickwall:
* IDEs and Build Tools page consists of links out to important scala projects
* Restored link to above page, and renamed to IDEs and Build Tools
* Added a missing link into the documentation.md page

So - this pull request is certain to be wrong! Why did I submit it then? Because it's a lot easier to argue about the way something is when you have a strawman, and this is that strawman.

I have no illusions that before the end, I will drop this PR, squash up a new one with all the incorporated changes, and submit that instead. That said, let's get going.

After talking with Adriaan yesterday, missing out Scalariform (and scalap) because they are not very active is the opposite of the point - we really want to get the battle cry out for these projects, so I put them in a "Needs particular help" section. 

Will the page be a maintenance drag? Almost certainly, but that doesn't mean it shouldn't be there. Following through to the logical conclusion of no maintenance drags would result in an empty site. The real question is, is the potential benefit of the material greater than the cost of updating it? These projects should be reasonably stable (i.e. unlikely to disappear altogether) although I expect the links to change. There is a note at the top reminding people they can scratch their own itch and we just have to decide if it's worth it to keep it up to date. I don't mind if the answer is no, but I would argue against that. Ultimately let's make the decision of whether to include this page, and what should change on it if we do.

Also, formatting - layout and design are not my forte (or even in my skills set) so while I don't think it looks horrible, there are a few bits that don't line up and if we change the look and feel completely I don't mind.

Images are links to external sites rather than pngs uploaded to our site. I suspect I am breaking rules with this, but the idea is to allow the logos (taken from github avatars) to be updated by the project owners without having to make changes to scala-lang. Exception is scalariform which has a picture of Matt (and I didn't know if he would want that) so I made it blank.

Review by @heathermiller @adriaanm

### DIFF
--- a/contribute/documentation.md
+++ b/contribute/documentation.md
@@ -60,5 +60,5 @@ to Scala and related projects) is provided on the main
 [scala-lang github project](https://github.com/scala/scala-lang) which may be forked to create pull requests. 
 
 Please read both the
-[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the scala-lang.org github README file before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
+[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the [scala-lang.org github README](https://github.com/dickwall/scala-lang#scala-langorg) before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
 
--- a/contribute/documentation.md
+++ b/contribute/documentation.md
@@ -60,5 +60,5 @@ to Scala and related projects) is provided on the main
 [scala-lang github project](https://github.com/scala/scala-lang) which may be forked to create pull requests. 
 
 Please read both the
-[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the [scala-lang.org github README](https://github.com/dickwall/scala-lang#scala-langorg) before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
+[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the [scala-lang.org github README](https://github.com/scala/scala-lang#scala-langorg) before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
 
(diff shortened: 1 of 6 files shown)


contribute/documentation.md
```diff
@@ -60,5 +60,5 @@ to Scala and related projects) is provided on the main
 [scala-lang github project](https://github.com/scala/scala-lang) which may be forked to create pull requests. 
 
 Please read both the
-[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the scala-lang.org github README file before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
+[docs.scala-lang.org contribution](http://docs.scala-lang.org/contribute.html) document and the [scala-lang.org github README](https://github.com/dickwall/scala-lang#scala-langorg) before embarking on any changes to the Scala language site, as it uses the same Jekyll markdown tool and many of the same conventions as the Scala documentation site.
```
adriaanm: should be scala/scala-lang, not dickwall/scala-lang

adriaanm:
LGTM except for the link typo

jsuereth:
LGTM

contribute/tools.md
```diff
@@ -1,8 +1,123 @@
...
+<a href="https://github.com/scoverage/scalac-scoverage-plugin/issues">Issues</a> | 
+<a href="https://github.com/scoverage/scalac-scoverage-plugin/blob/master/README.md">ReadMe</a> | 
+<a href="https://groups.google.com/forum/#!forum/scala-code-coverage-tool">Contributing</a></p>
+</div>
+</div>
+</div>
+
+### Projects in Particular Need
+
+The following projects are important to the Scala community but are particularly in need of contributors to continue their development.
```
dwijnand: Perhaps, as such, they could appear before the above list?

contribute/documentation.md
dickwall: Nice catch. Corrected.

contribute/tools.md
dickwall: A compromise - I feel the current order reflects (or ought to) the importance to the scala ecosystem of these projects right now (although, maybe weigh in with your opinions on that), but instead I will put a link up at the top pointing down the page to the projects in particular need. It's a halfway solution, but I think it might work. I also shuffled the order of the projects to be sbt, scaladoc, scala IDE, dbuild, partest, scoverage and abide in that order, if anyone feels a different priority is warranted, let me know (but the icons also kind of work like that too).

dwijnand: I like it.

dickwall:
OK - going to run this one by folks at the eng meeting this week, along with the larger SLIP proposal (which I imagine will generate more discussion) but assuming no further comments it will probably go up next week.

dickwall:
Reviewed by Martin, Adriaan, Jason, Greg, Lukas, Josh et al.

PR #278, opened 2015-02-19
Add Chris' GSOC 2015 project suggestions

cvogt:


cvogt:
@heathermiller @sjrd @VladUreche 

I put 4 projects up and would mentor one of them if we find a qualified student. Potentially @szeiger may want to mentor the Slick one.

Please merge at will.

cvogt:
@sjrd: That's what @jsuereth said. He is equally skeptical. And that's ok. I may be totally wrong about the feasibility and the project may face a dead end. The fact that you and josh both know sbt much better than I do actually points to that being the case actually. But I have some ideas, that I would love to see prototyped and potentially taken to success. In my experience most industry Scala developers I met simply don't understand sbt. And it seems to me, that the largest reason for that is sbt's frameworky nature. Again, I may be wrong, but I think it is worth more investigation. Best case scenario would be a better user interface on top of sbt. Not a competitor.

PR #291, opened 2015-03-26
2015-03-26-SpringPeople.md

rakeshsingh9901:
This two-day course is designed to give experienced developers proficient in Java, C++, C# or Ruby the know-how to confidently start programming in Scala. The course ensures you will have a solid understanding of the fundamentals of the language, the tooling and the development process as well as a good appreciation of the more advanced features. If you already have Scala programming experience, then this course could be a useful refresher, yet no previous knowledge of Scala is assumed.


sjrd:
The file name should match the start date.

heathermiller:
Also, this is also a course, not an event. So please move this over to the training feed instead.

heathermiller:
(The training feed is in directory `training` rather than `events`)

dickwall:
Assuming this is April 13th and 14th 2015 which is past - closing the PR

PR #292, opened 2015-03-26
2015-03-27-SpringPeople.md

rakeshsingh9901:
If you already have programming experience with Scala and want to understand its advanced features, this two-day course is for you. It is designed to help developers fully understand topics such as advanced object-functional programming, the power of Scala's type system, implicits, etc. The course also covers how to leverage these features to create well-designed libraries or DSL’s utilizing proven best practices


sjrd:
The file name should match the start date.

rakeshsingh9901:
Thanks for helping.
Done

heathermiller:
This is also a course, not an event. So please move this over to the training feed instead.

heathermiller:
(The training feed is in directory `training` rather than `events`)

dickwall:
One is still in events, the other has expired - closing PR.

PR #294, opened 2015-03-27
2015-04-20-Scala.md

rakeshsingh9901:
Advanced Scala Training By SpringPeople


heathermiller:
Thanks, LGTM :)

heathermiller:
Also note, that in general, we prefer to have the date and title of the course in the file name. So, something like 2015-04-13-fast-track-to-scala.md would have been better. It helps us keep all of our training stubs organized. So, next time, please include the correct title in the file name.

PR #296, opened 2015-04-17
fix session urls for the training feed

### LINKED ISSUE #260: Training feed broken
Events appear in the feed, but when you click on one, you are brought to the URL: http://scala-lang.org/test which returns "Not found."

gourlaysama:
In the JSON, the url is in "session", not "training", hence the
"undefined" result.

Fixes #260.


heathermiller:
LGTM! Thanks!

PR #303, opened 2015-06-23
2.11.7 released

SethTisue:
not ready for merge yet. missing pieces:

* update download sizes (in index.md and the new file in _posts) after we build bundles
* add release notes to the new news post, once they're done

PR #304, opened 2015-06-24
2.11.7 release notes

adriaanm:
tweak of #303


adriaanm:
Ready now for merge to do soft launch, as far as I'm concerned. /cc @SethTisue

heathermiller:
Hi guys, I just noticed this now, but the front page of scala-lang is broken. I'm assuming it's from some rendering error in the release notes. Haven't had a chance to look at it, noticed it while getting ready for bed. If someone gets a second, would you mind debugging? Otherwise I can fix in the am.

adriaanm:
I'll take a look. Probably need to move up the marker for the summary.

adriaanm:
Fixed by #307

heathermiller:
Awesome, thanks a lot Adriaan!

PR #305, opened 2015-06-24
make wording more balanced,

adriaanm:
upped the embeddability factor accordingly


adriaanm:
fast tracking to erase traces of unbalanced attribution

PR #306, opened 2015-06-25
Unbreak front page by limiting blurb?

adriaanm:


adriaanm:
Breakage spotted by @heathermiller

adriaanm:
Fasttracking in hopes this restores the front page.

adriaanm:
Nope -- looks like it was the comment in the 2.12.0-M1 news.

PR #309, opened 2015-06-29
Updated the leman lake name

joan38:
The lake next to Lausanne is called the Leman lake not Geneva lake.


heathermiller:
In English, it's referred to as Lake Geneva. https://en.wikipedia.org/wiki/Lake_Geneva

PR #316, opened 2015-07-15
Replaced Scala Tribes with Scala Space

propensive:
Scala Tribes is no longer maintained, and the data is out of date.


SethTisue:
outstanding

PR #321, opened 2015-07-29
Add an RSS feed for the blog

jeantil:
creates the feed/blog.xml file
adds the link to the feed as an alternate link in the header for the blog layout
which will let rss readers autodetect the feed from the blog url

### DIFF
--- a/_layouts/blog.html
+++ b/_layouts/blog.html
@@ -4,6 +4,7 @@ toc: true
 ---
 
 {% include headertop.html %}
+<link href="/feed/blog.xml" rel="alternate" type="application/rss+xml" title="Scala-lang's blog" />
 {% include headerbottom.html %}
 
 <div class="darkstrip"></div>
--- a/feed/blog.xml
+++ b/feed/blog.xml
@@ -1,0 +1,25 @@
+---
+layout: nil
+---
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom">
+  <title type="text" xml:lang="en">{{ site.root_desc }}</title>
+  <link type="application/atom+xml" href="http://www.scala-lang.org/feed/blog.xml" rel="self"/>
+  <link type="text" href="http://www.scala-lang.org/" rel="alternate"/>
+  <updated>{{ site.time | date_to_xmlschema }}</updated>
+  <id>http://www.scala-lang.org/</id>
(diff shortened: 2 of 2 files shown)


feed/blog.xml
```diff
@@ -0,0 +1,25 @@
+---
+layout: nil
+---
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom">
+  <title type="text" xml:lang="en">{{ site.root_desc }}</title>
+  <link type="application/atom+xml" href="http://www.scala-lang.org/feed/index.xml" rel="self"/>
```
sjrd: This should reference `feed/blog.xml`

sjrd:
That's all. I assumed you tested this locally?

jeantil:
I just updated the commit to fix the alternate link on the blog.xml file. 

I did start jekyll locally and checked that the /feed/blog.xml url works. 
capture d ecran 2015-07-29 a 11 49 36

I am not sure how to test the link for rss autodiscovery in the header of the layout (but I confirmed it did not break the layout)

jeantil:
ok my github foo seems to be a bit rusty I force pushed the new commit but it isn't picked up ...

jeantil:
ah there it is, the updated commit fixes the link in feed/blog.xml to reference itself instead of feed/index.xml

_layouts/blog.html
```diff
@@ -4,6 +4,7 @@
 ---
 
 {% include headertop.html %}
+<link href="/feed/blog.xml" rel="alternate" type="application/rss+xml" title="Scala-lang's blog" />
```
jeantil: Should I use a fully qualified URL here ?

sjrd:
LGTM

SethTisue:
awesome, thank you!

jeantil:
thanks !

jeantil:
the auto detect for the feed from the index page works in feedly 
capture d ecran 2015-07-29 a 14 20 14

dickwall:
Ha - I got up early to write this before I started work today and you beat me to it - thanks - love this community spirit :+1:

jeantil:
@sjrd 's suggestion seemed easy enough to implement that I could do it. 
Sorry you got up early for this, you get to enjoy an extended breakfast instead of coding an rss feed :)

PR #324, opened 2015-08-30
Scala 2.9.3 not compatible with "1.6 or later"

refried:
right?

### DIFF
--- a/download/_posts/2007-05-21-2.5.0.final.md
+++ b/download/_posts/2007-05-21-2.5.0.final.md
@@ -6,7 +6,7 @@ release_version: 2.5.0.final
 release_date: "May 21, 2007"
 show_resources: "true"
 permalink: /download/2.5.0.final.html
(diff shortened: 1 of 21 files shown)


SethTisue:
Right, and thanks for pointing out that the download pages need updating.

Before we could merge this, I think we'd need at least these two changes:

* Update all affected download pages (everything 2.9 and before), not just 2.9.3 in particular. It looks like the wording of this sentence didn't change from release to release, so it should be an easy search and replace.
* It isn't true that Java 6 and 7 are no longer available for public download. They remain available from http://www.oracle.com/technetwork/java/archive-139210.html. So I would suggest simply omitting the entire "which can be downloaded..." clause.

refried:
@SethTisue My bad — I was reading at http://www.java.com/en/download/faq/java_7.xml and saw "<b>updates</b> [emphasis added] are not no longer available to the public" and thought it meant the other thing.

SethTisue:
LGTM. Thank you!

PR #330, opened 2015-09-15
Added Lambda World event

nickelsberry:
Added Lambda World to the events list


heathermiller:
Wow, looks like a cool event guys! LGTM

PR #333, opened 2015-09-30
SI-9495 Add note about configuring Ant for HTTP proxies

retronym:
Review by @SethTisue


SethTisue:
LGTM. I will make a separate pull request in the scala repo making the same change in README.md. (The hacker's guide is awesome, but there is some unfortunate redundancy with the Scala README).

PR #334, opened 2015-10-05
Adds initial GSOC 2016 page

cvogt:
review by @sjrd or @VladUreche 

Wanted to add the Slick codegen project already

### DIFF
--- a/gsoc/2016.md
+++ b/gsoc/2016.md
@@ -1,0 +1,145 @@
+---
+layout: page
+title: "Google Summer of Code 2016 Scala Projects"
+---
+
+*This page is work in progress for the upcoming GSOC 2016*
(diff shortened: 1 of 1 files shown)


VladUreche:
LGTM. Since I never merged into scala-lang, I'll let @heathermiller do it.

SethTisue:
where will this be linked from?

cvogt:
@SethTisue it was usually linked only from news entries and from the mailing list and twitter, etc if I am not mistaken

cvogt:
in the old website there was a section for it, but not anymore. maybe should be under community?

SethTisue:
yes, I think that would be appropriate

cvogt:
Different target than this PR though :). Let's get this merged first.

PR #336, opened 2015-10-06
SI-9501 add prominent links to relevant docs in main Scala repo

SethTisue:
the text of the guide already links to the same documents at
appropriate places in the narrative, but I also think it's worth
having the same links right up front


SethTisue:
review by @dickwall and/or @heathermiller

it would be good to do something to reduce the redundancy between the three documents (hacker guide, readme, contributor guidelines) but in the meantime, I think this and https://github.com/scala/scala/pull/4790 will improve the chances of people encountering all three

heathermiller:
LGTM :)

PR #338, opened 2015-10-07
Gitter room reorg

SethTisue:


SethTisue:
merging right away so the site has up-to-date info, but I will respond promptly to any feedback

PR #343, opened 2015-10-27
simplyscala.com is offline

napcae:
Therefore it should be removed from this section.
Up for discussion.

[This](https://www.reddit.com/r/scala/comments/3didyu/where_is_simplyscala_gone/) reddit thread indicates, that it is at least offline since 3 months.

Also take a look at [archive.org](https://web.archive.org/web/*/http://www.simplyscala.com/).


sjrd:
LGTM, sadly ...

heathermiller:
I was just looking at this... Sad, but Simply Scala was quite useful. Should we offer one of those replacements at least?

napcae:
To recommend something, you should at least know them. Since I don't know any of these, I can't really do that. But I agree.

heathermiller:
Well I don't know any either, but recommending these three is likely better than doing nothing. I'll take care of it.

PR #345, opened 2015-10-28
Updated ScalaCourses.com entry

mslinn:


heathermiller:
Thanks! – updated PR in #345

mslinn:
Heather, the linked web site was updated, but the link text was not. Would you please modify "Independant Courseware" to read "ScalaCourses.com"?

Thanks, Mike

heathermiller:
Hi Mike, I believe we already discussed this a year or so ago in some other PR. I kept the link text I originally used so as to make a clear distinction that your courseware is independent from EPFL and any other related entity.

mslinn:
How about "Independent Courseware from ScalaCourses.com"? I believe that should address both of our concerns.

mslinn:
Heather,

It is important to provide the name of the product/service in the link. Last year I had also suggested "Independent Courseware from ScalaCourses.com" but there was no response. I believe this suggestion should be acceptable. If not, then there must be some other unexpressed reason. Every other product / service contains the word "Scala" in its name, so that cannot be the reason. What could the reason be? It is important that our listing state the name ("ScalaCourses.com"), and that would only be fair. Here are two suggestions:

ScalaCourses.com (Independant Courseware)
Independent Courseware from ScalaCourses.com

Thank you,
Mike

PR #360, opened 2015-11-23
link to both 2.11 & 2.12 nightly API doc, not just 2.11

SethTisue:
as suggested by (iirc) @som-snytt

review @heathermiller


som-snytt:
Thanks, I owe you one. One what, I don't know.

I was actually thinking of the front page link, under the blinking red Scala detonation, but this location is also helpful.

SethTisue:
oh whoops, I didn't see that link. I will follow up with an additional commit

SethTisue:
done, but you might not like it

som-snytt:
I like it. You might go further, since the big green API DOCS button is redundant with current link.

Left: DOWNLOAD, Getting started, All Releases.
Right: API DOCS, Language Specification, All Versions.

"What you want", "More info" and "all the things."

SethTisue:
> You might go further

another time :-)

PR #362, opened 2015-11-25
Fix incorrect URL in GSOC pages

bhargav:
Incorrect hyperlink to 2014 and 2015 GSOC pages. Browser tries
to open it relative to current page.


bhargav:
cc @cvogt

heathermiller:
LGTM, thanks!

PR #368, opened 2015-12-10
Minor changes to the getting-started page

stew:
- don't use null
- don't use procedure syntax


sjrd:
Sounds good to me, thanks. Any objection?

SethTisue:
thanks Stew!

PR #375, opened 2015-12-17
restore instructions for running with global Jekyll

SethTisue:
at @heathermiller's request


heathermiller:
Can you reverse the order of the instructions? Global first, and then bundler? Essentially everyone who is currently building the site is using global Jekyll, so it makes sense to list that first to avoid confusion for people who have already adopted this workflow

dwijnand:
Personally I think we should promote and encourage not installing things globally, and using the versioned Gemfile, etc.

Still restore the instructions, but in the order they are presented here currently.

SethTisue:
I agree with Dale. If there are a few people who still want to do it the old way, fine. The Bundler way is the correct direction to steer new contributors.

SethTisue:
> Essentially everyone who is currently building the site

I'm nobody?

heathermiller:
I disagree with both of you. 63 out of 65 (or whatever) contributors who have contributed to the site already have Jekyll installed. They reach for the README to remember "jekyll serve". 

They do not want the distraction of figuring out what bundler is etc.

Further, I work in an office where people come into my office regularly to complain about having to use something other than Scala/sbt to build websites. Using Jekyll system-wide is how we do it across all projects here. Confusing several people and forcing them to change their workflow when there's no measurable benefit for them is unnecessary.

I'm not willing to receive the heat from my co-workers for your personal preference of not wanting to install Jekyll system-wide. It makes my daily working life harder.

You have gemfiles etc now, you can have enabled peyote the option of using bundler. Please leave the instructions and the build as they were for the people who were working on the site before you arrived.

SethTisue:
> for your personal preference of not wanting to install Jekyll system-wide

please understand and grant that my personal preference is not what is driving this.

SethTisue:
I have a pushed an additional commit that restores the original order.

dwijnand:
I don't think it was the intent to force anything upon your colleagues.

I also think that if anyone were to change there would be a benefit for them (much like there is for python, etc).

Sounds like you work with a bunch of regularly complaining, easily confused bunch of colleagues.. :(

heathermiller:
> please understand and grant that my personal preferences are not what is driving this.

I can agree that the concept of allowing people to non-globally install and run with the same environment as on the production server is a good idea. 

What I can't agree with is the idea of forcing this upon people who don't need it forced upon them. If their workflow is working fine, let them be. When they start having build problems, let them discover the wonders of bundler.

heathermiller:
Thanks :)

SethTisue:
LGTM — I'm fine w/ how this ended up.

PR #378, opened 2015-12-17
add a Travis build

SethTisue:
this should be useful for lightly sanity-checking pull requests --
at least we'll know if it stops the site from generating


SethTisue:
(eventually it might be nice to add things like running html-proofer, but this is a start)

heathermiller:
:clap: :+1: LGTM! Thank you!

dwijnand:
nice :)

PR #380, opened 2015-12-17
misc small tweaks

SethTisue:
review by @heathermiller

(the changes are unrelated to each other, but are in separate commits,
at least, to facilitate review)


_includes/footer.html
```diff
@@ -16,21 +16,22 @@
...
+      <li><a href="{{ site.baseurl }}/community/index.html#community-libraries-and-tools">Libraries and Tools</a></li>
     </ul>
     <ul>
       <li><h5>Contribute</h5></li>
-      <li><a href="{{ site.baseurl }}/contribute/bug-reporting-guide.html">Report an Issue</a></li>
       <li><a href="{{ site.baseurl }}/contribute">How to Help</a></li>
-      <li><a href="{{ site.baseurl }}/contribute/hacker-guide.html">Contributor's Guide</a></li>
-      <li><a href="{{ site.baseurl }}/contribute/scala-fame.html">Contributor's Hall of Fame</a></li>
+      <li><a href="{{ site.baseurl }}/contribute/bug-reporting-guide.html">Report an Issue</a></li>
+      <li><a href="{{ site.baseurl }}/contribute/hacker-guide.html">Hacker's Guide</a></li>
```
heathermiller: Actually, I think this renaming came up before which is why in the links it's now called "Contributor's Guide". For people who haven't been to the github repo, they don't know that hacker = contributor, and maybe they would think that the guide is for people who are hardcore Scala hackers and not people who are interested in contributing to the stdlib/compiler. I think someone once made this argument to us which is why we called it "Contributor's guide" in the links here... So maybe it's best that we keep it?

SethTisue: well, I don't have a real strong opinion on this. I just thought it was a little weird that it was called "hacker's guide" everywhere but in one place.

hmm, how about sidestepping the issue by simply removing the link from the footer? I think we want new people to come in through the main Contribute page (aka "How to Help"), rather than through the Hacker's Guide.

heathermiller: Ok

SethTisue: done

heathermiller:
Thanks! LGTM

PR #383, opened 2015-12-18
add four 2016 events

SethTisue:
* Northeast Scala Symposium
* ScalaUA
* Typelevel Summit US
* Typelevel Summit Europe

ScalaUA and Typelevel logos are crappy quick (and not correctly proportioned) efforts on my part, nicer looking replacements would be welcome:

* @milessabin @larsrh you don't have a better Typelevel logo to use do you? if so, please feel free to submit a followup up with ([properly sized](https://github.com/scala/scala-lang/blob/master/events/README.md)) image files
* @rssh ditto on ScalaUA


SethTisue:
merging without review, but I will respond promptly to any feedback

PR #384, opened 2015-12-18
Add Philly ETE 2016 to events

heathermiller:


heathermiller:
Since uncontentious, merging...

PR #387, opened 2016-01-04
retire the Scala Contributor Hall of Fame

SethTisue:
https://api.github.com/repos/scala/scala/stats/contributors
wasn't giving us good information to base this on, because
it only gives you the top 100 contributors by number of
all-time commits

so e.g.  Felix Mulder's great Scaladoc contribution for December
wasn't even showing up in the data, since it was just one commit.  one
of the main purposes of this page was supposed to be recognizing new
contributors, but the 100-contributor limit prevents that.

the result was embarrassingly empty pages (which we usually forgot to
even update anyway)

if people are interested in all-time stats, they can visit
https://github.com/scala/scala/graphs/contributors , and
if people want to see recent activity, they can visit
https://github.com/scala/scala/pulse


SethTisue:
review by @heathermiller, @sjrd

SethTisue:
(P.S. if somebody wanted to bring something like this back, I would suggest querying for recently merged PRs)

sjrd:
Sounds good. That means we should also retire the cron job that was supposed to update this, if it still exist? /cc @fsalvi

fsalvi:
It was already disabled...

sjrd:
LGTM, then.

PR #388, opened 2016-01-06
2016 events: add flatMap(Oslo) and ScalaSphere

SethTisue:
review by @heathermiller


SethTisue:
hmm, I guess there's no reason I shouldn't just merge. I'll respond promptly if there's any problem

PR #391, opened 2016-01-07
add lambdaconf 2016 event

jdegoes:


SethTisue:
lambdaconf.png should be 40x40, and we'll also need a lambdaconf@2x.png at 80x80, too. LGTM otherwise

jdegoes:
@SethTisue Done!

SethTisue:
heh, I missed this the first time around, but both the filenames and the content should have 2016 not 2015, and you're working off an 11-month old clone of the repo

SethTisue:
I wrote "2015" on a check yesterday :-)

jdegoes:
Arg, those damn year changes! Can't we all just agree to stay in 2015 forever!!!

SethTisue:
thanks!

PR #395, opened 2016-01-16
clarify structure of the code of conduct

SethTisue:
after proposing the adoption of the CoC on the #scala IRC channel,
I got some feedback from a few people that the "Be excellent"
part was... a little weird.

I've addressed that as follows in two ways:

1) Replacing the "be excellent" wording (because it sounds too much
   like Bill and Ted)

2) Separating the four points of the code into "rules" and
   "guidelines", so it's clearer which parts of the code are about
   what is actually allowed

review by @heathermiller @propensive @dickwall @jsuereth @copumpkin

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -9,16 +9,18 @@ forum, mailing list, wiki, web site, Gitter or IRC room, hackathon, public
 meeting or private correspondence.  (See our
 [list of community fora](community/).)
 
-Scala moderators are appointed by EPFL / Typesafe to maintain the
+Scala moderators are appointed by EPFL and Typesafe to maintain the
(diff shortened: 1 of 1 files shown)


SethTisue:
"be excellent" is also now associated with a philosophy that no CoC is needed other than to say "Be excellent to each other". I don't think that association existed when this code was originally drafted, but it does now. see e.g. https://plus.google.com/+HannahGrimm/posts/5517czLTefa

SethTisue:
@heathermiller ping

heathermiller:
Thanks for the ping. I completely missed this. LGTM! Thanks!

PR #398, opened 2016-02-04
Promoting blog to frontpage of scala-lang

heathermiller:
Blog articles are now on the front page of scala-lang:
screen shot 2016-02-04 at 7 18 37 pm

The Liquid templating language only lets you do horrible, horrible things if you want to filter out elements of an array. Horrible. (Necessary because `site.posts` contains events as well as announcements and blogs.)

Fun fact... "Filtering" on an array in the land of Liquid/Jekyll is what we call `map` in Scala.


heathermiller:
Merging now, so Martin's new blog article is more easily found by people arriving on scala-lang due to his announce. If there are any issues anyone has with this, I can address in a follow up PR.

heathermiller:
Nevermind–there's already an issue!

SethTisue:
what was the issue?

heathermiller:
A stupid mistake that I made that somehow didn't get picked up in a local re-generation of the site before committing. Jekyll looked lit it rebuilt on my local machine but hadn't really. So I had an outdated generation of what I actually pushed. 

The solution was in the follow-up PR (slightly wrong label):
https://github.com/scala/scala-lang/pull/400/files

SethTisue:
all's well that ends well. glad to see this change in.

PR #411, opened 2016-02-17
Included Scala.IO 2016 in the list of Scala events

jeantil:

PR #412, opened 2016-02-17
Fix typos raised by SethTisue in #411

jeantil:


heathermiller:
Thanks. Little secret though, we actually don't use this `description` field anymore :/ So I should honestly just remove it, it's not required. But good to correct anyway!

PR #419, opened 2016-03-10
add two books from Packt Publishing

SethTisue:
review by @heathermiller


heathermiller:
LGTM :) Thanks @SethTisue!

PR #421, opened 2016-03-17
Fix formatting on "Getting into Scaladoc" blogpost

felixmulder:
The old one used GitHub flavoured markdown - which didn't sit well with the markdown parser on scala-lang.

review: @heathermiller / @soc


heathermiller:
Thanks! LGTM

PR #429, opened 2016-04-29
Prioritizes bundler build instructions in README

heathermiller:
It's gotten to the point where everyone (except Martin) is building with bundler :) So, bundler should be promoted over making people install a very old version of Jekyll...

Hopefully Fabien will be able to update Ruby on the machine serving the site soon...

cc @SethTisue


SethTisue:
yay!! 🎉

PR #430, opened 2016-05-05
Don't show legacy atypical shebang preamble in intro shell script example

ches:
Using a conventional shell script shebang like `#!/usr/bin/env scala` has been possible for awhile now. When I see this:

```scala
#!/bin/sh
exec scala "$0" "$@"
!#
object HelloWorld extends App {
  println("Hello, world!")
}
HelloWorld.main(args)
```

I pause and think, "Gee, that's kinda weird." Anything that makes me think that way on what is quite possibly the *first* page that a potential new Scala user reads attentively should be improved, IMO.

The `scala` man page (referenced in this same section) still illustrates this legacy script form as well, but then again it also still references `sbaz`… Maybe I'll find time to contribute an update to that as well, after I figure out what works for Windows batch scripts these days.

A few other minor trifles in the diff, I hope you won't mind the noise.

It might be nice to show the shell script example as simply:

```scala
#!/usr/bin/env scala

println("Hello, world!")
```

to make it known that that convenience is possible, but perhaps that introduces too much magic needing a hand-wavy explanation that would detract from this succinct intro doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# SCALA-LANG.ORG
+# scala-lang.org
 
 This repository contains the _static_ source of [scala-lang.org](http://scala-lang.org). It does not contain the source of any content found under the [docs.scala-lang.org](http://docs.scala-lang.org) subdomain (instead, visit the [scala.github.com repo](http://github.com/scala/scala.github.com) for that source).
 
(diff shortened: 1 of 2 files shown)


SethTisue:
@ches sorry this got neglected. there are merge conflicts; could you rebase against current master? if you do that, I promise to merge this promptly. (I'm back from a long leave.)

ches:
Cool, a few of my nits got fixed already. I've rebased this.

PR #439, opened 2016-06-07
Fix scalaio filename so it is correctly picked up by the website

jeantil:
It looks like in my previous PR I used an underscore(_) instead of a dash(-) in the filename and the conference doesn't appear on scala-lang.org ...

thank you !


heathermiller:
Oops, you're right. Thanks for catching it and for the fix!

PR #440, opened 2016-06-07
Fix scalaio image link

jeantil:
well it looks like the original PR was made on a bad day, the image link was incorrect too :(


heathermiller:
All good, thanks for the fix! :)

PR #443, opened 2016-07-04
[WIP] update versions

lrytz:
Uses the `github-pages` gem, see https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/

Issues I see so far:

When running `bundle exec jekyll serve`, accessing the main site logs `[2016-07-04 12:01:48] ERROR '/resources/img/view-leman-opt2@2x.jpg' not found.`

Pages that have a `permalink` not ending in `/` or `.html` are not served under `bundle exec jekyll serve`, but chrome downloads them. Example: https://github.com/scala/scala-lang/blob/master/news/_posts/2016-06-29-release-notes-2.12.0-M5.md. If I understand the Jekyll doc correctly, that should not happen.. https://jekyllrb.com/docs/permalinks/#extensionless-permalinks

Running `bundle install` for the first time, you might hit https://github.com/sparklemotion/nokogiri/issues/1445. Running `xcode-select --install` helps.

Another Issue I see is that the follwing markdown

```
Hi
  - item
```

does not render as a bullet-list, an empty line is needed before the list. On GitHub (comments, issues, gists) this is not the case, the above would render as a bullet.

This is not related to our site in any way, it seems to be the case in general. I created a small test site here:
  - https://github.com/lrytz/site-test/blob/gh-pages/_posts/2016-07-04-welcome-to-jekyll.markdown vs.
  - https://lrytz.github.io/site-test/jekyll/update/2016/07/04/welcome-to-jekyll.html

I contacted GitHub support about this, we'll see.


lrytz:
updated description.

lrytz:
Here's an answer from GitHub support. My question was: on GitHub, the following renders as a bullet:

```
Para
  - bullet
```

But not when using kramdown in `GFM` (Github Flavored Markdown) mode, then the line continues ("Para - bullet"). Here's the answer:

> I'm sorry that didn't solve the problem. It turns out kramdown doesn't support that part of GitHub Flavour Markdown the way I'd hoped. It looks like you'll need to add a new line before those bullets to get the correct output in kramdown.

I checked on karmdown's issue tracker, it looks like they don't want to fix this: https://github.com/gettalong/kramdown/issues/336.

So it seems GitHub is using their own markdown parser and we won't be able to get full compatibility.

heathermiller:
Oh no, that's not very nice news at all, especially given the fact that GitHub pages [appears to use kramdown as the markdown parser there](https://pages.github.com/versions/), so one would think they also used it elsewhere too. I guess not.

Thanks for tracking this one down – I was really hoping it's something that would be resolved by merely updating the Gemfile :/

heathermiller:
For what it's worth, I prototype documents in [Typora](https://www.typora.io/), a nice markdown editor that renders styling as you type. When you save, it's a valid markdown file that always renders (so far, for me) in Jekyll and all over GitHub.

It also has a nice full-screen distraction-free mode. I know this isn't a fix, but maybe it helps someones' day-to-day workflow?

lrytz:
the ticket got re-opened - now we just need to write some ruby :)

soc:
@lrytz Before you waste time wondering why things like events are not displayed anymore: https://github.com/soc/scala-lang/blob/gh-pages/_config.yml#L14

SethTisue:
@lrytz should some version of this be merged?

lrytz:
it cannot, i think it would also require updates on the epfl server. i don't time to follow up on this issue now.

another issue was that kramdown was not actually implementing the same markdown as github, maybe things improved since: https://github.com/gettalong/kramdown/issues/336

fsalvi:
FYI, here are the versions on the web server:
$ jekyll -v
jekyll 3.1.6
$ kramdown -v
1.11.1

Blaisorblade:
@fsalvi that is, the webserver uses much newer versions than in our Gemfile config, and those even support incremental builds (which take 1 second rather than ~1 minute)? Or is the webserver in fact using bundler?

jarrodwb:
I suggest that the server use Bundler if it is not already and respect the Gemfile/Gemfile.lock in the repository.

Of course I also suggest updating our Gemfile after we know that it is being used by the server.

jarrodwb:
Maybe this issue should be reopend?

SethTisue:
I've opened a new ticket at https://github.com/scala/scala-lang/issues/498

fsalvi:
The website is now using bundler with jekyll v3.1.6 and kramdown v1.11.1

PR #449, opened 2016-08-07
Fix Atom feeds

alexandru:
Fixes these issues:

- the `<title>` tag was broken, as can be seen by [the W3C validator](https://validator.w3.org/feed/check.cgi?url=http%3A%2F%2Fwww.scala-lang.org%2Ffeed%2Findex.xml) and this breaks feed readers that are less tolerant, for example Mozilla Thunderbird
- the `<id>` for `blog.xml` is the same as the one in `index.xml` and this is a mistake, because then feed readers may not be able to differentiate between them
- both `<link rel="alternate">` should be present on *all* html pages, as a best practice for discovery by feed readers
- the original `<link rel="alternate">` in the HTML head where indicating `application/rss+xml` as the MIME type, but that's wrong, as it should be `application/atom+xml` (Atom != RSS)

### DIFF
--- a/_includes/headertop.html
+++ b/_includes/headertop.html
@@ -27,5 +27,5 @@
     <script type="text/javascript">try{Typekit.load();}catch(e){}</script>
 
     <!-- Atom feeds -->
-    <link rel="alternate" type="application/rss+xml" title="News Feed" href="{{ site.baseurl }}/feed/index.xml" />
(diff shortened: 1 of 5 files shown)


heathermiller:
Thanks @alexandru :) 
Looks good to me. I really appreciate it!

SethTisue:
thank you @alexandru, very much appreciated.

PR #458, opened 2016-08-17
update (and slim down) books page

SethTisue:
keeping the page up to date is too much work.
so this commit makes the criteria more stringent: general books
about Scala for a general audience only, in English only, and only if a recent
edition exists.

I removed some outdated and/or not-general-enough books, and I updated
a couple entries (the Payne/Wampler and the Chiusano/Bjarnason).

even so, I can think of at least two important books still missing:
"Atomic Scala" and "Scala Cookbook". sigh

### DIFF
--- a/documentation/books.md
+++ b/documentation/books.md
@@ -3,20 +3,13 @@ layout: page
 title: Books
 ---
 
-There are more and more books being published about Scala. Here are some
-of the titles with their availability.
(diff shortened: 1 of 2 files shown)


SethTisue:
here's the new page, up on the site: http://scala-lang.org/documentation/books.html

PR #467, opened 2016-09-08
Update prettify.css

mslinn:
Here is a quick hack. It forces all code to white, enlarges it to 10 points and removes the annoying horizontal scroll bar. I do not care about an artistically beautiful color palette, I just want to be able to read the code.

### DIFF
--- a/resources/css/prettify.css
+++ b/resources/css/prettify.css
@@ -27,3 +27,19 @@ pre.prettyprint {
 /* Specify class=linenums on a pre to get line numbering */
 ol.linenums { margin: 0 0 0 40px; } /* IE indents via margin-left */
 ol.linenums > li { color: rgba(67, 94, 101, 0.6); line-height: 20px; width: 100%; s}
+
+/* Quick hack to make the code readable; the above settings use low contrast colors and an annoying horizontal scroll bar shows */
+.prettyprint code,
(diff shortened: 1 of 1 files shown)


SethTisue:
this is in reference to http://www.scala-lang.org/blog/2016/05/06/multiversal-equality.html#comment-2663350706

SethTisue:
can someone help review this...?

SethTisue:
@mslinn your commit seems to have to disappeared from your fork of the repo? I think you may have deleted the branch the PR was submitted from?

SethTisue:
I see a couple problems here:

* you've created duplicate entries in `resources.css` instead of updating the existing ones
* all of the syntax highlighting is lost by forcing everything to white. this seems like a substantial loss to me. can we make the colors lighter without losing syntax highlighting?

mslinn:
@SethTisue As I originally stated, my goal was not to provide syntax highlighting, my goal was to simply make the text legible. It is merely a quick fix. If someone wants to provide syntax highlighting, they will probably want to take into consideration the color scheme for the rest of the site. I'm not taking that on.

SethTisue:
OK. I'm going to close with this without merging, then, but I invite the community to submit a new PR that improves legibility without losing the highlighting.

PR #468, opened 2016-09-13
2.12.0-RC1 is released

SethTisue:
but not announced yet


SethTisue:
there will be a followup PR once the release notes are finalized over at https://github.com/scala/make-release-notes/pull/176

PR #476, opened 2016-09-22
Second cource is not free any more

omakoleg:
Courses are not free any more

### DIFF
--- a/documentation/index.md
+++ b/documentation/index.md
@@ -64,7 +64,7 @@ There are a few interactive resources for trying out Scala, to get a look and fe
 
  * [Functional Programming Principles in Scala](https://www.coursera.org/course/progfun), free on Coursera. This is a course about functional programming given by Martin Odersky himself. You can access the course material and exercises by
  signing up for the course.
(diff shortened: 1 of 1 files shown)


heathermiller:
No, it is still free. Coursera has just gotten a bit greedy and is hiding the option to skip verification and to thus enroll in the course for free. It's a pity that they try so hard to hide the free option to our course materials, but I assure you that if you look around a little more you will find that it is still free.

heathermiller:
Oh, oops! Ok, wow. So yes, Coursera is still offering courses for free. My comment above is still accurate.

However, they seem to have taken down the reactive course materials! We were not running the automatic graders for the reactive course anymore, but the materials _were_ available online.

It seems someone at Coursera decided to forward people `/reactive` to `/specializations/scala` which is not accurate at all. Principles of Reactive Programming was an entirely different course, which isn't part of the specialization on Coursera.

I'll contact Coursera about this incorrect forwarding they're doing. Though just to reiterate, if you do want to take any of the courses in the specialization, you are able to for free. You just have to search around quite a bit for the option on the page to skip verification.

SethTisue:
@heathermiller is there any need to leave this ticket open?

MariosPapasofokli:
I've made a pull request regarding reactive course removal.  Since it is not available any more we can reference the new specialisation in the documentation instead.

https://github.com/scala/scala-lang/pull/487

PR #478, opened 2016-09-26
Open Enrollment, Instructor-Led Online Introduction to Scala Course

mslinn:


training/_posts/2016-10-18-scala-intro.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Introduction to Scala Online
+description: Hands-on course, scheduled over 30 days, 43 lectures, 8 hours of video, weekly office hours (videoconference), individual support
+link-out: https://www.getscala.com/#live
+where: Online, world-wide, office hours in multiple time zones
```
heathermiller: Shorten to just "online" else it will break the formatting. This field is not meant to wrap.

mslinn: Done

heathermiller:
Cool, thanks! LGTM :)

mslinn:
The training event is visible but the link is broken: http://scala-lang.org/undefined

andy1138:
@mslinn have you seen the scala-lang homepage. Main news headline is 'Introduction to Scala Online' , with a working like so not all bad :-)  

I added a comment to the merge request, but I couldnt see whats wrong

mslinn:
Yes, I am reporting that the scala-lang home page shows the event, but the link is broken. The link I supplied was:

    link-out: https://www.getscala.com/#live

 Perhaps whatever parses the link threw an exception when it encountered the # character, hence the `undefined`.

mslinn:
Ouch!

mslinn:
I figured out a way to work within AWS redirection limitations. Instead of a # the URL now has a ?. Please reinstate this pull request. Hopefully the scala-lang site processor, whatever it is, will handle a query string better than it does anchors.

andy1138:
I've found the problem with external links. There is a 'bug' in the code and the url gets lost for training events. see:  https://github.com/scala/scala-lang/pull/481

There is another issue where training events is not being filtered from 'news' items.  It can be fixed by adding 'category: event'  to the PR. However I dont know if 'event' is "conference only" or "conf and training"  will have to ask @heathermiller / @SethTisue / etc what it actually means

mslinn:
I wonder how the other training events worked, or was the bug recently introduced?

andy1138:
It looks like in Feb there was some work done on the website and yours is the first non-Typesafe/Lightbend training event to be added.  (TS/LB have a special feed to auto add training sessions). This is when  'category: event' was added to the conf. events but looks like they missed/forgot training :-(

(see bottom of:  https://github.com/scala/scala-lang/commit/f34dfd5aab45cf739742f636fe638b9d768e6ebe )

heathermiller:
Thanks @andy1138 for looking into it! Your help is much appreciated.

heathermiller:
Ok, the proposed fix didn't work. I do not have time to debug this now, so in order to restore the frontpage of scala-lang to functional state I'm reverting this pull request. If someone else has time to debug, I will happily merge!

Else, this will have to wait.

soc:
@mslinn Usually values with special characters require the whole value to be quoted, no need to change the URL scheme!

PR #481, opened 2016-09-28
Fixed external link on training

andy1138:
For training events the external link is being lost as its was looking in 'session' not 'training' where its being stored. 

NOTE: I dont get Lightbend training events when I run everything locally. So cant check this hasn't broken everything coming from this feed.

### DIFF
--- a/resources/js/main.js
+++ b/resources/js/main.js
@@ -417,7 +417,7 @@ $(document).ready(function(){
         result.push({
           title: training.title,
           description: training.description,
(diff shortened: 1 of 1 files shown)


SethTisue:
> I dont get Lightbend training events when I run everything locally

that's https://github.com/typesafehub/Typesafe.com/issues/1158

andy1138:
thanks :-)

Think that's a private repo but I did manage to hack my way around it and get working locally

PR #482, opened 2016-09-28
Fixes Lightbend training links

andy1138:
The previous 'fix' fixed the link for locally stored training events however broke the events coming from Lightbend. 
This now changed the local-events to look like the json ones

I think there are actually 2 problems
1) When 'category: event' was added to 'conference events', the 'training events' were not also updated and so they are appearing in news events. Adding this to local-training-events fixes the problem.  (this is the first non-LB event since the change)
2) this 'bug'

### DIFF
--- a/resources/js/main.js
+++ b/resources/js/main.js
@@ -393,13 +393,13 @@ $(document).ready(function(){
     {
       title: "{{ training.title }}",
       description: "{{ training.description }}",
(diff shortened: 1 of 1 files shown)


heathermiller:
Seems reasonable. Thank you Andy for digging around and providing a fix here!

heathermiller:
😞 something's still not right 😿

PR #484, opened 2016-09-29
Add 'category' to ScalaCourses training

andy1138:
Hopefully this is makes Mikes training event finally look correct

### DIFF
--- a/training/_posts/2016-10-18-scala-intro.md
+++ b/training/_posts/2016-10-18-scala-intro.md
@@ -1,4 +1,5 @@
 ---
+category: event
 title: Introduction to Scala Online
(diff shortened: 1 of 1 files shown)


heathermiller:
hrm

heathermiller:
🎉 🍻 Thank you Andy!!

andy1138:
Thats a good way to finish the week  😄 

And we should also invite @mslinn to the party 🍺   and hope is training course goes more smoothly than adding it to website 🍹

mslinn:
@andy1138, why thank you for the invitation! I love parties!

Also thank you for taking care of this. So nice to have a working link!

If this course gets enough students we will follow up with our Intermediate Scala course. We'd like to get the word out. [Retweets](https://twitter.com/ScalaCourses/status/781588917363572737) would be much appreciated!

PR #489, opened 2016-10-15
Updated company name to Lightbend

arunkpatra:


SethTisue:
thanks, but I don't think it's necessary to update blog posts that actually predate the name change.

PR #495, opened 2016-10-18
change .bundle/config to the format `bundle install` prefers

SethTisue:
when I run `bundle install`, it makes this change. then I have
an untracked change in my repo. so let's commit it?

review by @jarrodwb and/or @charafau


jarrodwb:
Lol. When I do `bundle install` it takes off the quotation marks.

I was using Bundler version 1.12.5. I just updated to 1.13.5 and it uses the quotation marks.

So this change looks good to me.

charafau:
I'm also using 1.13.5 and it works for me. LGTM for me.

PR #438, opened 2016-06-07
WIP: Update Scala Hacker Guide

chaselal:
Changes for [Issue #437](https://github.com/scala/scala-lang/issues/437).

I have not built the site, and I can't verify that all of the edits are accurate. However, this should at least provide a starting point for updating the guide.

PR #504, opened 2016-10-26
update hacker guide

adriaanm:
Follow up for #438


SethTisue:
LGTM. ideally we'd probably want to consider and update this and the Scala repo README.md together. but for now this seems like a clear improvement so I'll go ahead and merge. thanks @chaselal and reviewers

PR #499, opened 2016-10-22
get Gemfile in shape for 498

jarrodwb:
First shot at #498. It seems to build just fine.

I would be very interested in hearing about how the site is actually built and put on the server.

I think it would be just fine to build before it hits the server and just push the `_site` directory. This is how I push my site to S3. Then we would not need to worry about what is and is not installed on the server.

PR #507, opened 2016-11-03
Fix URLs to Jenkins jobs in README

ashawley:


jarrodwb:
Looking at the Travis checks it looks like the Ruby version should be stated in either the Gemfile or the Travis file. My bet is the Gemfile.

What version of Ruby is installed in production?
@SethTisue

ashawley:
@jarrodwb Probably using 2.2 for Travis is as good a guess as any.  So you did just that in 02c2e3c of #499

SethTisue:
@jarrodwb I'm baffled by this Ruby version issue. can you open a separate ticket on it?

SethTisue:
oh, I see now the Ruby thing is under discussion at https://github.com/scala/scala-lang/pull/499

PR #513, opened 2016-11-08
make Scalafix blog post not so entirely Dotty-centric

SethTisue:
we hope and expect that Scalafix will be useful in a Scala 2
context, too.  here are some proposed changes that reflect
that hope.

review by @heathermiller @olafurpg

reference: https://twitter.com/adriaanm/status/791077865672495104

Ólafur: using Scalafix in a Scala 2 context isn't really something
we've discussed, but it's certainly part of our plans!  We hope it's
part of yours, too.

### DIFF
--- a/blog/_posts/2016-10-24-scalafix.md
+++ b/blog/_posts/2016-10-24-scalafix.md
@@ -2,24 +2,21 @@
 layout: blog
 post-type: blog
 by: Ólafur Páll Geirsson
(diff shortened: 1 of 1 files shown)


olafurpg:
LGTM 👍 I'm glad you noticed there is nothing Dotty specific about scalafix. The post focused on Dotty only because that's my assignment at the Scala Center.

Thanks!

heathermiller:
LGTM! Thanks for the additions, @SethTisue

Issue #512, opened 2016-11-07
use same version of Jekyll as docs.scala-lang.org


SethTisue:
now that the production build is `bundler` based (thank you @fsalvi!!) we ought to move to Jekyll 3.3.0 to match the doc site, shouldn't we?

jarrodwb:
I think so. For that matter I think we should just use the same verbatim Gemfile.

https://github.com/scala/scala.github.com/blob/master/Gemfile

I have not tested it locally yet.

jarrodwb:
It looks like life is always a little bit more complicated than it should be.

https://github.com/github/pages-gem/issues/351

I suggest we use the simple workaround with `>=104`. This problem may also come up on the docs site in the future.

jarrodwb:
@SethTisue Do we still need these two gems? I see you added them to the Gemfile about a year ago. The site builds localy without error when they are removed. Maybe they are needed in production?

https://github.com/rtomayko/posix-spawn
https://github.com/brianmario/yajl-ruby

SethTisue:
> Do we still need these two gems

I have no idea. I would suggest removing them, then monitoring the next Jenkins run after we merge. if it still succeeds, we're good.

fsalvi:
It seems posix-spawn and yajl-ruby are used by pygments.rb

gem dependency pygments.rb
Gem pygments.rb-0.6.3
  posix-spawn (~> 0.3.6)
  rake-compiler (~> 0.7.6, development)
  yajl-ruby (~> 1.2.0)

SethTisue:
where do we use or have a dependency on pygments?

fsalvi:
I'm not sure, maybe it was used in earlier versions of jekyll. 
I guess these dependencies can be removed. I just tried on the server and it doesn't change anything...

SethTisue:
@fsalvi excellent

fixed by #517

PR #517, opened 2016-11-09
make Gemfile depend only on github-pages

jarrodwb:
PR for issue #512 

I don't think losing `posix-spawn` and `yajl-ruby` will cause any problems. If they do we can add them back in.


SethTisue:
nice!

PR #523, opened 2016-11-11
Adding quotations to event

heathermiller:


SethTisue:
http://www.scala-lang.org/events/ looks OK even without this...?

heathermiller:
You're right. Sorry for the noise then!

andy1138:
As a little side note.
As far as I can tell, 'description' is not (currently) used for conference events so will not make that much difference. But think its a good idea to at least keep in a good format, I may even have a look to see if it can be used as a tooltip.
 
It is used in the js for training events however is not displayed.

SethTisue:
okay!

PR #530, opened 2016-11-15
recommend sbt, not Activator

SethTisue:
and remove all mention of Activator from the site
(well, except from some old release announcements)

### DIFF
--- a/_layouts/downloadpage.html
+++ b/_layouts/downloadpage.html
@@ -29,18 +29,12 @@ layout: page
   <div class="bigcircle" style="float: right;">2</div>
   <div class="bigcircle-content" style="float: right; margin-top: 32px;">
     <p class="center" style="text-align: right;">
-      <b><a href="http://www.lightbend.com/activator/">Get started with Lightbend Activator</a></b>
+      <b><a href="http://www.scala-sbt.org/0.13/docs/Setup.html">Get started with sbt</a></b>
       <br/>
-      Lightbend Activator is a browser-based or command-line<br/>
-      tool that helps developers get started with Scala.
-      <div class="download-wrapper">
-        <a id="download-button" class="activator" style="margin-top: 12px; margin-right: 24px; width: 200px;" href="http://downloads.lightbend.com/typesafe-activator/1.2.10/typesafe-activator-1.2.10.zip">
-          <span class="slider">Download</span>
-          <span class="activator icon"></span>
-          <span class="version-long">&nbsp;Lightbend Activator</span>
-        </a>
-        <p class="install"><a style="margin-left: 124px;" href="http://www.lightbend.com/activator/download">Need help installing?</a></p>
-      </div>
+      sbt is a popular build tool for Scala projects.<br/>
+      With sbt, you don't need to download and install Scala.<br/>
+      Just set <code>scalaVersion</code> in your build definition,<br/>
+      and sbt will retrieve that version.
     </p>
   </div>
 </div>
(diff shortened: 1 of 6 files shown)


SethTisue:
review by @heathermiller @adriaanm @eed3si9n @dwijnand @corruptmemory @tpolecat...?

SethTisue:
screen shot 2016-11-15 at 3 32 32 pm

tpolecat:
I think I might prefer landing at `http://www.scala-sbt.org/` since it's more attractive and it links to a nicer download page that you would otherwise miss. It's would also sidestep the problem of the link being version-specific. But in any case this is a step in the direction I'd like to go so 👍

SethTisue:
> I think I might prefer landing at http://www.scala-sbt.org/ since it's more attractive and it links to a nicer download page that you would otherwise miss

yeah, I was unsure about that. Eugene, Dale, you want to make the call there?

eed3si9n:
Sure. Landing on the landing page makes sense.

_layouts/downloadpage.html
```diff
@@ -29,18 +29,12 @@
...
-      tool that helps developers get started with Scala.
-      <div class="download-wrapper">
-        <a id="download-button" class="activator" style="margin-top: 12px; margin-right: 24px; width: 200px;" href="http://downloads.lightbend.com/typesafe-activator/1.2.10/typesafe-activator-1.2.10.zip">
-          <span class="slider">Download</span>
-          <span class="activator icon"></span>
-          <span class="version-long">&nbsp;Lightbend Activator</span>
-        </a>
-        <p class="install"><a style="margin-left: 124px;" href="http://www.lightbend.com/activator/download">Need help installing?</a></p>
-      </div>
+      sbt is a popular build tool for Scala projects.<br/>
```
jarrodwb: Should sbt be capitalized here? It is the beginning of a sentence.

contribute/documentation.md
```diff
@@ -7,7 +7,6 @@ title: Documentation Contributions
 There are several ways you can help out with the improvement of Scala documentation. These include:
 
 * API Documentation in Scaladoc
-* Code examples and tutorials in activator templates.
```
jarrodwb: Sbt 13.13 has a `new` command that provides templates. Maybe this is worth mentioning.

_layouts/downloadpage.html
eed3si9n: sbt is a proper noun like iPhone and eBay.

contribute/documentation.md
charafau: Would it be possible for `sbt` to provide some examples for basic projects?

heathermiller: I agree that it would be good to link people to either documentation about `new` or some kind of template. @dwijnand and/or @eed3si9n, do you have any suggestions here?

heathermiller: I agree that it would be good to link people to either documentation about `new` or some kind of template. @dwijnand and/or @eed3si9n, do you have any suggestions here?

heathermiller: Anyway, this is the wrong place to suggest these changes. I think we're making suggestions for improving the getting started page, which is not the file we're commenting on (we're commenting on a contributing page)

dwijnand: For the getting started page, you can point to http://www.scala-sbt.org/0.13/docs/sbt-new-and-Templates.html

SethTisue:
I'm going to go ahead and merge this, and leave it to someone else to add stuff about `sbt new` if they like, in a separate PR. (I think it's a good idea.) Thanks all for the review. We can always go on and make further changes.

SethTisue:
on the subject of possible revisions to the getting-started page, I opened https://github.com/scala/scala-lang/issues/536 (but I probably won't tackle it myself)

PR #533, opened 2016-11-16
fix a lot of broken links

jarrodwb:
I am used https://github.com/gjtorikian/html-proofer to find a ton of broken links. Here is some fixes.


SethTisue:
Could htmlproofer be part of the Travis build?

jarrodwb:
Sure. It would take some work to do. I would probably set it up as a rake task.

It would require different PR though.

SethTisue:
made that suggestion #535.

this is fantastic, thanks!! merging

PR #537, opened 2016-11-16
Fix Jekyll exit status for Travis

ashawley:
Grep the output of Jekyll for errors since it returns zero exits with
YAML errors.  See jekyll/jekyll#5257.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,6 @@ install:
 - bundle install
 
 script:
-- bundle exec jekyll build
+- bundle exec jekyll build 2> error.log
+- cat >&2 error.log
+- ( ! grep -qie Error error.log )


.travis.yml
```diff
@@ -16,4 +16,6 @@ install:
 - bundle install
 
 script:
-- bundle exec jekyll build
+- bundle exec jekyll build 2> error.log
+- cat >&2 error.log
+- grep -qie Error error.log && false || true
```
jarrodwb: How should I read this line? Just from reading it it seems like:
```
true && false || true ==
(true && false) || true ==
false || true ==
true
```
and
``` 
false && false || true ==
(false && false) || true ==
false || true ==
true
```
So either way it will return true.
I fully admit I maybe missing something here.

dwijnand: You're right. (http://stackoverflow.com/questions/17203122/bash-if-else-statement-in-one-line#comment33428887_17203203)

I recommend instead using:

```bash
[ "$(grep -cie Error error.log)" -eq 0 ]
```

ashawley:
Agreed, it might not be obvious right away what this means in shell:

```sh
grep -qie Error error.log && false || true
```


This would be the simplest thing:

```sh
! grep -qie Error error.log
```

Less cryptic than the previous?

I assume putting a bang there is supported by Travis, but I didn't check

jarrodwb:
It is not really about being cryptic. It is about having the intended behavior. I don't think the two code snips do the same thing. I suspect that your second version will work.

Are you interested in testing it out and updating the PR? You could play around with Travis on your fork and then when it all works you could push the changes into this PR.

Thanks for your work on this. :-)

ashawley:
You're right.  Thanks for catching this.

Looks like a bare exclamation point won't work, but I can use bang in parens:

```sh
( ! grep -qie Error error.log )
````

**Comment by ashawley** [comment]
Introducing Seth's broken patch in 748c635 causes Travis to fail.

**Comment by ashawley** [comment]
The errors are produced by the `cat` command, so they show up out of order, but it works.

```
$ bundle exec jekyll build 2> error.log
Configuration file: /home/travis/build/scala/scala-lang/_config.yml
            Source: /home/travis/build/scala/scala-lang
       Destination: /home/travis/build/scala/scala-lang/_site
 Incremental build: disabled. Enable with --incremental
      Generating... 
                    done in 12.212 seconds.
 Auto-regeneration: disabled. Use --watch to enable.
The command "bundle exec jekyll build 2> error.log" exited with 0.
$ cat >&2 error.log
Error reading file /home/travis/build/scala/scala-lang/download/index.md: (<unknown>): did not find expected ',' or ']' while parsing a flow sequence at line 7 column 17 
The command "cat >&2 error.log" exited with 0.
$ ( ! grep -qie Error error.log )
The command "( ! grep -qie Error error.log )" exited with 1.
```

File: .travis.yml [hunk 2]
```diff
@@ -16,4 +16,6 @@ install:
 - bundle install
 
 script:
-- bundle exec jekyll build
+- bundle exec jekyll build 2> error.log
+- cat >&2 error.log
```
jarrodwb: Under what conditions will this line exit with a non-zero status?

jarrodwb: I guess what I am getting at is I don't understand what the `>&2` is for.

ashawley: The `2>` pipes stderr to error.log

The `>&2` pipes stdout to stderr

> I guess what I am getting at is I don't understand what the >&2 is for.

It makes the font of the error message "Yellow" in Travis?

jarrodwb: Okay `>&2` is starting to make more sense.

But why `cat >&2 error.log` instead of `cat error.log >&2`.

Thanks for your time. I know your job is not to spend all day educating me. ;-)

ashawley: Yeah, no difference.  Let me know if `>&2` at the end is clearer.

Thank you for the careful review!

SethTisue:
wonderful!

PR #516, opened 2016-11-08
little download page tweaks

SethTisue:

PR #538, opened 2016-11-17
Fail Travis builds if Jekyll outputs empty files

ashawley:
While reading #516, it seemed to me that  checking for empty files might be useful.

Turns out it is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,5 @@ script:
 - bundle exec jekyll build 2> error.log
 - cat >&2 error.log
 - ( ! grep -qie Error error.log )
+- find ./_site -empty ! -name error.log
+- find ./_site -empty ! -name error.log | ! read
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,5 @@ script:
 - bundle exec jekyll build 2> error.log
 - cat >&2 error.log
 - ( ! grep -qie Error error.log )
-- find ./_site -empty ! -name error.log
-- find ./_site -empty ! -name error.log | ! read
+- find ./_site \( -type d \( -name events -o -name training \) -prune -false \) -o \( -type f ! -name error.log -empty \)
+- find ./_site \( -type d \( -name events -o -name training \) -prune -false \) -o \( -type f ! -name error.log -empty \) | ( ! read )


ashawley:
Failed with the following empty files:

```
$ find ./_site -empty ! -name error.log
./_site/training/2015/02/24/essential-scala-online.html
./_site/training/2015/03/19/essential-scalaz.html
./_site/training/2015/03/20/shapeless.html
./_site/training/2015/03/28/essential-essential-scala.html
./_site/training/2015/03/30/advanced-scala.html

The command "find ./_site -empty ! -name error.log" exited with 0.

$ find ./_site -empty ! -name error.log | read && false
The command "find ./_site -empty ! -name error.log | read && false" exited with 1.
```

ashawley:
The pages are blank:

www.scala-lang.org/training/2015/02/24/essential-scala-online.html
www.scala-lang.org/training/2015/03/19/essential-scalaz.html
www.scala-lang.org/training/2015/03/20/shapeless.html
www.scala-lang.org/training/2015/03/28/essential-essential-scala.html
www.scala-lang.org/training/2015/03/30/advanced-scala.html

Are they redirects?

jarrodwb:
I like this idea. :+1:

.travis.yml
```diff
@@ -17,3 +17,5 @@ install:
 
 script:
 - bundle exec jekyll build
+- find ./_site -empty ! -name error.log
+- find ./_site -empty ! -name error.log | read && false
```
ashawley: Like #537, I'm trying to be explicit about return status, but this would likely work as well:

```bash
find ./_site -empty ! -name error.log | ! read
```

ashawley: But you're right, I had forgotten to add `|| false`

ashawley: I'm completely confused here.  Looks like ! will work in parens

jarrodwb:
check out your travis logs https://travis-ci.org/scala/scala-lang/builds/176708368

.travis.yml
ashawley: You're right, I've changed it to ` ! read`

ashawley:
It turns out these empty files output by Jekyll are empty purposefully.  There should be even more caught by the find command, since there are many markdown files that contain only YAML front matter.  Some of those files have an extra line after the front matter, so they aren't seen as empty.  The find command could surface those with `-size 1c`, in addition to the `-empty`.

Paths that contain these purposeful empty files could be ignored.  Currently, those paths are

- events
- training

The find command to ignore two paths is so unwieldily it's probably not worth mucking up the build process with it.

So, I'm not sure how useful it is now.  We should probably just see how #537 goes.

SethTisue:
> It turns out these empty files output by Jekyll are empty purposefully

why?

ashawley:
Files with only front matter is how these sections of the site function:

https://github.com/scala/scala-lang/blob/master/training/README.md
https://github.com/scala/scala-lang/blob/master/events/README.md

It's not well documented, but this code is relevant I believe
https://github.com/scala/scala-lang/tree/master/resources/php
https://github.com/scala/scala-lang/blob/master/resources/js/main.js

SethTisue:
> The find command to ignore two paths is so unwieldily

would adding `-name events -prune -o -name training -prune -o` work?

ashawley:
Yes, doesn't it require a lot of parens?

This what I had

```shell
find ./_site \( -type d \( -name events -o -name training \) -prune -false \) -o \( -type f ! -name error.log -empty \)
```

jarrodwb:
I am starting to think that some of these Travis checks should be pulled out into a Scala script. I am not a fan of using Bash and Unix commands if it is not immediately obvious what is going on.

ashawley:
I agree, the shell code is ugly.

Looks like a script was written in Ruby in the past

https://github.com/scala/scala-lang/tree/master/_tools/validation

Not sure if it's used or not, still.

jarrodwb:
I don't think it is. I am actually working on a `spring-cleaning` branch locally where I removed the `_tools` directory.

SethTisue:
> I am starting to think that some of these Travis checks should be pulled out into a Scala script

SGTM. I see no problem in taking on a Scala dependency.

SethTisue:
regardless, shall we merge? (I suggest we do.)

jarrodwb:
Sure. :+1:

Issue #532, opened 2016-11-16
We need a dev Scala version in _config.yml


jarrodwb:
713e699c8cfd2e454f25fb0595555788b36c703d introduced a bug that breaks the Development Release link in the footer of all the pages.

We need a value for `devscalaversion` that points to an actual page. The current rendering gives a link of `http://www.scala-lang.org/download/.html`.

SethTisue:
hmm. not clear how to fix until 2.13.0-M1 comes out, which won't be for a while (a few months at least).

I really don't think we need to make the development version _that_ easy to find, so I'd be fine with a PR that simply removed devscalaversion and removed the link from the footer.

SethTisue:
fixed (in a better way) by #539

PR #539, opened 2016-11-17
Remove develepment version from footer

jarrodwb:
fix for #532 

Also note that I changed Stable to Current. I can change this back or to something else people prefer.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,7 +1,6 @@
 title: The Scala Programming Language
 
 scalaversion: "2.12.0"
-devscalaversion: ""
 
 baseurl: ""
 markdown: kramdown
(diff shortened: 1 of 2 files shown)


SethTisue:
I didn't have this idea until just now, but what do you think about replacing "Development version" with an "All versions" link that goes to http://www.scala-lang.org/download/all.html ?

jarrodwb:
Hey Seth. Yeah I can do that. Probably tomorrow though. :-)

jarrodwb:
We now have "Current Version" and "All Versions" in the footer.

I also ran `grep -r site.devscalaversion .` and did not get any hits. I think we are ready to go with this.

heathermiller:
LGTM! :)

heathermiller:
Thanks @jarrodwb!

PR #543, opened 2016-11-19
Fix formatting problems in hacker guide

ashawley:


jarrodwb:
Thanks for catching this Aaron. :-)

Issue #518, opened 2016-11-09
regression: Twitter section on front page is empty


SethTisue:
I assume we're using some ancient API that doesn't work anymore?

a very old related issue: https://github.com/scala/scala-lang/issues/100

jarrodwb:
This bug may actually be an improvement. ;-)

jarrodwb:
What do people think about removing this section from the front page? I have been playing around with the idea locally and I like it. People can still find the Twitter account in the header.

SethTisue:
I lean, mildly, towards keeping it. For better or worse, Twitter is our main way of getting news out.

PR #547, opened 2016-11-21
Fix twitter feed on home page

ashawley:
Use JavaScript from https://publish.twitter.com/

Parameter reference is

https://dev.twitter.com/web/embedded-timelines/parameters


SethTisue:
fixes #518

jarrodwb:
I don't remember how the site looked before. It seems to work fine with this fix. The Twitter feed is prominent.

In short the merge looks like it will work. Thanks!

SethTisue:
@ashawley thanks for the fix!

as for the appearance, I've opened https://github.com/scala/scala-lang/issues/548 on that

PR #549, opened 2016-11-21
Fix #548 too big Twitter feed

### LINKED ISSUE #548: tweets section is way too big
can we control the appearance of this? I don't remember what it looked like before, since it's been broken for some time, but how it looks now — well, “prominent” is the nice way @jarrodwb put it, I was thinking more along the lines of "gigantic and hideous" :-)

ashawley:
Configure width of Twitter widget to 520 pixels


sjrd:
LGTM

PR #552, opened 2016-11-23
Move Jekyll base template back to default

ashawley:
Follow-up to changes in #550

Keep the naming consistent with the Jekyll convention:

https://jekyllrb.com/docs/templates/

Additionally, drop the extra YAML front-matter in the default layout.


jarrodwb:
LGTM. I like the YAML front matter but it is not a strong preference.

PR #554, opened 2016-11-26
Use maxcdn for bootstrap.min.js

ashawley:
I was going to do the bootstrap CSS file as well, but noticed the CSS files uses was a forked version.  I'll create resolving the fork, possibly upgrading bootstrap to a more current version, and using a CDN in a separate issue.


heathermiller:
Great, thanks @ashawley! LGTM!

PR #555, opened 2016-11-26
Use Google CDN for jQuery 1.8.3

ashawley:
This obviously helps optimize the site at the expense of making it difficult to work on the site off-line.  Not sure how many people do that these days, though.  Also, not sure how many sites still use jQuery 1.8.3, so it may be a silly optimization to use Google's CDN.


heathermiller:
Ah yeah, I didn't consider the off-line aspect of this and the other maxcdn change. Personally, I do generate the pages offline and use that quite a bit for development. Perhaps it's best to keep the assets local?

ashawley:
Firefox still has an offline mode.  Firefox needs to load the site's assets once and have them cached before going in to offline mode.  It would work for offline editing of the site locally.  Any more serious offline development would require a network connection to download different frameworks.   It looks like edits to main.js or main.css are only a few times a year.  Was there a use case in-between those that you had in mind?

Should we investigate an "offline" config for Jekyll, so it can switch between local assets and a CDN?

I find CDNs more developer-friendly than maintaining blobs of CSS or Javascript or by making them gem dependencies. That's my personal experience, though.

SethTisue:
While I don't have a strong feeling about it, on the balance I suggest we merge. Offline development seems like a niche concern to me. I don't think it's unreasonable for web work to require an internet connection.

jarrodwb:
LGTM

SethTisue:
I guess I'll merge and we'll see how we like it.

jarrodwb:
:+1:

Issue #372, opened 2015-12-16
Make layout responsive (again, again?)


ebruchez:
I see #19, but the layout is clearly not responsive now. The main Scala web site _must_ be readable on a smartphone.

heathermiller:
This complaint could be a bit more helpful. Currently you tell us what we  _must_ be doing, which is of course not very constructive. What browser are you using? What do you want to see?

At the time that I designed the Scala website, there were no responsive-first CSS frameworks and many aspects of the current layout wouldn't fit into any currently popular responsive CSS framework. The Scala website was a (large) contribution that I made on the side as a student. Given that we have no dedicated web/design person any large overhauls to the design/layout of the website is quite out of scope. 

Although we would be happy to receive such a contribution!

SethTisue:
@ebruchez I don't know very much about CSS or about responsive design, but... can you (or anyone else who happens to see this ticket) identify any modest changes we could make that would improve the situation?

ebruchez:
Sorry @heathermiller not a criticism of your work, I thought this was an obvious one. Take Chrome dev tools and set it e.g. for an iPhone 6S Plus. You see this:

![iphone6plus](https://cloud.githubusercontent.com/assets/105769/11876892/c9ad0250-a4a0-11e5-9e39-c6f67ca19342.png)

What I expect on a mobile phone is (at least for the news part but ideally for the whole site):

- single-column layout
- larger, readable font

I can't look at this immediately so can't tell if it's easy or hard to just hack in a couple of CSS fixes to improve the experience.

heathermiller:
It would require pretty significantly changing the layouts of all pages. One would have to rebase all pages on newer "responsive-first" CSS frameworks. Many aspects of the front and inner pages would have to change because they can't be easily translated to a 12-col "responsive" grid format as in frameworks like Bootstrap. So the way we do the sidebars on inner pages, and the way we display blog entries etc on the front page would all have to change.

As this requires non-trivial changes to the layout across the entire website, and since we have no one who can invest time in this, I have to mark this as out-of-scope.

So this will have to fall into the realm of "big contributions" if we would like to see it done.

ebruchez:
I think that a full modern layout would be best, and I agree that would be a relatively big contribution. It might be possible to graft something purely in CSS, as @SethTisue suggests. Hard to know for sure without trying. I suggest keeping this open as ultimately it will have to be addressed.

soc:
[...]

PR #558, opened 2016-11-27
Unfork bootstrap CSS file

ashawley:
The current bootstrap CSS version is 2.3.0

Fork is essentially 08bab9f, but also d985737

I spent some time experimenting, and it seemed that bootstrap.css wasn't entirely forked, but that the non-responsive CSS files was used.  I've switched to the copy provided by MaxCDN for 2.3.0 in the hopes it can switch to that (see below).

Changing the bootstrap CSS file, required small subsequent style changes.  For instance, I converted `span12` to `span10` on the frontpage.  I also changed the width of the main page column on other pages.  I hope they're close enough to the spirit of the original.

The bootstrap JS is 2.2.1.  It could probably be 2.3.0, as well.  But I'll hold off so we can identify any regressions with these CSS changes.

With the fork resolved, the motivation here is that

1. minification could be used, 
1. bootstrap could be moved to the last version 2 release, 2.3.2,
1. a CDN of the published bootstrap version could be used instead

See #554 for a PR that uses a CDN for the bootstrap JS file.

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -8,7 +8,7 @@
   <div class="darkblue-overlay">&nbsp;</div>
   <div class="container">
     <div class="row">
-      <div class="span12">
+      <div class="span10 offset1">
         <div class="scala-logo">
           <img src="{{ site.baseurl }}/resources/img/scala-logo-white.png" alt="Scala Logo" />
         </div>
@@ -16,7 +16,7 @@
       </div>
     </div>
     <div class="row">
-      <div class="span12">
+      <div class="span10 offset1">
         <div class="splash-text ontop">
           <h2 style="font-weight: 700;">Object-Oriented Meets Functional</h2>
           <p class="splash-pitch">
@@ -69,7 +69,7 @@
   <div class="bullet-row-1-wrap">
     <div class="container">
       <div class="row">
-        <div class="span12 bullet-point">
+        <div class="span10 offset1 bullet-point">
           <h2><span>Scala in a Nutshell</span></h2>
           <p class="click">&nbsp;click the boxes below to see Scala in action!&nbsp;</p>
         </div>
@@ -129,7 +129,7 @@
   <div class="where-to-learn-more">
     <div class="container">
       <div class="row">
-        <div class="span12">
+        <div class="span10 offset1">
           <a class="learn-more-btn" href="{{ site.baseurl }}/what-is-scala.html">Learn More</a>
           <p>or visit the <a class="italicize" href="{{ site.baseurl }}/documentation">Scala Documentation</a></p>
         </div>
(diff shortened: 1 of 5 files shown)


jarrodwb:
I like the idea here. It may not be ready though. I think the new bootstrap version maybe causing the site to be even less responsive.
scala-logo-off

Do you mind playing around with various screen sizes on your end? I am mostly worried about people on smartphones.

ashawley:
Thanks for reviewing this.  Don't know how I forgot to look at it in other devices.

To be clear, I'm trying to use an official bootstrap version, the same one, not a newer one.

I've reverted the change to the download and API buttons.  This fixes it on phones, but changes the look on the desktop.

Reverting the change I made to the spiral's position will also work on other devices.  However, unlike the download and API buttons, the spiral will just not be where it should on the desktop.  I feel like positioning the spiral logo on the site in a responsive way is difficult.  It's like throwing darts at a dartboard.  I'm not good at playing darts.

jarrodwb:
I also spend most of my time on the backend so I don't have a good idea about positioning the logo.

I checked again using e41bbe52b618db03 and seems like the smaller widths have a problem. I am using Chrome developer tools.

1220px width
-------------------
width-1220

1100px width
------------
width-1100

Maybe we can address this when we do a complete overhaul of the site? What do you think is best?

vossad01:
> Maybe we can address this when we do a complete overhaul of the site?

@jarrodwb Is there anything tracking this.  This was perhaps alluded to in https://github.com/scala/scala-lang/issues/372 as well.

ashawley:
Hey @jarrowdb, thanks for investigating it.

If the bootstrap CSS file can't be re-aligned with the released version, I guess I was going to try and investigate converting the site to a newer version of bootstrap.  Hopefully, newer versions would behave more sanely for both mobile and desktop layouts.

heathermiller:
Hey @ashawley, thanks a lot for looking at this and trying to take it on. So far I think no one has actually succeeded in making the frontpage truly mobile-friendly, even in the forks. The trouble is the download/API buttons and the Scala swirl. This is outside of Bootstrap, if I remember correctly. So I'm actually not sure that updating Bootstrap will solve the problems that @jarrodwb pointed out :-/

SethTisue:
so I guess that means we should go ahead and merge, since there are problems on mobile regardless, but being on a newer bootstrap is a good thing...?

jarrodwb:
Please don't merge. The problem with the logo is introduced on desktops which was not present before. Unless this is fixed I don't see a net benefit from merging.

Cheers. Heading to bed but will be available on Monday for a more detailed discussion if anyone desires it.

SethTisue:
ok, thanks for the heads up :-)

ashawley:
Indeed, hold off on this given the rendering issues.

PR #562, opened 2016-11-30
remove SLIP references from community page

SethTisue:
SPP stuff will be added soon instead, in the meantime let's not link
to dead stuff


jvican:
Thanks @SethTisue :+1:

PR #565, opened 2016-12-01
Provide some historical context for the platform

adriaanm:
I think it's important to mention the history behind this effort,
and also to clarify the core will still include e.g., the collections,
which are truly core to the language.


SethTisue:
@jvican the blog post has your name on it, so I think we would need your review before merging?

jvican:
LGTM

PR #568, opened 2016-12-05
keep download pages out of "What's New"

SethTisue:


SethTisue:
before:

screen shot 2016-12-05 at 3 15 18 pm

after:

screen shot 2016-12-05 at 3 16 19 pm

merging on the fast track since 2.12.1 just came out so this is a problem right now

PR #580, opened 2016-12-20
Remove jQuery parallax and stellar JS files

ashawley:
Parallax and stellar were disabled a while back in 3d8db7d

Alternatively, we could give it another go

 Perhaps it might be possible to make it work?  I guess one could start here:

http://markdalgleish.com/2012/10/mobile-parallax-with-stellar-js/


SethTisue:
(If they're disabled, removing them from the repo seems like clearly the right thing to me. No clue whether another try is worthwhile.)

SethTisue:
But, there is still a reference to `jquery.stellar.min.js` in `_includes/footer.html` ...?

ashawley:
So there is.  Good catch. Thanks. 

 When I'm back from holiday i'll fix.  Closing for now.  Sounds like a site redesign is in the works anyway.  That's good news.

heathermiller:
Hey @ashawley, you're right that these were disabled, never to have been reenabled. I don't think we need these at all, we don't need to try to re-enable them, and a site redesign isn't in the works, not exactly. We're just fixing a few major issues like responsivity, no complete overhaul.

So I think this is good to merge. I can make a followup commit removing stellar.min.js from the footers.

Thank you!!

heathermiller:
Thank **you** for the cleanup!

PR #585, opened 2017-01-06
Advertising the spreadsheet of events

propensive:


heathermiller:
Thanks! Seems like a good idea to me :)

PR #594, opened 2017-02-15
add "Who's behind Scala?" to the community page

SethTisue:
just a draft, feedback welcome. this came out of a discussion on https://gitter.im/scala/center

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -26,7 +26,7 @@ Scala was created by
 [Prof. Martin Odersky](http://lampwww.epfl.ch/~odersky/).  Martin and
 his research group at EPFL remain heavily involved with the language,
 including developing [Dotty](http://dotty.epfl.ch), a
-next-generation research compiler for Scala.
+next-generation compiler for Scala.
 
 ## Forums and Mailing Lists
 
--- a/community/index.md
+++ b/community/index.md
@@ -22,6 +22,12 @@ compiler, and standard library.  The Scala Center focuses on education
 outreach, and tooling.  Community participation in all of these
 efforts is strongly encouraged.
 
+Scala was created by
+[Prof. Martin Odersky](http://lampwww.epfl.ch/~odersky/).  Martin and
+his research group at EPFL remain heavily involved with the language,
+including developing of [Dotty](http://dotty.epfl.ch), a
+next-generation research compiler for Scala.
+
 ## Forums and Mailing Lists
 
 For discussions around Scala, we are in the process of migrating from
--- a/community/index.md
+++ b/community/index.md
@@ -25,7 +25,7 @@ efforts is strongly encouraged.
 Scala was created by
 [Prof. Martin Odersky](http://lampwww.epfl.ch/~odersky/).  Martin and
 his research group at EPFL remain heavily involved with the language,
-including developing of [Dotty](http://dotty.epfl.ch), a
+including developing [Dotty](http://dotty.epfl.ch), a
 next-generation research compiler for Scala.
 
 ## Forums and Mailing Lists
--- a/community/index.md
+++ b/community/index.md
@@ -8,9 +8,24 @@ Popular ways to connect with the Scala community include forums, chat rooms, loc
 
 The community is also the source of many libraries, tools, and other resources around Scala.
 
+## Who's behind Scala?
+
+The Scala language and associated websites are cooperative projects of
+the [Scala Center](https://scala.epfl.ch) at [EPFL](http://epfl.ch),
+[Lightbend](https://www.lightbend.com), and the Scala community
+more broadly, with participation from many companies, organizations,
+and individuals.
+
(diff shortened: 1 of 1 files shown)


community/index.md
```diff
@@ -8,9 +8,30 @@ Popular ways to connect with the Scala community include forums, chat rooms, loc
...
+compiler, and standard library.  The Scala Center focuses on education
+(especially online courses), documentation, open source community
+outreach, and tooling.  Community participation in all of these
+efforts is strongly encouraged.
+
+Scala was created by
+[Prof. Martin Odersky](http://lampwww.epfl.ch/~odersky/).  Martin and
+his research group at EPFL remain heavily involved with the language,
+including developing of [Dotty](http://dotty.epfl.ch), a
+next-generation research compiler for Scala.
```
sjrd: Should we mention Scala.js and Scala Native here in addition to Dotty?

SethTisue: Maybe... it's hard to know where to stop! They are already mentioned farther down the page, and I left the door open with “including”. I don't have a strong feeling about it, one way or the other.

SethTisue: It's complicated because Scala.js and Scala Native are now partly Scala Center efforts, too.

sjrd: > They are already mentioned farther down the page

Ah yes, I didn't notice that. Let's not add them here too, then.

SethTisue: Another possibility would be to omit Dotty here, since it's mentioned later, and end the sentence with “remain heavily involved with the language and community."

DarkDimius: I think it's better to keep Dotty here, as `remain heavily involved with the language and community.` sounds vague. In case we don't want Dotty mentioned twice, we could remove second mention, but I don't see any harm in mentioning twice.

community/index.md
```diff
@@ -8,9 +8,30 @@ Popular ways to connect with the Scala community include forums, chat rooms, loc
...
+
 ## Forums and Mailing Lists
 
-We are in the process of migrating from Google Groups to Discourse.
+For discussions around Scala, we are in the process of migrating from
+Google Groups to Discourse.
```
ktoso: make this a link?

SethTisue: The contributors.scala-lang.org link is intentionally postponed because newbies are likelier to want scala-user. (Eventually, users.scala-lang.org will be linked instead, but it isn't ready yet.

Or do you mean link to discourse.org, to answer the "what is Discourse?" question, since many people won't know? I think in an earlier draft I actually had linked to discourse.org, but I took it out because we want to keep the focus on Scala — we don't want someone going off on a tangent. We'd rather they just clicked on the link to contributors.scala-lang.org, which follows pretty quickly — they'll see very quickly there what Discourse is like.

ktoso: Ah I see, I missed the fact user wasn't ready yet - yeah I meant users.scala-lang, let's add when ready then :) Great addition overall.

community/index.md
```diff
@@ -8,9 +8,30 @@ Popular ways to connect with the Scala community include forums, chat rooms, loc
...
+compiler, and standard library.  The Scala Center focuses on education
+(especially online courses), documentation, open source community
+outreach, and tooling.  Community participation in all of these
+efforts is strongly encouraged.
+
+Scala was created by
+[Prof. Martin Odersky](http://lampwww.epfl.ch/~odersky/).  Martin and
+his research group at EPFL remain heavily involved with the language,
+including developing [Dotty](http://dotty.epfl.ch), a
+next-generation research compiler for Scala.
```
propensive: I'm hesitant to critique this description without suggesting a better alternative, but "next-generation research compiler" has the potentially-undesirable implication that Dotty is *just* an alternative compiler for Scala 2.x code, and that the "research" is into "better compilation methods", rather than "language improvements".

OK, how about just moving "next-generation" to immediately before "Scala", and deleting the second "research"?

s3ni0r: I totally agree with @propensive, i think many are confused by this, just today i read a blog post about how dotty will diverge from what Scala is up to 2.12.x, and that all the major contributors to the language are working only on Dotty and ignoring scalac, and the feedback from the community was awesome to weaken the false claim, and some were using scala 3 to indicate that it will be a continuation.
I found the post when i searched "Scalac vs Dotty" to know what feature Dotty has.

SethTisue: I have pushed a commit omitting the word "research". That brings the wording in line with http://dotty.epfl.ch which simply says "A next generation compiler for Scala". I don't feel I should take @propensive's other suggestion unless/until the same change were also made on the Dotty site.

I understand that there is some confusion and uncertainty in the community about the status of Dotty, but there's only so much it's appropriate to do in this paragraph to address it. I think http://dotty.epfl.ch does a good job with this already, and about all we can do here is say a few words and give a link.

SethTisue:
merged this round of work. there can still be more rounds, followup PRs welcome.

PR #599, opened 2017-02-22
GSoC ENSIME debugger and data annotation

fommil:


fommil:
More to come after ScalaSphere!

@xeno-by I hope you're ok with being mentioned in this idea :smile:

julienrf:
Thanks a lot @fommil !

PR #616, opened 2017-03-12
Add ScalaC performance work to GSoC proposals

rorygraves:


rorygraves:
@julienrf Can you review/merge this please?

fommil:
@julienrf ping

jvican:
Hey @rorygraves, thanks for proposing this!
  
I'm not in charge of GSoC but I'd like to share my mind with you. In my opinion, I don't think this is a good GSoC project. These are the reasons:
  
1. The proposed topic is too complex. The likelihood of finding a student that has the skills to carry out benchmarking and performance work is close to zero. It requires expertise of both Scala, JVM and compiler internals. 
2. The time to carry it out is only three months. Even if you find a good student, she won't be able to learn everything in that limited period.
3. The description is a little bit vague and unclear.
4. The likelihood of doing progress on this area is very low, and even more if it's noticeable.
  
Because of this, I would prefer to have other GSoC projects that are more achievable. I think there are lots of interesting projects that would deserve a slot before this one because they are more likely to get finished and be a success.

SethTisue:
Well put, Jorge. I concur. (And, thanks for writing the reasons up before any of us at Lightbend got around to it.)

I don't want to give the impression that the Scala team doesn't welcome participation in this kind of work, though... it's just not a good fit for GSoC in particular.

here's some links for Rory and anyone else interested in this area of work:

* in the Scala 2.13 roadmap, @adriaanm gives an overview of the compiler performance work happening in the 2.13 timeframe: http://www.scala-lang.org/news/roadmap-2.13.html
* @retronym and @lrytz are building new compiler benchmarking infrastructure at https://github.com/scala/compiler-benchmark
* some @retronym's recent speedup attempts are in https://github.com/scala/scala/pull/5785

rorygraves:
Fair enough - things are rapidly changing in this area - when I was originally thinking of this item it was before the 2.13 roadmap and work was as public (esp @retronym s recent fixes.
I still think there is room for more tooling in this area that is a student sized project, but I agree on the complexity of finding individual changes.  Happy to see the focus on performance that is happening (and happy to be involved)

PR #619, opened 2017-03-14
Scala.js support for Scastie is almost done

MasseGuillaume:


julienrf:
Maybe the project idea was not clear but it was not just about having Scala.js support in scastie. It was about providing tools to make it easy to write creative (with images, shapes, animations, or sounds) and interactive programs, easy to embed, etc. Maybe something a bit like [woofjs](http://woofjs.com/teach.html).

SethTisue:
should this be merged?

julienrf:
Actually, no :)

PR #626, opened 2017-04-01
Fixes to Skala blog links.

propensive:
Quick fix; no review.


som-snytt:
I guess I'm off the hook this April. Instead of "fully-functional" (which is controversial in the FP sense) you could say "fully operational", which also sounds funnier in a German accent, like the death star. 

"skalac" is my alias for my working scalac, so I'll have to come up with another name. Probably just `skol`.

som-snytt:
So actually, I said: https://gitter.im/scala/center?at=58df663cb52518ed4dcb5a74 b/c had to say something.

Issue #627, opened 2017-04-01
Downlaods over HTTP?


lol768:
Currently scala-lang.org links this URL:

`http://downloads.lightbend.com/scala/2.12.1/scala-2.12.1.tgz`

But, `downloads.lightbend.com` supports HTTPS. Why are the download links given using HTTP?

Related: #463

jarrodwb:
Good point.

`grep -r 'downloads.lightbend.com' . | wc` gives `336` results. Visual inspection shows that they are all http calls.

It would not be to hard to put together a PR for this.

lol768:
> It would not be to hard to put together a PR for this.

Agreed, see #628. I don't want to treat the symptom only though - where do these links come from? Are they generated automatically? Can I fix the source too?

SethTisue:
> where do these links come from? Are they generated automatically? Can I fix the source too?

they come from https://github.com/scala/make-release-notes; a PR on this in that repo would be very welcome

a PR addressing https://github.com/scala/make-release-notes/issues/35 would also be very welcome; that would address the root cause of it not being obvious where the files come from

ashawley:
Does this script need to be updated, too?

https://github.com/scala/scala-dist/blob/2.12.x/scripts/jobs/release/website/archives

It's not clear to me what it does, exactly, but it has a lightbend URL, as well.

SethTisue:
> Does this script need to be updated, too?

That would be nice. It's only using that URL to retrieve things, but it might as well use the more secure form.

As for the purpose of that section of the script, I don't recall having dealt with it before, but it takes the artifacts available at downloads.lightbend.com and also makes them available at http://www.scala-lang.org/files/archive/. (Digression: I'm not sure if that's just, like, a backup in case an asteroid hits Lightbend, or what? In the past, download pages [like this one](http://www.scala-lang.org/download/2.11.0-M5.html) used the scala-lang URLs instead of the typesafe/lightbend URLs. Maybe lightbend.com has better connectivity or cheaper bandwidth or something.)

SethTisue:
> That would be nice.

Maybe it's more that nice, maybe it's really a good idea. If the scala-lang.org download URLs aren't usually used, this would be a really bad place to fall prey to MITM.

lol768:
Thanks all for your input on this.

I think the only thing I haven't addressed here is:

> a PR addressing scala/make-release-notes#35 would also be very welcome; that would address the root cause of it not being obvious where the files come from

[...]

PR #628, opened 2017-04-02
Change all HTTP links to HTTPS

lol768:
Since no redirect or HSTS is in place, these downloads are
vulnerable to being MitM'd. This commit changes all relevant
links to use "HTTPS" in the protocol part instead of "HTTP".

Ref #627


SethTisue:
thanks for noticing, and for the fix!

PR #629, opened 2017-04-03
link to scala-user-groups list, link to scala/job-board room

SethTisue:


jvican:
Thank you @SethTisue.

PR #638, opened 2017-04-20
Update scala symposium

b-studios:
Due to a rescheduling the dates of the symposium are no longer correct and require an update.

Review by @heathermiller ?


b-studios:
Thanks Seth!

Issue #641, opened 2017-04-29
Concurrency & Distribution example uses deprecated method


robinske:
On the homepage in the `Scala in a Nutshell` section, it references the `future` method which has been deprecated.

https://github.com/scala/scala/blob/cc7acdde787fe32ba5894c35d1019963ed094c87/src/library/scala/concurrent/package.scala#L118

PR #642, opened 2017-04-29
Remove deprecated future in concurrency example

robinske:
Addresses https://github.com/scala/scala-lang/issues/641


travissarles:
@robinske thanks!

PR #606, opened 2017-02-27
Scala Lang website redesign

dominv:
This PR contains a changes related to design, layouts and responsive support. We've been coordinating with Heather and the rest of the Scala Center team during the different stages of this work, receiving feedback from their end from wireframe design to the actual implementation in Jekyll.

Here is a summary of the main changes code-wise:

- All static content (i.e.: courses, trainings, events...) was changed from categorized blogs to Jekyll custom collections. In the previous version, categorized blogs were used to hold this data (as empty posts with a YAML front matter). As part of this work, the blog and the news section were condensed into one categorized blog, and we took this chance to move that content to independent custom collections as it made more sense and simplified the blog listing code. This involves moving the previous content to different folders (Jekyll custom collections live in folders named as the collection but with an underscore prefix), that's the main reason this PR has so many changed files: a large part of them are simply moved to different places.

- As the main part of the work was related to design, new layouts were created (and the old were removed to avoid confussion). Part of the static information that was hardcoded in some of these was moved to either custom collections (i.e.: online courses), YAML variables inside each page markdown file, or YAML data files (under the _data directory) where it did make sense, to help making this data easier to handle and update.

@heathermiller @MasseGuillaume could you please review it?

Thanks!

PR #648, opened 2017-05-17
Update README.md

travissarles:


README.md
```diff
@@ -22,11 +22,15 @@ want anything else installed system-wide.
...
-
-Start the server in the context of the bundle:
-
-    bundle exec jekyll serve
+Install ruby and rbenv. `cd` into the directory where you cloned this repository, then execute the following:
+```
+rbenv install 2.4.0
+rbenv local 2.4.0
+rbenv rehash
+gem install bundle
```
ashawley: Does the command `rbenv shell 2.4.0` do the same thing?  I never knew it existed until recently.  Works well, for me.

ashawley:
Ruby 2.4.0 doesn't work for me, or is this in advance of the reworked version of the site?

heathermiller:
Turns out that we use Ruby 2.3.1 on the server. There's apparently no harm in keeping on version 2.3.1 – Jekyll doesn't require 2.4.0 or anything. Fabien asked that we not upgrade, yet.

That means changing deps in #606 not to require Ruby 2.4.0. I can do that. That means that these instructions would then be slightly inaccurate.

heathermiller:
@travissarles tweaked the README, assuming #606 will be updated

PR #649, opened 2017-05-19
blog post about scastie release

MasseGuillaume:

PR #650, opened 2017-05-19
Masse guillaume scastie

heathermiller:


MasseGuillaume:
rebased in https://github.com/scala/scala-lang/pull/649

PR #651, opened 2017-05-23
Add blogpost: Tribulations of CanBuildFrom

julienrf:
The plan is to publish the blogpost next Monday.


julienrf:
This is how the diagram looks like locally:

image

milessabin:
Perhaps I'm being picky, but why use "Kind polymorphism" as the heading of a section which doesn't further discuss kind polymorphism (either [as implemented](https://github.com/typelevel/scala/blob/typelevel-readme/notes/typelevel-4.md#minimal-kind-polymorphism-pull5538-mandubian) in Typelevel Scala or elsewhere) and which illustrates without comment exactly the sort of arity-specific boilerplate which kind polymorphism in some form could eliminate?

julienrf:
@milessabin Actually I think “Kind polymorphism” is probably not an appropriate heading.

Also, unfortunately, in practice the way kind-polymorphism is implemented in Scala does not really make things simpler:

~~~ scala
trait MapDef[A, B, CC <: AnyKind] {
  type Out
  def map(in: In)(f: A => B): Out
}

trait IterableOps[A, CC <: AnyKind] {
  def map[B](f: A => B)(implicit mapDef: MapDef[A, B, CC]): mapDef.Out = mapDef.map(this)(f)
}

trait Map[K, V] { … }

object Map {
  implicit def mapDef[K, V, L, W]: MapDef[(K, V), (L, W), Map] { type Out = Map[L, W] } = …
}
~~~

milessabin:
@julienrf I think changing the heading is the best bet.

Yes, the minimal kind polymorphism in TLS isn't up to this job, but I think it's worth exploring what can be done with something quite that minimal, and I think it would be worth exploring more sophisticated implementations.

blog/_posts/2017-05-29-tribulations-canbuildfrom.md
```diff
@@ -0,0 +1,193 @@
...
+def map[B, That](f: Char => B)(implicit bf: CanBuildFrom[String, B, That]): That
+~~~
+
+When the implicit `CanBuildFrom` parameter is resolved it fixes the return type `That`.
+The resolution is driven by the actual `B` type: if `B` is `Char` then `That` is fixed
+to `String`, otherwise it is `immutable.IndexedSeq`.
+
+The drawback of this solution is that the type signature of the `map` method looks cryptic.
+
+In the new design we solve this problem by defining two overloads of the `map`
```
szeiger: You may want to point out that this new design required improvements to type inference which are only available in 2.12.

blog/_posts/2017-05-29-tribulations-canbuildfrom.md
```diff
@@ -0,0 +1,193 @@
...
+def map(f: Char => Char): String
+def map[B](f: Char => B): Seq[B]
+~~~
+
+Then, if you call `map` with a function that returns a `Char`, the first overload is
+selected and you get a `String`. Otherwise, the second overload is selected and you
+get a `Seq[B]`.
+
+Thus, we got rid of the cryptic method signatures while still supporting the feature
+of returning a different type of result according to the type of the transformation function.
```
szeiger: but we lose the ability to abstract over arbitrary `map` implementations in a unified way

szeiger:
You may also want to mention `breakOut` as a separate usecase.

blog/_posts/2017-05-29-tribulations-canbuildfrom.md
julienrf: I think this point is not very important from the point of view of users. That’s more related to our internal implementation.

julienrf:
About `breakOut`: this is an escape hatch in case the implicit resolution of CBF does not give the expected type. So, it’s not really a “feature” but a workaround related to the way CBF works.

szeiger:
I disagree. `breakOut` is a feature, not a workaround. It allows you to build whatever target collection type you need without copying from an intermediate collection. Can we replace *all* current use cases by Views? I checked for some obvious cases (like `groupBy`) where you would return some type that wraps `That` (instead of returning `That` directly) but they don't support `CanBuildFrom`. Are there others that do? These would be the cases that can't work with Views.

julienrf:
OK, I added a section about `breakOut`. WDYT?

blog/_posts/2017-05-29-tribulations-canbuildfrom.md
```diff
@@ -190,6 +190,54 @@ could be supported even without implicit builders: you could just use an existin
...
+We can fix this issue by using the `breakOut` escape hatch:
+
+~~~ scala
+val xs: List[Int] = 1 :: 2 :: 3 :: Nil
+val xsWithSquares: Map[Int, Int] =
+  xs.map(x => (x, x * x))(collection.breakOut)
+~~~
+
+`breakOut` selects a `CanBuildFrom` instance irrespective of the initial collection type.
+In our case the `Map[Int, Int]` type annotation fixes the target type of the builder
```
szeiger: How about "This requires the target type to be known, in this case via an explicit type ascription"?

Issue #646, opened 2017-05-07
No link from blog posts/news items to list of 


Blaisorblade:
Try to click somewhere in https://www.scala-lang.org/news/releases-1Q17.html to get to https://www.scala-lang.org/news/.

Or try to get from https://www.scala-lang.org/blog/2017/02/27/scalafix-v0.3.html to https://www.scala-lang.org/blog/ by links.

I just noticed there's a link snuck at the bottom right in the end (after three times I've been on the verge of filing an issue). Could we have something more prominent at the top? I feel it's a small but not insignificant usability issue.

heathermiller:
Now fixed in redesign (blog link on the top of every page)

PR #657, opened 2017-05-31
possible fix for #646

jarrodwb:
This fixes the issue. 

I am thinking that blog, news, and announcements should all have their own index page. Right now the blog posts and news are mixed together and the announcements and news are also mixed together.
See http://www.scala-lang.org/news and http://www.scala-lang.org/blog.

It may be a good idea of go through and redo the page tags on the old posts and then redo the index pages. Maybe add an index for announcements.


SethTisue:
@jarrodwb I'm closing this on the presumption that the new design at http://scala-lang.org/blog/ is okay, but if you disagree, we could take it up again

PR #659, opened 2017-05-31
Use consistent transliteration of my name

DarkDimius:


DarkDimius:
Rebased over your changes @heathermiller

som-snytt:
I thought `Dotty` was diminutive of `Dmitry`, like the names in Tolstoy that are so hard to keep straight, `Kolya`, `Kostya`, etc.

PR #660, opened 2017-05-31
Remove Scastie from the dotty announcement

smarter:
Various dotty features are not working in Scastie because the scala.meta
parser does not support
everything (https://github.com/scalacenter/scastie/issues/217), this is
confusing people so it seems best to not mention Scastie in the
announcement.

/cc @MasseGuillaume

### DIFF
--- a/blog/_posts/2017-05-31-first-dotty-milestone-release.md
+++ b/blog/_posts/2017-05-31-first-dotty-milestone-release.md
@@ -29,8 +29,8 @@ all be part of regular time-based releases.
 
 ## How can you try it out?
 
-You have several alternatives: use the `sbt-dotty` plugin, get a standalone
-installation, or try it online on [Scastie].
+You can either use the `sbt-dotty` plugin or get a standalone
+installation.
 
 ### sbt
 Using sbt 0.13.13 or newer, do:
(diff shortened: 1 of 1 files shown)


MasseGuillaume:
Let me disable the worksheet mode in Scastie right now and let's close this PR.

som-snytt:
Is it Scastie like nasty or like pastie?

MasseGuillaume:
@som-snytt Scastie is amazing :).

DarkDimius:
@smarter, the example from https://github.com/scalacenter/scastie/issues/217 seems to work now.

DarkDimius:
Scastie runner is down....

MasseGuillaume:
@smarter @DarkDimius I just redeployed Scastie disabling the worksheet mode.

MasseGuillaume:
see https://github.com/scalacenter/scastie/commit/bffaa40839ef14ac4ea9dfa8d6fd05993324b182

DarkDimius:
@MasseGuillaume, I don't know scastie codebase to evaluate this commit.
Let's be optimistic and keep it for now, I've tried an example with enums and an example with implicit-by-name, both worked.
In case we'll see any other reports that Scastie doesn't work, we may reconsider.

@smarter, thanks for tracking feedback and reacting fast to it. 
@MasseGuillaume, thanks for an amazingly fast fix & redeploy.

PR #664, opened 2017-06-06
Replace Disqus with Discourse

travissarles:


heathermiller:
I think the blog posts that already had disqus comments enabled should probably keep the disqus comments. Only because most of the time, these are language design discussions; so it can be useful to keep the conversation around. 

(This isn't always true for questions on documentation though. Questions/discussions on documentation have a tendency to become obsolete, so that's why I think it makes sense to drop disqus completely on docs.scala-lang but maybe not on Martin's blog posts about new language designs...)

PR #667, opened 2017-06-09
47 deg's redesign + some minor reorganization

travissarles:
To see a live preview, you'll have to click on these links individually because Jekyll doesn't work well on forks of repos and links are broken.

https://travissarles.github.io/scala-lang/
https://travissarles.github.io/scala-lang/blog/
https://travissarles.github.io/scala-lang/download/
https://travissarles.github.io/scala-lang/community/
https://travissarles.github.io/scala-lang/contribute/


heathermiller:
I'd like to suggest some rewordings throughout the website, though the diff is too big to select individual files to make comments on. So it's probably best to make changes as follow-up commits.

A couple of questions though–I understand that the links are broken, but something weird is happening on the front page with blogs and events. That seems like it'd be unrelated to the linking issue, no? E.g., 
screen shot 2017-06-09 at 2 09 09 pm

There are also no blog articles appearing in the blog list. That could have something to do with linking though, I suppose.

travissarles:
Yes, I think the missing blog posts is an issue with linking. The problem is that Github pages uses travissarles.github.com/scala-lang/ as the baseurl but it's not getting picked up at generation time so all of the links are missing /scala-lang.

travissarles:
I can ask 47 Degrees how they got it running on Heroku.

heathermiller:
You can set the `baseurl` field in the _config.yml file so it generates correctly on travissarles.github.com

travissarles:
Yes, I tried that and it didn't seem to work. I'll try again in case there was something else going on.

heathermiller:
@jarrodwb: the biggest issue I see in this PR is that the blog/events/etc are mixed up/not working. If you'd like to have a look, that'd probably be the first place to start. You could probably just make a PR to @travissarles's `master` branch fork of scala-lang since he's working on master anyway.

PR #668, opened 2017-06-11
Update docs to refer to new property in VersionUtil

estsauver:
The docs currently refer to a script that doesn't exist anymore that was used to calculate the current sha. Since that was refactored and moved into the VersionUtil, this change refers to a newly exposed sha property.

### DIFF
--- a/contribute/hacker-guide.md
+++ b/contribute/hacker-guide.md
@@ -177,7 +177,7 @@ Listening for transport dt_socket at address: 8002
 // skip docs for local publishing
 publishArtifact in (Compile, packageDoc) in ThisBuild := false
 // set version based on current sha, so that you can easily consume this build from another sbt project
-baseVersionSuffix := s"local-${Process("tools/get-scala-commit-sha").lines.head.substring(0, 7)}"
+baseVersionSuffix := s"local-${VersionUtil.gitProperties.value.sha}"
 // show more logging during a partest run
 testOptions in IntegrationTest in LocalProject("test") ++= Seq(Tests.Argument("--show-log"), Tests.Argument("--show-diff"))
 // if incremental compilation is compiling too much (should be fine under sbt 0.13.13)


estsauver:
scala/scala#5937 is the PR to add this feature into the compiler. Presumably, this should hold until the other change has been accepted.

contribute/hacker-guide.md
```diff
@@ -177,7 +177,7 @@ Listening for transport dt_socket at address: 8002
 // skip docs for local publishing
 publishArtifact in (Compile, packageDoc) in ThisBuild := false
 // set version based on current sha, so that you can easily consume this build from another sbt project
-baseVersionSuffix := s"local-${Process("tools/get-scala-commit-sha").lines.head.substring(0, 7)}"
+baseVersionSuffix := s"local-${VersionUtil.gitProperties.value.sha}"
```
lrytz: It seems you need to fully qualify `scala.build.VersionUtil`. Also, it's a good idea to start the suffix with `bin`, so that the version is considered binary compatible by sbt. For example, `2.12.3-bin-what-ever-you-want` has binary version `2.12` in sbt. So i'd suggest something like

```
baseVersionSuffix := s"bin-${scala.build.VersionUtil.gitProperties.value.sha}-local"
```

estsauver: Great, I'll make this change. Kind of bizarrely, I didn't need to fully qualify the scala.build.VersionUtil locally. Is that a convention thing or did it break in your environment?

lrytz: Sorry, I missed the notification. It didn't work for me without the full name. Could you give it another try, maybe on a clean checkout?

estsauver: Sure, can do.

SethTisue:
@estsauver interested in picking this back up...?

SethTisue:
closing for inactivity. would someone else like to pick this up...?

PR #669, opened 2017-06-17
Extend partest guide with sbt and note removals

janekdb:
Based on comments by @hrhino and @martijnhoekstra.

### DIFF
--- a/_includes/sidebar-toc.html
+++ b/_includes/sidebar-toc.html
@@ -6,7 +6,7 @@
 
 		</div>
 		<hr>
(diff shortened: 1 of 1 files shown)


contribute/partest-guide.md
```diff
@@ -3,15 +3,35 @@ layout: page
...
+
+You can get a summary of the usage by running `partest --help`.
+
+If you would like to run particular tests pass the test paths as arguments
+
+        > partest test/files/pos/bounds.scala test/scaladoc/run/diagrams-base.scala
+
+To run only the Scaladoc tests use `--srcpath` with the location of the tests
+
+        > partest --srcpath scaladoc
```
hrhino: You could also mention that `partest pos` works, and executes all of the `pos` tests (likewise for `neg`, `run`, `jvm`, &c.).

janekdb: Definitely has a benefit but I'm happy to leave a fuller update to a separate PR. This is about plugging a gap related to 2.12 and sbt rather than documenting partest.

PR #670, opened 2017-06-20
Update bug reporting links

janekdb:


janekdb:
* contributors.scala-lang.org instead of Google Groups
* GitHub instead of Jira

heathermiller:
Thank you!! This is great. We really appreciate it ❤️

PR #677, opened 2017-07-12
dotty 0.2 announcement: Working Scastie link

smarter:
See https://github.com/scala/scala-lang/pull/661 for why this is needed.


heathermiller:
Merci!

PR #678, opened 2017-07-12
remove location.href-redirect and let event take its course

### LINKED ISSUE #644: Can't ctrl-click/Cmd-click on links
I often open links in new tabs by Cmd-click on OS X. That typically fails on scala-lang.org. Take https://www.scala-lang.org/contribute/documentation.html#updating-scala-langorg
Strangely, the link to http://docs.scala-lang.org/contribute.html works OK, but the one to https://github.com/scala/scala-lang doesn't.

Probably some JavaScript interfering? Maybe related to 
https://github.com/scala/scala-lang/issues/91#issuecomment-235873979

megri:
Resolves #644. Tested locally by building the site with Jekyll.

Potential issues: google analytics track-events of the type 'Downloads' (?) *may* be lost for some clicks.

### DIFF
--- a/resources/js/main.js
+++ b/resources/js/main.js
@@ -451,12 +451,7 @@ _gaq.push(['_trackPageview']);
 })();
 var f = function(event){
   var href = $(this).attr('href');
-  var target = $(this).attr('target');
   _gaq.push(['_trackEvent','Downloads','Download',href]);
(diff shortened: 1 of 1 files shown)


heathermiller:
Thanks!

PR #693, opened 2017-07-13
Less strict about version of ruby 2.3

ashawley:
I have 2.3.3 installed, and 2.3.4 is the latest, so might as well use the `~>` operator to specify Ruby version.  Otherwise,

```
$ bundle
Your Ruby version is 2.3.3, but your Gemfile specified 2.3.1
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.1'
+ruby '~> 2.3'
 gem 'jekyll'
 gem 'kramdown'
 gem 'rack-jekyll'


ashawley:
This doesn't affect the `Gemfile.lock` which is a good thing, IMO

heathermiller:
Nah, it's really better to be exact with version numbers. We have a lot more dependencies now. Less possibilities for random people to fail to reproduce the build when we stick to exact version numbers.

In general, it's best practice to use RVM. So I'd suggest grabbing it if you don't have it. Switching Ruby versions with RVM is a one-liner.

ashawley:
> We have a lot more dependencies now.

Sure, but you have a `Gemfile.lock` so no possibilities for random people to fail to reproduce the build.

Issue #688, opened 2017-07-13
improve list of Gitter rooms on front page


SethTisue:
currently:

<img width="792" alt="screen shot 2017-07-13 at 10 26 40 am" src="https://user-images.githubusercontent.com/161079/28171244-c7937934-67b5-11e7-944c-6f6cd9aa0968.png">

the first three are good choices, but then it goes south. @jducoeur suggested including the scala-js room, for example

heathermiller:
Please make a list and I can add them. These are merely placeholders.

jducoeur:
A few opinions:

I think scala-js/scala-js is totally appropriate here: it's a very large and open user-centric group, increasingly central to the ecosystem.  It's a good group for people to hit early on.

Possibly also scala-native, but that's still in-development enough that I don't know whether it's appropriate yet.  (I'd probably recommend asking the folks there whether they want to be listed.)

Also possibly scala/job-board, which seems to be developing steady if modest traffic -- helps underline that there are jobs in this technology.

The pickling and scala-xml groups are low-volume and extremely specialized; I don't think they're ideal for here, because they look a little moribund if you just drop in.  (And the link for scala-parser-combinator is broken.)

scala-india is interesting, but so low-volume that, again, I sort of think it doesn't put our best foot forward.

I'm tempted to suggest typelevel/cats, which is an increasingly significant part of the ecosystem, and is a lively and relatively friendly discussion.  Your call on whether that makes sense for this page or not.

SethTisue:
There is already a list at http://www.scala-lang.org/community/, in the Chat Rooms section: scala/scala, scala/center, scala/contributors, scala/moocs, spark-scala/Lobby, scala/job-board.

I'd suggest using this same list, but adding scala-js, since it's nearly 1.0 now.

Heather, you would know best if scala/moocs should be included or not.

I think it's better to keep the list on the site front page short. Pretty much only newbies will use the list — it's better not to overwhelm them. (So, no scala-native or dotty.)

And for the same reason I _don't_ suggest we link to specific rooms on e.g. Cats or Akka...

...but I _do_ suggest we make an exception for Spark, because the volume of beginner questions involving Spark is _really_ high, as a glance at Stack Overflow will indicate.

To summarize, my proposed list: scala/scala, scala/job-board, scala/contributors, scala/center, spark-scala/Lobby, scala-js/scala-js. and possibly scala/moocs.

jducoeur:
[...]

PR #696, opened 2017-07-14
Added Cats gitter channel.

milessabin:
The header is driven by `_data/chats-forums.yml` which is being discussed in https://github.com/scala/scala-lang/issues/688.


heathermiller:
Merci!

PR #715, opened 2017-07-25
Extend copyright into 2017

Philippus:


heathermiller:
Thanks @Philippus! 

I would have merged this, but recently @jiminhsieh managed to set the copyright date when the site is generated, so we don't have to manually change it anymore. Thanks a lot though for noticing this and taking the initiative to update it! We really appreciate it 😊

PR #716, opened 2017-07-26
Fix example dependency to add to build.sbt

bibliotechy:
Following the example in the [Adding a Dependency](https://www.scala-lang.org//documentation/getting-started-sbt-track/getting-started-with-scala-and-sbt-in-the-command-line.html#adding-a-dependency) section of 'Getting started with scala and sbt on the command line' leads to an error because the `libraryDependencies +=` part of the line was left off of the example line to add to the `build.sbt`. 

This PR fixes that.


SethTisue:
thank you for this PR which somehow fell through the cracks, sorry about that! it got fixed anyway in https://github.com/scala/scala-lang/pull/737

PR #718, opened 2017-07-31
Add ScalaFiddle to the Downloads page

lihaoyi:
While not as flexible in terms of SBT/scalaVersion configuration as Scastie is, and only running on Scala.js, ScalaFiddle is really fast (compile times typically 0.2-0.3 seconds) and allows cool graphical/interactive code snippets like those linked to in the diff.

Not sure what other people think of this diff, but if we're including Scastie and Ammonite I think ScalaFiddle definitely deserves a seat at the table.


MasseGuillaume:
Indeed, ScalaFiddle deserves to be there too.

PR #726, opened 2017-08-14
Fix link to 'API Docs (other versions)

raboof:
And change the language specs link to https


heathermiller:
Oops, I missed the notification for this! Thank you **so much** for this! ...I thought I'd tested the other versions link and it seems I didn't, doh!

Thanks again :)

PR #731, opened 2017-08-26
Fix release notes link in 2.13.0-M2 announcement.

TJSomething:
Looks like the Markdown got switched around on the link.


heathermiller:
👍 ❤️ 
Thank you for catching this!!

heathermiller:
Thanks so much :)

PR #733, opened 2017-09-05
remove implicit.ly

xuwei-k:


heathermiller:
Aww  :(
This is so sad

heathermiller:
Thanks @xuwei-k for the fix and bringing this to my attention :(

PR #739, opened 2017-09-07
Fix documentation for blogs

allanrenucci:


heathermiller:
Thanks for the fix. Updating this must've fallen through the cracks in the scala-lang redesign.

Regardless, before you make a PR to this repository, especially for something big like a blog article, please build the site locally first and double-check that the thing you're building looks right. 

...You wouldn't make a PR to Dotty or some other piece of software without first checking that it compiles and runs, right?

PR #745, opened 2017-09-11
Fix gif link for tab completion.

olafurpg:
OSX is case insensitive


heathermiller:
Sorry, we contended for the same fix! :(

PR #754, opened 2017-09-28
Fix training feed

### LINKED ISSUE #698: Training feed on front page broken

ashawley:
Fixes #698 

I don't think the training feed is broken, but I've added an upcoming training to prove it still works, and updated the README while I was at it.

### DIFF
--- a/_includes/events-training-list-top.html
+++ b/_includes/events-training-list-top.html
@@ -56,8 +56,8 @@
 							</div>
 						{% endif %}						
 						<div class="training-text">
-							<h4>{{event.title}}</h4>
-							<p>{{event.location}}</p>
+							<h4>{{event.title | upcase}}</h4>
+							<p>{{event.where}}</p>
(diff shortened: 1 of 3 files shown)


SethTisue:
thanks!

heathermiller:
Thanks.

The issue seems to be fetching from Lightbend's training feed... I haven't had the time to look into it, but this is still "broken" in that we're not getting events from Lightbend.

SethTisue:
I'll let somebody at Lightbend know

SethTisue:
fixed in 35c0bb2143247f35c051239fd88ad994c39583eb

heathermiller:
Call me crazy, but it still seems broken to me. None of those feeds from the Lightbend training feed seem to be coming through: 
screen shot 2017-10-04 at 8 56 00 am

SethTisue:
let's continue on #698

PR #756, opened 2017-10-02
don't paginate events page

### LINKED ISSUE #710: events page is paginated but doesn't need to be
http://www.scala-lang.org/events/

I have to click through to page 2 to find out that Scala Exchange is coming in December

screen shot 2017-07-19 at 9 26 02 am

I had visited the page several times before I even realized that there was a December 2017 event it wasn't showing me

I think we know from experience that the number of upcoming events is never so large that there's a need to paginate.

SethTisue:
fixes #710

### DIFF
--- a/events/index.md
+++ b/events/index.md
@@ -5,6 +5,7 @@ permalink: /events/
 
 # Pagination
 paginate:
+  per_page: 9999
   collection: events
   permalink: /:num/
   discardPastItems: true


SethTisue:
I'm going to go ahead and merge this since I want scala_lang to tweet about upcoming events (last tweet on that was in April), but I'll respond promptly to any post-merge feedback

heathermiller:
👍 LGTM! Thanks Seth!

PR #765, opened 2017-10-17
Tweak how sbt is written on the homepage

dwijnand:


SethTisue:
the output of `git grep -l SBT` includes some other pages that should probably get the same treatment:

```
_includes/scastie.html
documentation/getting-started-intellij-track/building-a-scala-project-with-intellij-and-sbt.md
documentation/getting-started-intellij-track/getting-started-with-scala-in-intellij.md
```

dwijnand:
Some of those are references to how they're spelt in IntelliJ, so I left them as is; some others I changed. Thanks for checking.

som-snytt:
I'll have to edit all my blog posts where I say `S.B.T.`.

dwijnand:
Spelt is common outside the US.

som-snytt:
@dwijnand Also common for us from the eastern seaboard. I keep forgetting to add #notironic. I removed the extra period from my previous comment. I'll also add that I had once hoped we might wind up at the tool called SOBT, b/c kind of an SOB.

SethTisue:
it's important to get these sbtleties right

som-snytt:
@SethTisue Damn you're good.

PR #767, opened 2017-10-25
Restore google analytics

adriaanm:
With the site refresh, main.js was no longer included.
Moved the analytics logic to a separate file, and
include it again.


adriaanm:
Thanks!

PR #768, opened 2017-10-26
remove Scalawags from community page

SethTisue:
on the one hand, the podcast isn't officially terminated... but on the
other hand, we haven't done a new episode in quite a long time now.
better not to link to something stale, I think. the link could always
come back if we start up again.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -96,7 +96,6 @@ Official:
 Community:
 
 * [Scala Times](http://scalatimes.com) weekly Scala newspaper
-* [The Scalawags](http://scalawags.tv) monthly podcast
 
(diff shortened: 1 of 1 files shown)


heathermiller:
Aww... Instead of removing it, should we still link to it somewhere at the bottom with a note that it's no longer running? There are still episodes posted. (Not sure if they're useful though since they're from 2016)

SethTisue:
I lean towards just removing it, but it's not a strong lean.

SethTisue:
If we do keep it, I guess we could add a note like ("on hiatus: last episode was Octember 1492") or whatever

heathermiller:
Ok, fair enough. We can remove it then :(

PR #769, opened 2017-10-30
add Scala Matsuri 2018 to events page

SethTisue:


SethTisue:
dammit, this made the events page go blank. it happened locally too, but rather than believe it was a real bug, I figured it was some problem with local previewing. fixing...

SethTisue:
this seems to be a site bug, rather than something I did wrong. if I do nothing but change the file name to something beginning with `2017-12-...`, the page stops going blank. investigating...

SethTisue:
don't know what's going on here. reverted in 2c298ed8f95d71575e86a6e4ab4460bf4fbec7a4

SethTisue:
opened #770 on whatever the problem is here

PR #772, opened 2017-11-01
Issue 770/fix cross year limitation

Philippus:
The old implementation assumes that all events in the collection take place in one year. Then the last event in the collection will be the highest 'month'.
But since Scala Matsuri 2018 is in March, the events page showed as empty.
Additionally an event in December 2018 would be mixed in with the events of December 2017.


Philippus:
ping @SethTisue

SethTisue:
outstanding! thank you very much!

PR #776, opened 2017-11-07
Fix presentation of one-day events

Philippus:
F(BY) 2017 currently shows as being from '17 Dec 2017' to '17 Dec 2017'
after this PR it will show as being on '17 December 2017'


SethTisue:
sweet! just in time for latest scala_lang tweet (https://twitter.com/scala_lang/status/927986211775328256)

PR #784, opened 2017-11-14
Issue 779/show location on events page

Philippus:


SethTisue:
super! thanks for tracking down the cause of the regression.

heathermiller:
THANK YOU! ...What an oversight, ohy. Really appreciate that you tracked this down, @Philippus! Thanks!

PR #785, opened 2017-11-15
Fix links to Hacker guide

Philippus:


heathermiller:
Thank you @Philippus!! You're really on a roll btw ❤️ 
WE REALLY APPRECIATE IT!!! 😊

PR #791, opened 2017-11-29
Update numbers from macro analysis.

olafurpg:
I must have done something wrong when counting the number "3 million"
because I can only count 2.3 million lines now and when I deduplicate
cross-built jvm/js files then it's only 2 million lines of code.

The distribution of the results does not change too much however, so I don't think too much damage was done. js+jvm macros like those in spire/minitest get less usage.

See related discussion here
https://github.com/scala/community-builds/issues/59#issuecomment-347643985


SethTisue:
thanks for following up on this

PR #804, opened 2017-12-23
Broken scala platform links

joshlemer:
* Many links were pointing at some kind of staging site, they now point at the correct SPP links
* There was a link to the first SPP meeting which was said to be in the future but is now in the past. I replaced that with a link to the Scala Process youtube channel.


SethTisue:
ping @jvican

jvican:
I have no idea how I missed this, thank you very much @joshlemer! As a footnote, you can also use platform.scala-lang.org.

joshlemer:
Cheers!

PR #805, opened 2018-01-01
Fix example for auto-rerun and rebuild with sbt

FlorianJanosch:
Fixes the proposed code changes to reflect the currently used example.


SethTisue:
thank you for the fix!

PR #807, opened 2018-01-11
add learn-fp functional programming course

dehun:


SethTisue:
can you `s/excercises/exercises`? after that, I'll hit "merge"

I was a liiiiitle hesitant about this PR when I first saw it because the other listed resources are Scala in general, whereas this is more specific. but meh, it's fine, the list isn't that long currently, and if we end up needing to add subcategories in the future, we can.

dehun:
Fixed a typo. Thanks!

PR #813, opened 2018-01-19
Update 2018.md

cvogt:


MasseGuillaume:
Thanks for adding your name!

gsoc/2018.md
```diff
@@ -60,6 +60,24 @@ for a 1.0 version.
...
+
+Supervised by [@cvogt](https://github.com/cvogt).
+
+### CBT Coursier integration
+
+Allow using [coursier](https://github.com/coursier/coursier) as the resolver in [cbt](https://github.com/cvogt/cbt).
+
+Supervised by [@cvogt](https://github.com/cvogt).
+
+### CBT Ticket hunt
```
MasseGuillaume: I would not add this project because it does not have a concrete goal. With the two tickets above, a student can approach you durring the selection process and propose annother idea related to CBT. WDYT ?

cvogt: We had this type of project for Slick twice. It was taken both times and successful. In that light I'd suggest we keep it. It gives the students the flexibility to pick what's most interesting to them which can be a positive motivational factor.

MasseGuillaume: @cvogt If it worked previously in the past, let's go for it!

PR #798, opened 2017-12-04
Multilingual chatrooms blog

propensive:
I wrote a quick blog entry to promote the French, German, Italian, Spanish and Polish Gitter channels I whimsically set up this morning...

PR #818, opened 2018-01-19
update lists of Gitter rooms

### LINKED ISSUE #688: improve list of Gitter rooms on front page

SethTisue:
fixes #688
also relevant: #798

### DIFF
--- a/_data/chats-forums.yml
+++ b/_data/chats-forums.yml
@@ -11,31 +11,16 @@ discourseForums:
 
 # Gitter channels
 gitterChannels:
-  - name: scala
+  - name: scala/scala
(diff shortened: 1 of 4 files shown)


SethTisue:
merged, since this we already discussed this at #688 pretty thoroughly and I believe the changes here reflect the consensus there

but of course further changes remain possible; and I'll respond promptly to any post-merge feedback

SethTisue:
in particular, I think the list of library/framework-specific Gitter rooms can still grow, and indeed _should_ grow.

PR #819, opened 2018-01-20
Added two rooms

wi101:


SethTisue:
thanks! I also added `cn` in d10364b4f0768697e63964034e30a87c4077bbd3

propensive:
I actually removed these two from the original blogpost as I was told that there are already larger online communities for Italian and Polish. I didn't go as far as to remove the rooms (which left them in a kind of limbo...), but I didn't want to compete with already-established groups.

Any thoughts on whether I should go ahead and delete the rooms, @SethTisue?

SethTisue:
yes, I think you should delete the rooms. if the already-established groups care to move to Gitter, they can self-organize that.

but, before you delete them, let's get those already-established groups listed in community/index.md. do you have the URLs?

wi101:
Sorry, I didn't know that.

PR #820, opened 2018-01-20
Add GSOC proposal for Bloop + test UI

Duhemm:


Duhemm:
Actually, the project is not **tied** to bloop. The idea is to use and extend the [BSP](https://github.com/scalacenter/bsp) and its implementation in Bloop, and develop a client to control Bloop using the BSP. The client should not be tied to Bloop in any way, but should be usable with any other tool that implements BSP and supports these commands. Improvements in that area could also benefit other tools (embedding test execution in an IDE for instance).

The client I'm envisioning would not sit between Bloop and test frameworks, but between the user and Bloop. Is that unclear?

olafurpg:
@propensive I would claim there is interesting metadata that is common to most testing libraries:

* number passed tests
* number of failed tests
* string output per individual test run
* running time per individual test run
* ID/name of each individual test run
* execute a selected list of IDs/name

A testing library does not have to support all of that functionality, but the most common ones like junit and scalatest have that data available. The most common way to consume this data is through a console but I personally find that interface suboptimal when your test suite grows into 1000s of tests. IDEs like IntelliJ allow you to see the results as a table, sort results by failing tests, filter out noise etc. This proposal is about making that sort of functionality available to everyone through a localhost web interface, regardless of what build tool they use or if they use an IDE or not.

olafurpg:
And to repeat @Duhemm , this is not bloop specific the idea is to use a standard protocol to communicate test reports/execution so that any build tool can make us of this. See https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#168-test-request for an example how this test protocol might look like.

jvican:
There seems to be consensus here, I'm merging since the deadline is soon. Feel free to continue the discussion in this pull request.

propensive:
Thanks for the feedback, everyone. I agree with merging on the basis of the deadline, @jvican, but would generally suggest that consensus doesn't really arrive until I've (at least) given some sort of acknowledgement... ;)

I'm generally happy with @Duhemm's explanation, if I've understood it correctly. I think I was misled by the prominence of "Bloop" in the title.

The information @olafurpg lists is useful, and I don't deny that other tools may use that information, but to reiterate (if I've not already made the point enough!), the number of tests that pass, or how long they take to run, is not information that *Bloop* would use in any way. I can't see how Bloop should make any decisions on the basis of the test output. All it should do is relay that information to the user. There's no point imposing a protocol or data structure to make clients "Bloop compatible" if all Bloop were to do is accept that structure and discard it.

Now, Bloop *could* provide the test results in a nice output that looks the same regardless of which test framework you use. But why? Why should that be Bloop's job? It's entirely orthogonal to a build loop, and the task of reporting could evolve better separately, and test frameworks or libraries would be a better place for innovation to happen in test reporting.

I agree, however, that *inputs* to a test framework are something that could be part of Bloop. I think these inputs *could* be passed as parameters to a much more general `run` command, and when developing, a "build/run-tests" loop would literally *be* a "build/run" loop for certain types of application, but I don't have any problem optimizing for the common case.

More generally, though, anything which moves Bloop towards a more integrated build system is a bit of a red flag for me. I really *like* that it's a build loop, because that's what I want it for, and I can compose it. Every step it takes towards doing more than that is a step towards becoming another integrated build system, any part of which I may not like, and any constraints it imposes make it less composable, and less useful. So this is very much a pushback towards the original goals. If the intention is to make a better build system than SBT, that's great, but make it a separate project which depends on Bloop, and keep Bloop focused on doing a single task very well (which it already seems to)!

jvican:
> Thanks for the feedback, everyone. I agree with merging on the basis of the deadline, @jvican, but would generally suggest that consensus doesn't really arrive until I've (at least) given some sort of acknowledgement... ;)
  
You're right Jon, sorry about that.

olafurpg:
@propensive "Bloop" in the title is probably misleading here, the interface between the test ui and build tool should be via a standard protocol that any build tool could hook into. An example for how that protocol might look like is here https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#168-test-request but exact details are yet to be decided

PR #823, opened 2018-01-22
GSoC: Add Scala Native bindgen proposal

jonas:
I didn't follow Scala Native development a lot lately in terms of status with respect to C interop, but I think this would be a fun project to work on for a student and something that could help lower the barrier for using Scala for native development.

/cc @densh


julienrf:
What’s missing to merge this PR?

MasseGuillaume:
Ah ops, I forgot about this one!

PR #827, opened 2018-01-27
Add Lambda World 2018

Philippus:


SethTisue:
looks good at http://scala-lang.org/events/

PR #830, opened 2018-01-31
2.13.0-M3

adriaanm:


sjrd:
Would it make sense to mention in the release notes that this version of Scala is compatible with Scala.js 0.6.22+ and 1.0.0-M3+, but not earlier versions? Otherwise I'll have to answer the question "When will Scala.js support 2.13.0-M3?" a dozen times over with the answer "It already supports it, but only those new versions".

FTR, this is because of https://github.com/scala-js/scala-js/issues/3257 which itself was caused by https://github.com/scala/scala/pull/6160.

SethTisue:
@sjrd I've added a notice to https://github.com/scala/scala/releases/tag/v2.13.0-M3, feel free to tweak it further if you like

sjrd:
Looks good to me, thanks! :)

PR #839, opened 2018-02-14
Update link and size of the Windows msi installer

Philippus:


SethTisue:
interesting, I wonder how that happened. oh well, 2.10.x series is kaput anyway

PR #846, opened 2018-02-23
Don't merge. Just a conversation.

jarrodu:
@MasseGuillaume 
Is it safe to delete this branch? It looks like it is not needed anymore.


SethTisue:
👍 to reaching Stale Branches Zero

PR #847, opened 2018-02-26
Announce edX support for our MOOCs

julienrf:


SethTisue:
I would suggest naming all of the courses, and linking to them individually, too. Technically it's redundant but it's free advertising. I actually think it's worth re-announcing the existence of these courses again (and again and again)... many people reading the blog only just tuned in.

julienrf:
Yes, good point Seth. I’ve added a commit to fix that.

SethTisue:
lgtm, one further thought: should "doesn’t deliver certificates" be expanded upon? someone just tuning in might not know what that means. if we are hoping that people want the certificates (since it's a revenue driver), that might be worth touting in more detail as well...? not sure

julienrf:
Thanks Seth for your meticulous review! I’ve added a short explanation of what a certificate is. I didn’t expand much, the goal is mainly to announce the availability of the courses on an alternative platform, for people who are not interested in certificates.

PR #852, opened 2018-03-05
Make the front-page title less jargony

lihaoyi:
The current title is pretty strange; it really doesn't mean anything to me, someone who has been using Scala for a long time, and would mean even less to a newcomer. We should dispense with the jargon and say what people would be interested in hearing, in plain language that any programmer can understand.

PR #854, opened 2018-03-10
further edits to copy on landing page

SethTisue:
this is a sequel to #852 

* restore mention of functional programming (and OO)
* add reference to type system
* tighten phrasing
* avoid saying "fast runtime" which I found potentially confusing, as I commented on #852

### DIFF
--- a/index.md
+++ b/index.md
@@ -3,7 +3,7 @@ layout: frontpage
 
 # Header texts
 headerTitle: "The Scala Programming Language"
-headerSubtitle: "Scala is a concise high-level programming language. Scala's compiler helps you avoid bugs when writing complex applications, and its fast runtime lets you build efficient, high-performance systems, with easy access to a huge ecosystem of libraries."
+headerSubtitle: "Scala combines object-orientation with functional programming into one concise, high-level language. Scala's type system helps avoid bugs in complex applications, and its JVM target lets you build high-performance systems with easy access to a huge ecosystem of libraries."
 headerButtonTitle: "Learn More"
 headerButtonUrl: "/what-is-scala/"
 
--- a/index.md
+++ b/index.md
@@ -3,7 +3,7 @@ layout: frontpage
 
 # Header texts
 headerTitle: "The Scala Programming Language"
-headerSubtitle: "Scala combines object-orientation with functional programming into one concise, high-level language. Scala's type system helps avoid bugs in complex applications, and its JVM target lets you build high-performance systems with easy access to a huge ecosystem of libraries."
+headerSubtitle: "Scala combines object-oriented and functional programming in one concise, high-level language. Scala's static types help avoid bugs in complex applications, and its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries."
 headerButtonTitle: "Learn More"
 headerButtonUrl: "/what-is-scala/"
 


tudorzgureanu:
Looks like it's almost there. 👍

lihaoyi:
Some bikeshedding:

- `combines object-orientation with functional programming` would sound better as `combines object-orientated and functional programming`. `object orientation` is a bit of an awkward phrase, and the implied `object-oriented (programming) and functional programming` I think sounds a lot more natural.

- The phrase `JVM target` seems somewhat out of place. `target` is used by compiler engineers to say what runtime their compiler `target`s, but most non-compiler engineers would simply refer to it as a `runtime`

Overall looks good to me.

index.md
```diff
@@ -3,7 +3,7 @@ layout: frontpage
 
 # Header texts
 headerTitle: "The Scala Programming Language"
-headerSubtitle: "Scala is a concise high-level programming language. Scala's compiler helps you avoid bugs when writing complex applications, and its fast runtime lets you build efficient, high-performance systems, with easy access to a huge ecosystem of libraries."
+headerSubtitle: "Scala combines object-orientation with functional programming into one concise, high-level language. Scala's type system helps avoid bugs in complex applications, and its JVM target lets you build high-performance systems with easy access to a huge ecosystem of libraries."
```
gmethvin: I think it sounds better to start with "Scala is ..." and state some of the main goals of the language, something like:

"Scala is a concise, statically typed, high-level programming language supporting object-oriented and functional styles."

jpallas: For "JVM target": the page below talks about "backends", so "JVM backend" might be more consistent.  But both "target" and "backend" are compiler jargon.  "JVM support" seems like a good compromise, since it encompasses both compiler and runtime.

yoeluk: can we change `high-level` for `general-purpose` language? https://en.wikipedia.org/wiki/General-purpose_programming_language

Jasper-M: High level and general purpose are different things.

tabdulradi: I vote for "JVM runtime" instead of "JVM target".

acjay: "[...] and it runs on the JVM, yielding high performance and easy access to a huge ecosystem of libraries."

geoffjentry:
It'd be nice to see the OO meets functional line come back. That was one of the niftier marketing bits on the original text, IMO

Sciss:
I liked "object-oriented meets functional", it's sad that it is currently gone.

dadepo:
I also feel there is nothing wrong or confusing about "object-oriented meets functional" it is nifty, succinct and gets the points across. 

In fact, it captures what the new text is trying to say. Just that now, the text is more verbose: "Scala combines object-orientation with functional programming...", that's more words for the very same thing.

LPTK:
Indeed, "Object-Oriented Meets Functional" is a nice catch phrase.
I think "JVM" is jargony, and that we should mention also the JS target (even though it is also mentioned later on in the page together with the Java target).

So how about this? With emphasis for the important keywords to flash.

> Scala – Object-Oriented Meets Functional
> ---
> Scala is a **concise**, **statically typed**, and **high-level** programming language which supports both object-oriented and functional programming. The Scala compiler helps **prevent bugs** in complex applications, and compiles your programs to **high-performance** Java or Javascript applications.
Easily specify your dependencies from its **huge ecosystem** of libraries, and you are good to go!

lihaoyi:
I am strongly against Object Oriented Meets Function, for reasons described [earlier](https://github.com/scala/scala-lang/pull/852#issuecomment-371839520). It is nice and catchy to people already in the clique, and totally meaningless to those outside. `The Scala Programming Language` is a fine title for the front-page page of the website about the Scala programming language, for people who aren't sure what Scala is in the first place.

I'm not sure if we want to put `Javascript applications` in the above-the-fold blurb, given my uncertainty around Scala.js' support status. I know Lightbend has explicitly dissociated themselves from Scala.js in the past to make it clear that this is not something they support officially.

I think the rest of @LPTK's full blurb is great but slightly too long. The reason I left of `statically-typed` was because we already say that the compiler helps you catch bugs in the second sentence, which is mostly the same meaning for all the non-Scala programmers out there. I think we should try hard to limit the blurb to two sentences given it's position on the website (I know my second sentence was kind of long-ish and run-on, but the first one being short kinda/sorta made up for it)

sjrd:
I would be in favor of putting JavaScript up there. Btw, it's written **JavaScript**, with an uppercase S. While Lightbend's management does not encourage Scala.js, the Scala Center has a Recommendation to ensure Scala.js' future, so at least it has been officialized on that side.

Unrelated: I'd put "statically typed" in there too; that's probably the number 1 thing people wonder about a language, whether it's statically or dynamically typed. It seems every language front page gives that info above the fold.

Jasper-M:
I don't think we should be underestimating potential Scala users either. In many cases `object oriented meets functional` _is_ what people are looking for in Scala.

acjay:
I agree with @lihaoyi on "Object Oriented Meets Functional". I think that 10 years ago, that was a major differentiator. But now, many (if not most) popular new programming languages have taken their cues from Scala on this front (e.g. Swift, TypeScript, Kotlin, Rust), and most programming languages from the 90s to early 2000s have grown enough features and patterns to claim they are functional (e.g. Python, Ruby, Java).

I think what actually makes Scala unique is that it is statically typed, functional, and object-oriented at the very core of its evaluation model, which makes these _three_ pieces fit together arguably better than any other language. Why is that important? Because it allows tighter abstractions to be built using the building blocks provided by the language. You can get an almost Pythonic/Ruby-like experience of application code that maps tightly to your business domain without being clogged with language boilerplate.

For the purposes of an introductory blurb, the challenge is whether that can be messaged as a concrete advantage to a programmer with no context. If it can't be, I think it is best to leave it out and dig into it where there is more space.

lihaoyi:
@LPTK here's another re-wording that I think maintains the essence but is even shorter

> Scala is a concise, statically typed, high-level programming language which supports object-oriented and functional programming. The Scala compiler helps prevent bugs, produces high-performance Java or Javascript executables, and gives you access to a huge ecosystem of libraries.

LPTK:
@lihaoyi seems good to me. Though perhaps keep the "easy", or even "effortless"? :smiley:

> The Scala compiler helps prevent bugs, produces high-performance Java or Javascript executables, and gives you [effortless] access to a huge ecosystem of libraries.

Also, I think the bold face on keywords is an important feature to make the text easily scanned by someone who would have not taken the time to actually read it in detail.

gmethvin:
@lihaoyi SGTM. Minor tweak: it should be "JavaScript" as @sjrd mentioned.

huntc:
I think it is useful te keep in mind that this page isn’t a Wikipedia entry. It’s purpose is to sell Scala IMHO. That’s something Wikipedia itself won’t let you do. 

Let’s make sure we capture the “why Scala” strongly for those we are trying to attract.

index.md
ekrich: How about Java runtime?

ekrich:
I really think the text portion stands out better as bullet items as in the original suggestion rather than a paragraph of text.

ashawley:
@fdietze asked:

> It would be interesting to know if this PR actually changes conversion rate. Is this somehow measured on the website?

I went further back on archive.org to look at the blurb on the Scala front page over the years and found it contained elements eerily similar to some of the current proposals.

**2007-2009**

> What is Scala?
>
> Scala is a general purpose programming language designed to express common programming patterns in a concise, elegant, and type-safe way. It smoothly integrates features of object-oriented and functional languages. It is also fully interoperable with Java.

**2009-2013**

> What is Scala?
>
> Scala is a general purpose programming language designed to express common programming patterns in a concise, elegant, and type-safe way. It smoothly integrates features of object-oriented and functional languages, enabling Java and other programmers to be more productive. Code sizes are typically reduced by a factor of two to three when compared to an equivalent Java application.

**2013-2018**

> Object-Oriented Meets Functional
>
> Have the best of both worlds. Construct elegant class hierarchies for maximum code reuse and extensibility, implement their behavior using higher-order functions. Or anything in-between.

Looking at Google Trends, it seems that search popularity increased most in early 2009 and late 2012. 

Google Trends 2005-2018

Not sure if the blurb can take much credit for this activity.  Probably other factors attracted people to Scala.  The incorporation of Typesafe, the publishing of Coursera courses and then Heather building out this site and the docs site are probably more noteworthy.  I'm sure there are others I'm forgetting.

jpallas:
> I'm sure there are others I'm forgetting.

The explosion in popularity of Spark is probably a factor.

SethTisue:
The "Object-Oriented Meets Functional" tagline seems to inspire both love and hate. I've attempted to honor the love by returning wording about OO and FP to a prominent position (more prominent than in Haoyi's last-suggested version), but I didn't bring the actual tagline back.

There was other good feedback on this PR that I haven't specifically incorporated. I'm hitting "merge" regardless as my overall sense of the feedback is that we've taken a clear step forward here. Followup PRs remain possible.

fdietze:
@ashawly with conversion rate I meant the fraction of website visitors who are (for example) clicking the "learn more" button.

https://en.wikipedia.org/wiki/Conversion_marketing#Conversion_rate

Atry:
`help avoid` or `help avoiding`?

Atry:
Moderate number of buzz words help SEO. I think this PR is good when having `Object-Oriented`, `Functional`, `JVM`, and `JavaScript` here.

SethTisue:
candidate buzzwords: https://twitter.com/mlrawlings/status/974823927917641728  ;-)

LPTK:
@SethTisue why does the text now say "huge ecosystem**s** of libraries," instead of "a huge ecosystem"? Doesn't it make it unnecessarily sound like the Scala ecosystem is fragmented?

sjrd:
The way I understand it is that it refers to the JVM and JS ecosystems.

lihaoyi:
@SethTisue you missed:


screenshot 2018-03-18 11 06 10

tindzk:
First off, thanks to everyone involved in improving the wording on the landing page. The latest version is certainly more meaningful to newcomers. Here are a few more ideas.

The first sentence starts with "Scala", the second with "Scala's". We could improve it as follows:

1. Its static types help avoid bugs [...]
2. Its type system help avoid bugs [...]

> and its JVM and JavaScript runtimes let you build high-performance systems with easy access to huge ecosystems of libraries

This is not entirely accurate since Scala does not ship a JVM/JavaScript runtime. If JavaScript and JVM are mentioned, the LLVM target needs to be enumerated. Two proposals:

1. and its JVM, LLVM and JavaScript support let you build [...]
2. and its JVM, LLVM and JavaScript targets let you build [...]

Instead of listing all platforms, we could shorten the phrase to the following formulation which would cover all three targets:

* and its cross-platform support lets you build [...]

sjrd:
LLVM is an implementation detail. The target is Native code.

/cc @densh, if we're going to talk about including Scala Native in there.

SethTisue:
> if we're going to talk about including Scala Native in there

@tindzk that's an issue all to itself, quite distinct from wordsmithing the marketing. could you make a separate ticket on that, please? let's not discuss it here.

SethTisue:
> if we're going to talk about including Scala Native in there

@tindzk that's an issue all to itself, distinct from wordsmithing the marketing. could you make a separate ticket on that, please? promoting Scala Native to the level of JVM and JS would require coordinated changes in multiple places.

SethTisue:
re: "Scala" vs "its", I went back and forth on that a few times actually, not sure what I think.

I don't think "cross-platform" is a good wording, since the JVM itself is cross-platform in the sense of Windows/Mac/Linux/etc.

tindzk:
@SethTisue Thanks for your feedback! Sure, I will create a separate issue regarding promoting Scala Native.

As for "cross-platform", I used the official definition from: https://github.com/portable-scala/sbt-crossproject

PR #859, opened 2018-03-19
Include a link to "Accessible Scala" proposal

ScalaWilliam:
As I wanted to read more about it, yet couldn't find any actual reference to it.


ScalaWilliam:
Maybe I'll move it to the end of the article, so it acts as a good call-to-action.

propensive:
Good point, @ScalaWilliam! Thanks!

PR #862, opened 2018-03-23
swap order of "IDEs" and "Nutshell" sections. put nutshell first

SethTisue:
I can't see any reason for the old order -- seems to me it's clearly
better to show off the language before starting to talk about what
IDEs support it? but maybe someone wants to argue otherwise

### DIFF
--- a/_layouts/frontpage.html
+++ b/_layouts/frontpage.html
@@ -104,33 +104,6 @@
   </div>
 </section>
 
-<!-- IDEs for Scala -->
-<section class="ides">
-  <div class="wrap">
-    <div class="heading-line">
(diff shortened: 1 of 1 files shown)


SethTisue:
screen shot 2018-03-23 at 12 34 08 pm

propensive:
Remarkably, I independently thought literally exactly the same thing when looking at the site just now. (I thought a couple of other things too, but they can wait.)

As nobody else has spoken up, I'm going to merge optimistically, with the option to revert if a lot of people disagree. Maybe we can re-promote it once the tooling landscape improves. ;)

PR #867, opened 2018-03-29
Include Scala Native (via LLVM) on the front page

propensive:
Also removes the "more backends..." text.

Note that I've use the LLVM logo to indicate the "native backend" - I
didn't know what else to use.


propensive:
image

densh:
@propensive Looks good!

PR #874, opened 2018-04-11
Add Curry On 2018

Philippus:


heathermiller:
Yay, thank you @Philippus! \o/

PR #877, opened 2018-04-13
Add Scala.IO 2018

Philippus:


SethTisue:
thank you! looking good at https://scala-lang.org/events

(I don't think it's necessary to include the year in the event name, but I don't have a strong feeling about it. we haven't been consistent about this, but usually the year isn't included, I think? I removed some years not long ago in b5c46754e34d2a24ad4ed73bc8c62d24a496c814.)

Issue #876, opened 2018-04-13
Concurrency sample uses deprecated `future` function


kubukoz:
```scala
val x = future { someExpensiveComputation() }
val y = future { someOtherExpensiveComputation() }
val z = for (a <- x; b <- y) yield a*b
for (c <- z) println("Result: " + c)
println("Meanwhile, the main thread goes on!")
```

from the sources:
```scala
@deprecated("use `Future { ... }` instead", "2.11.0")
// removal planned for 2.13.0
def future[T](body: =>T)
```

The `future` usage should probably be replaced with `Future { }` in this example (and its description). If you think it's a good idea, I can prepare a PR in no time :)

SethTisue:
yes please!

kubukoz:
I can see that the correct version is used in `_includes/bullet-concurrency-distribution.html` since https://github.com/scala/scala-lang/pull/642 - why are there two places for these snippets?

Anyways, I should have a PR with a fix for this within minutes :)

kubukoz:
It's ready in #878 :)

PR #878, opened 2018-04-15
Remove deprecated sample from the 'concurrency&distribution' example

kubukoz:
Fixes https://github.com/scala/scala-lang/issues/876

Related: https://github.com/scala/scala-lang/pull/642


SethTisue:
thanks!

PR #882, opened 2018-04-20
Use headings for Scala 3 questions

ashawley:
Makes the questions more prominent and makes them clickable from the "contents" menu in the sidebar.


heathermiller:
Thank you @ashawley!!

PR #891, opened 2018-05-04
Increase color contrast of page title and link.

exoego:
Addresses the issue https://github.com/scala/docs.scala-lang/issues/941 along with PR https://github.com/scala/docs.scala-lang/pull/1065

# Before

lang-before

# After

lang-after

### DIFF
--- a/_sass/utils/_variables.scss
+++ b/_sass/utils/_variables.scss
@@ -6,11 +6,11 @@
 //------------------------------------------------
 $brand-primary: #DC322F;
 $brand-secondary: #859900;
-$brand-tertiary: #5CC6E4;
+$brand-tertiary: #0080A3;
 //-------------------------------------------------
 $gray-darker: #002B36;
 $gray-dark: #073642;
-$gray: #15414C;
+$gray: #133942;
 $gray-li: #244E58;
 $gray-light: #E5EAEA;
 $gray-lighter: #F0F3F3;
@@ -53,7 +53,7 @@ $font-black: 900;
 
 // Link Colors
 //------------------------------------------------
-$base-link-color: darken($brand-tertiary, 15%);
+$base-link-color: $brand-tertiary;
 $hover-link-color: $brand-primary;
 
 // Border


jvican:
For some reason, the darker blue of the background annoys me a little bit -- I prefer the previous one. Would it be possible to have a combination of the darker blue for the font links and the lighter blue for the background?

heathermiller:
While I'm thankful for PRs that attempt to improve the readability of the website, I have to say that we hired a professional designer to provide us with the current look of the website (released about 9 months ago). I would not like to crowdsource and overwrite the decisions that we obtained from a design professional, if that makes sense. So, thank you, thank you, but for the vast majority of people, I think the color scheme is considered to be OK, and I don't want to cast aside the decisions of our designer.

exoego:
I can understand the reason.

As a contributor, it would help not to waste time if issues are timely triaged (labeld `wontfix` and closed wth reason explaind for this case). The https://github.com/scala/docs.scala-lang/issues/941 had been posted 5 monthes ago...

heathermiller:
I understand, @exoego, and I'm sincerely sorry. I never saw scala/docs.scala-lang#941, actually (until now). With all of the GitHub notifications I get, it happens that on a busy week where I'm traveling or the like, that I might miss some notifications. This is my fault, and I'm sorry.

weedySeaDragon:
@heathermiller: I truly understand the value of using a professional designer vs. crowdsourcing.  (And for sites such of this, I would probably make the same call.)   But as  scala/docs.scala-lang#941 shows, even professional designers sometimes don't get everything right on the first version.

I'd also respectfully point out that this is not just a "readability" issue.  It's more important than that:  it's an _[accessibility **standards**](https://www.w3.org/WAI/intro/accessibility.php)_ issue.  It's recognizing that it's worth meeting minimal standards so that the information on your website can be shared with and used by the widest possible group of people; being inclusive instead of exclusive.  It is about recognizing the value of the **standards**, just like recognizing the value of standards in other areas of software development.  

Accessibility standards aren't new and they aren't hard to meet.  The software industry has been really slow and resistant to adopt them even though we, of all areas, have the _most_ ability to create and use tools and methods to incorporate and meet (and exceed) them easily.  Please don't be yet another person resisting them.

I won't insert hundreds of links and quotes and references as to why accessibility is important.  A simple search will yield all of the information you need about that.  (I did list a few in scala/docs.scala-lang#941 .)

But I will say this:  Consider that as our bodies age, our abilities change, including our eyesight.  For most people, the ability to focus changes (usually gets worse),  and as the lens gets cloudy, we need more light and our ability to distinguish colors gets worse.  Accessibility standards such as WCAG 2.0  help to address this.  Accessibility standards are not just about those with extreme differences.  (Some say that most people are just "temporarily able" until their bodies start to age.) 

If you're able, I suggest you talk about this to the designer you used for the website.  I'm sure s/he will have a solution.  

And if that designer doesn't think that it's important to change the colors, then consider that they are not able create a design that meets minimal, accepted, important standards. Would you use an interior designer that could not/would not design a space that would work for the largest group of people possible?  A designer that would ignore industry accepted standards about accessible access?  Would you want to be associated with such a space?  Consider that there are thousands of designers out that that can.

Issue #881, opened 2018-04-19
http://pgp.mit.edu/ causes intermittent htmlproofer failures


SethTisue:
perhaps we can exempt the domain from htmlproofer?

dwijnand:
Additionally (I can split this into a separate issue) https://travis-ci.org/scala/scala-lang/builds/380079726 from https://github.com/scala/scala-lang/pull/896/commits/38c24d1f47304baee8e13fafba1c57ca5bae2e3c failed with:

```
- ./_site/documentation/your-first-lines-of-scala.html
  *  External link http://java.sun.com/javase/ failed: 403 No error
- ./_site/news/2.12.0/index.html
  *  External link http://www.oracle.com/technetwork/java/javase/downloads/index.html failed: 403 No error
htmlproofer 3.8.0 | Error:  HTML-Proofer found 2 failures!
```

SethTisue:
I see at https://github.com/gjtorikian/html-proofer that a `--url-ignore` option is supported

PR #898, opened 2018-05-17
Switch a couple of URLs to HTTPS

dwijnand:
Hopefully addresses https://github.com/scala/scala-lang/issues/881#issuecomment-389791711


dwijnand:
Nope. No failures in https://travis-ci.org/scala/scala-lang/builds/380099316, but I think the error is transient anyways, so it's no proof.

So (to me) that leaves the question: should we use https over http?

PR #899, opened 2018-05-17
Show current events on the upcoming events page until they are over

Philippus:
moved the filtering logic to the layout html files to do so

### DIFF
--- a/_layouts/events.html
+++ b/_layouts/events.html
@@ -2,7 +2,16 @@
 layout: inner-page-parent
 ---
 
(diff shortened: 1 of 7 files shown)


SethTisue:
I don't speak Jekyll but you've got a good track record so let's try it :-)

SethTisue:
(I've set myself a reminder to check https://scala-lang.org/events/ on June 4 and make sure LambdaConf hasn't vanished.)

PR #906, opened 2018-06-11
ignore some domains when checking external links

### LINKED ISSUE #881: http://pgp.mit.edu/ causes intermittent htmlproofer failures

### LINKED ISSUE #905: Events not updating
Right now there's an 'old' event on the website, I think it's because it is only refreshed after a successful build. If that is the reason Travis CI could be set up to run the build daily.

SethTisue:
fixes #881
~~fixes #905~~


SethTisue:
actually I don't think this will help with #905, I'll comment there

SethTisue:
the scalafmt 404 was already fixed (https://github.com/scalameta/scalafmt/issues/1219)

PR #919, opened 2018-06-22
Fixed up the front page training feed

dylanowen:
Fixes: https://github.com/scala/scala-lang/issues/698

It looks like the logic for the training page has changed a lot. I did my best on the front page, but the /training/ page is still missing the lightbend data.

I've also deleted the main.js file. It looks to me like we moved over to functions.js but never cleaned it up.


SethTisue:
oh, I didn't notice the review-request. well, Heather can let us know if we should revert it.

pre-merge, https://www.scala-lang.org/training/ was blank, so it would hard to do worse :-)

dylanowen:
@SethTisue For this change I just focused on the front page under `Upcoming Training`. We should probably open up a new issue for https://www.scala-lang.org/training/ . I can tackle that bug as well when I have some time.

SethTisue:
opened https://github.com/scala/scala-lang/issues/921 on it

SethTisue:
👍 opened https://github.com/scala/scala-lang/issues/921 on it

thank you Dylan!

Issue #921, opened 2018-06-25
training page is blank


SethTisue:
https://www.scala-lang.org/training/ has nothing

sequel to #919, which fixed up the training section of the front page

Issue #923, opened 2018-06-30
/training page and the front page don't include "On Demand" training


dylanowen:
Lightbend training data includes "On Demand" training which we don't display on our front page or /training pages

```
sessions": Array[1][
        {
          "where": "Virtual Classroom",
          "when": "On demand",
          "trainers": null,
          "organizer": "ScalaCourses.com",
          "status": "live",
          "url": "https://www.scalacourses.com/?courseId=40&partnerCode=lightbend"
        }
      ]
```

SethTisue:
(not sure, but I wonder if it's better to just put ongoing, on-demand things on https://www.scala-lang.org/community/ and maybe add a link to that page from the training page)

PR #924, opened 2018-06-30
updated the /training page to include lightbend data

dylanowen:
Fixes: #921 

I moved the training data to it's own json file so we're not loading it for all pages that use `functions.js`

I also refactored the training functions to work for both the front page and the `/training` page.

Follow up work: https://github.com/scala/scala-lang/issues/923 - support for On Demand training

### DIFF
--- a/resources/js/functions.js
+++ b/resources/js/functions.js
@@ -321,106 +321,181 @@ function updatePointer() {
 
 // TRAININGS
 $(document).ready(function() {
-    // Stop early if the element does not exist, i.e.,
-    // we're not on the front page nor on the Trainings page
-    if ($('.training-items-list').length === 0) {
-        return;
-    }
+    var MONTH_NAMES = [ "January", "February", "March", "April", "May", "June", "July", "August", "September", "October", "November", "December" ];
+    var MONTH_NAMES_SHORT = MONTH_NAMES.map(function (month) {
+       return month.substr(0, 3);
+    });
 
-    var isFrontPage = $('.upcoming-training').length !== 0;
-    var MAX_TRAININGS = isFrontPage ? 5 : 999;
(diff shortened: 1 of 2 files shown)


SethTisue:
is there a reviewer out there who understands this stuff, at least a little, and could take a look?

dylanowen:
By the way, @SethTisue I enjoyed your talk "You Are A Scala Contributor" at Scala Days. It's what inspired me to start picking up issues 😄

SethTisue:
thanks for letting me know, that warms my heart!

SethTisue:
I tried previewing this locally but the training page was still blank. otoh, I think I recall that the training stuff never worked in local runs...? not 100% sure

I'm going to go ahead and hit merge and let's just see...

SethTisue:
hmm, https://www.scala-lang.org/training/ is still blank...

dylanowen:
Yeah, I believe local runs don't work for training, since the php scripts aren't running (as far as I can see). For development I just served up the json from a local endpoint.

https://www.scala-lang.org/training/ works for me, but has a bit of a lag. Are there any errors in your console?

Maybe we could ditch the php proxying if we could convince the owner of https://www.lightbend.com/subscription/training.json to add some cors response headers?

SethTisue:
> https://www.scala-lang.org/training/ works for me, but has a bit of a lag

oh working now, maybe I just closed the tab too quickly last time

🎉

thank you for this

SethTisue:
> Maybe we could ditch the php proxying if we could convince the owner of https://www.lightbend.com/subscription/training.json to add some cors response headers?

we have an internal ticket on this at https://github.com/lightbend/lightbend.com/issues/1158 (private link) that has now been open for 2.5 years. I pinged again just now

PR #927, opened 2018-07-12
get `bundle install` working again

SethTisue:
it was telling me:

>Downloading html-proofer-3.8.0 revealed dependencies not in the API or the lockfile (nokogiri (~> 1.8.1)).
>Either installing with `--full-index` or running `bundle update html-proofer` should fix the problem.

and Travis-CI was saying the same thing at e.g.
https://travis-ci.org/scala/scala-lang/builds/402903201

so I did `bundle update html-proofer` and these are the resulting
changes.


SethTisue:
I don't understand this Ruby stuff but Travis-CI was red before and it's green now and htmlproofer seems to still be doing its thing ("Checking 4654 external links..."), so that's good enough for me...

PR #930, opened 2018-07-17
Added support for Scala JS for the main scripts

dylanowen:
**Not Ready To Merge**

This is a partial implementation of using Scala JS for our site's scripts instead of Vanilla JS. 

I'm opening this pull request mainly to get feedback on the direction and to see if it's worth completing. I'm totally fine with tossing this code if we don't think it makes sense for the project.

**Should we actually do this?**
This is a large change with multiple moving pieces. Should we really be changing something that isn't broken?

**Other Concerns**
1. Maintainability: If our maintainers are relying on bundler vs the docker-compose scripts we'll be breaking their workflow.
2. I'm hoping this makes the code more accessible to scala-lang maintainers but I don't have any insights on if JS actually makes mores sense.
3. This makes our builds more complex.

**Missing Pieces**
1. Compiling the Scala in our build (do we use travis, jenkins, or both?)
   a. `sbt fullOptJS` I'm using the closure compiler to minify the output but we can turn that off to make prod debugging easier

 Anyway let me know if I should continue and if there is any feedback on my implementation.


docker-compose.yml
```diff
@@ -9,4 +14,5 @@ services:
       - 4000:4000
       - 35729:35729
       - 3000:3000
-      -   80:4000
\ No newline at end of file
+      -   80:4000
+    command: jekyll serve --incremental
```
dylanowen: I'm assuming that docker-compose is only used for development, but if this isn't the case I could be breaking things here

js-src/functions/src/main/scala/org/scalalang/utils/Logger.scala
```diff
@@ -0,0 +1,158 @@
...
+
+import org.scalajs.dom.console
+
+import scala.reflect.ClassTag
+import scala.scalajs.js
+
+/**
+  * A basic logger for the debugging console
+  */
+object Logger {
```
dylanowen: I browsed around for a few scala js logging implementation but didn't find one that fit my needs. I'm open to swapping this out for a different project vs my home-built solution

resources/js/functions.js
```diff
@@ -1,6 +1,3 @@
----
----
```
dylanowen: I believe this isn't needed after https://github.com/scala/scala-lang/pull/924

docker-compose.yml
SethTisue: correct it's optional dev-only stuff

js-src/functions/src/main/scala/org/scalalang/utils/Logger.scala
SethTisue: perhaps you could put a comment to that effect at the top of the file, and say briefly what the considerations were

sjrd:
I am skeptical. Even https://www.scala-js.org/ does not use any Scala.js code, and for one good reason: it is a mostly static website, with a couple scripts for isolated functions on the site. Most of the code consists in external libraries. The same applies to scala-lang.org.

Scala.js does not shine on small scripts. You will bloat the download times because the Scala.js code size "tax" will be much larger than the features you are implementing---unlike in larger applications, where Scala.js is appropriate because the tax is dwarfed by the actual codebase of the website.

Just look at the PR stats: +441 for -91. If you're increasing the codebase size by porting to Scala.js, you're likely doing yourself a disservice. One would say that most of it is about build scripts, and that is true. But let us also look at the features themselves. For example, this PR replaces the JavaScript code:
```js
// OS detection
function getOS() {
  var osname = "linux";
  if (navigator.appVersion.indexOf("Win") != -1) osname = "windows";
  if (navigator.appVersion.indexOf("Mac") != -1) osname = "osx";
  if (navigator.appVersion.indexOf("Linux") != -1) osname = "linux";
  if (navigator.appVersion.indexOf("X11") != -1) osname = "unix";
  return osname;
}

$(document).ready(function() {
    if ($(".main-download").length) {
        var os = getOS();
        var intelliJlink = $("#intellij-" + os).text();
        var sbtLink = $("#sbt-" + os).text();
        var stepOneContent = $("#stepOne-" + os).html()
        $("#download-intellij-link").attr("href", intelliJlink);
        $("#download-sbt-link").attr("href", sbtLink);
        $("#download-step-one").html(stepOneContent);
    }
});
```
by [a complex abstraction for the OS](https://github.com/scala/scala-lang/pull/930/files#diff-0c92759e660bdc011bc96754a88c9686) in addition to the following Scala.js code:
```scala
  private def setupMainDownload(): Unit = JsUtils.findElement(".main-download").foreach(_ => {
    val osLabel: String = OS().label

    val intelliJlink: String = $("#intellij-" + osLabel).text()
    val sbtLink: String = $("#sbt-" + osLabel).text()
    val stepOneContent: String = $("#stepOne-" + osLabel).html()

    $("#download-intellij-link").attr("href", intelliJlink)
    $("#download-sbt-link").attr("href", sbtLink)
    $("#download-step-one").html(stepOneContent)
  })
```
What have we gained? Virtually nothing. Just a little bit of type-safety, which TypeScript would have also been able to provide, without the Scala.js bloat.

I think the PR was a worthwhile experiment, if only to demonstrate that Scala.js is not the best option for this specific job.

js-src/functions/build.sbt
```diff
@@ -0,0 +1,18 @@
...
+name := "functions"
+
+libraryDependencies ++= Seq(
+  "org.scala-js" %%% "scalajs-dom" % Versions.Dom,
+  "org.querki" %%% "jquery-facade" % Versions.JQuery
+)
+
+// move our output folder to static
+artifactPath in(Compile, fastOptJS) := baseDirectory.value / ".." / ".." / "resources" / "js" / s"scala-${name.value}.js"
+artifactPath in(Compile, fullOptJS) := baseDirectory.value / ".." / ".." / "resources" / "js" / s"scala-${name.value}.js"
```
sjrd: Don't configure `fastOptJS` and `fullOptJS` to produce the same target file. It will confuse caches (`fullOptJS` after `fastOptJS` will not recreate the file, for example).

js-src/functions/build.sbt
```diff
@@ -0,0 +1,18 @@
+import org.scalajs.core.tools.linker.backend.OutputMode
```
sjrd: This import is unused.

js-src/functions/build.sbt
```diff
@@ -0,0 +1,18 @@
...
+
+libraryDependencies ++= Seq(
+  "org.scala-js" %%% "scalajs-dom" % Versions.Dom,
+  "org.querki" %%% "jquery-facade" % Versions.JQuery
+)
+
+// move our output folder to static
+artifactPath in(Compile, fastOptJS) := baseDirectory.value / ".." / ".." / "resources" / "js" / s"scala-${name.value}.js"
+artifactPath in(Compile, fullOptJS) := baseDirectory.value / ".." / ".." / "resources" / "js" / s"scala-${name.value}.js"
+scalaJSOptimizerOptions in (Compile, fullOptJS) ~= { _.withUseClosureCompiler(true) }
```
sjrd: This setting is redundant, as it is the default.

js-src/functions/build.sbt
```diff
@@ -0,0 +1,18 @@
...
+
+scalaJSUseMainModuleInitializer := true
+emitSourceMaps := false
```
sjrd: Prefer the more future-proof
```scala
scalaJSLinkerConfig ~= { _.withSourceMap(false) }
```

js-src/functions/src/main/scala/org/scalalang/utils/Logger.scala
sjrd: You have two actual uses of this logger. I suggest replacing them by a good old `println`.

js-src/functions/src/main/scala/org/scalalang/utils/OS.scala
```diff
@@ -0,0 +1,55 @@
...
+
+    val os: OS = Array(Windows, Mac, Linux, Unix)
+      .foldLeft(default)((last: OS, os: OS) => {
+        if (appVersion.contains(os.navigator)) {
+          os
+        }
+        else {
+          last
+        }
+      })
```
sjrd: More simply:
```scala
val os = Array(Windows, Mac, Linux, Unix)
  .find(os => appVersion.contains(os.navigator))
  .getOrElse(default)
```

js-src/functions/src/main/scala/org/scalalang/utils/OS.scala
```diff
@@ -0,0 +1,55 @@
+package org.scalalang.utils
+
+import org.scalajs.dom.window
+
+/**
+  * Find which OS our client is running
+  */
+sealed trait OS {
+  val navigator: String
+  val label: String
```
sjrd: In order to cut down on the boilerplate, consider declaring `OS` as
```scala
sealed abstract class OS(val navigator: String, val label: String)
```
so that the instances are one-liners:
```scala
case object Windows extends OS("Win", "Windows")
```

js-src/functions/build.sbt
dylanowen: Oh good to know. What's the recommended way to setup a build like this? (Where I'd like my html file to be static and to just import the compiled output regardless of compilation target)

sjrd: I'm not sure. I think it depends on the web server and/or templating mechanisms that one use. The folks at https://gitter.im/scala-js/scala-js or https://stackoverflow.com/questions/tagged/scala.js might have more experience with that than I do.

dylanowen:
Thanks for the feedback and detailed comments @SethTisue and @sjrd 😄 

While this was a fun exercise I'm inclined to agree with @sjrd's analysis: we're better off sticking with vanilla JS.

I'm closing out this PR but I've updated my code with the feedback so if we ever decide Scala.js makes sense we'll know where to find it.

jvican:
Thanks for your work Dylan :+1:
  
Just in case you're thinking of playing around with Scala.js more, you can have a look at https://github.com/scala/docs.scala-lang/issues/1107.

Issue #701, opened 2017-07-17
autocompletion search in blog not working properly


MasseGuillaume:
up and down arrow is not selecting the previous/next item

![out](https://user-images.githubusercontent.com/921490/28263885-e3461d2c-6ae8-11e7-9cc9-3c56462a916c.gif)

heathermiller:
Ooh, good catch! Thanks :)

marnux:
The issue is situated in the Simple Jekyll Search plugin. Which is out of scope for the scala spree.
Either this needs to be fixed in the plugin or another plugin needs to be chosen with this functionality.
I made a pull request to pull in the latest version of the plugin, it at least prevents the result container from being cleared when pushing the arrow keys.

SethTisue:
verified that that the PR makes the improvement described, at least. thank you!

jvican:
Would it make sense to remove this Jekyll search and use instead the new algolia search that we use in docs?

PR #931, opened 2018-07-18
Update to latest version of Simple Jekyll Search

marnux:
This fixes the clearing of the result container when pressing the arrow keys. 
As it turned out the issue is situated in the Simple Jekyll Search plugin. Which is out of scope for the scala spree.
Either this needs to be fixed in the plugin or another plugin needs to be chosen with this functionality.


SethTisue:
references #701

Issue #933, opened 2018-07-19
new htmlproofer failures


SethTisue:
seen at e.g. https://travis-ci.org/scala/scala-lang/builds/405576488

```
- ./_site/documentation/your-first-lines-of-scala.html
  *  External link http://java.sun.com/javase/ failed: 403 No error
- ./_site/news/2.12.0/index.html
  *  External link http://www.oracle.com/technetwork/java/javase/downloads/index.html failed: 403 No error
```

SethTisue:
not sure whether these should be replaced with better links, or whether we should just exempt the domains from checking (as in #906)

PR #939, opened 2018-07-23
Update JDK link

cb372:
The previous URL is a 301 Redirect to this one, which makes htmlproofer
unhappy. Not sure why it thinks the response is a 403 - I don't get that
when I hit it with curl.


cb372:
Fixes #933

jvican:
thank you Chris, these spurious link errors happen from time to time, and we fix them as they turn up :smile:

PR #840, opened 2018-02-16
Fix #535: setup htmlproofer and fix all broken links

OlivierBlanvillain:

PR #942, opened 2018-07-26
Only check external links daily

ashawley:
To make contributing smoother, we should disable full site link checking on every PR commit.  In particular, disable external linking for commits--which is the major burden, IMO.   See [previous discussion where I suggested doing this](https://github.com/scala/scala-lang/pull/840#issuecomment-366275454). 

I've preserved external link checking as a scheduled job, which is now possible with [Travis scheduled jobs](https://docs.travis-ci.com/user/cron-jobs/).  Doing it less often will help avoid sporadic failures for external link checking that contributors experience.  Unreliable hosts results in needing to add them to a growing list of ignored domains. Good net etiquette would suggest avoid robot checking web sites on Travis for every commit.

I've also preserved internal link checking for PRs and merges.   Therefore, checking internal links is disabled in the scheduled job:  They should have been caught during the PR process or on the merge commit to master.

Downsides:

- If a PR adds a new remote URL, it won't be checked.  However, people should be checking new links manually.
- Someone needs to check if the cron job has found some bad URLs, periodically.  
- Travis config is a little hairier.

I'd say those downsides are outweighed by reducing the "red tape" contributors need to get through to contribute to the site.  External link checking seems to be side-tracking the PR review and approval process, and isn't scaling well.  For example, disabling external link checking tightens the build loop for PR commits by a few minutes when remote link checking is skipped.


ashawley:
There's usually a good crop of cron jobs to review for failures in a week:

https://travis-ci.org/scala/scala-lang/builds

Hopefully, the site admins are set to get notifications on such failures.

SethTisue:
> Hopefully, the site admins are set to get notifications on such failures

I've added myself to `.travis.yml`: 1ae55b09ab86fe75002f6917a5667bbda9346161

SethTisue:
haha and by doing so created a merge conflict

ashawley:
> I've added myself to `.travis.yml`

Good idea and thank you.

ashawley:
> haha and by doing so created a merge conflict

Shall I resolve?

ashawley:
I think only you can see those settings.

ashawley:
I thought of one other upside of this, later: Deployments will also work after a PR merge to master, regardless if there are external link checking errors.  Recently, a deployment failed because of spurious web site errors.  Being blocked from deploying like this seems less than ideal.

SethTisue:
> I think only you can see those settings.

heh

screen shot 2018-08-02 at 11 28 23 am

PR #951, opened 2018-09-05
Fix broken Github commit links

ashawley:
Travis cron jobs is complaining:

```
- ./_site/announcement/2013/05/29/release-notes-v2.11.0-M3.html
  *  External link https://github.com/scala/scala/commit/fff0f50 failed: 404 No error
```

For whatever reason, just adding an extra digit to the git sha fixes the link.

Fixed another link that had the same issue, that was likely unlinked for this reason.


Philippus:
thank you!

PR #957, opened 2018-09-29
Add SDKMAN for installation on download page.

marc0der:
[SDKMAN](https://sdkman.io) provides an easy way to install JVM related SDKs. We have supported Scala for a long time, and even allow installation of historic versions as early as 2.11.0.

It would be great to have SDKMAN mentioned as an installation option on the Scala download page.


dhinojosa:
SDKMan is pretty well known and perhaps may not have made itself known to some in the Scala community. Though it is already well received, my bet, with the new Oracle Java Licensing model, SDKMan will be as well know as `public static void main`.

adriaanm:
Does SDKMan make it easier to install a JDK for users without JVM experience? I think that can be a valuable addition to our installation experience (say, for users coming from Python/Ruby). 

I do have one hesitation -- once we add SDKMan, how many more alternatives should we expect to receive PRs for? (Especially since there's a commercial offering.) I'm okay merging this, but I'll let @sjrd decide.

adriaanm:
One alternative could be to add this to a new section in https://github.com/scala/scala-lang/blob/master/download/install.md

sjrd:
> I do have one hesitation -- once we add SDKMan, how many more alternatives should we expect to receive PRs for?

I also share this concern. Is SDKMAN somehow *the* such manager?

Additionally, the proposed link is not really at the same "level of abstraction" as the others. All the other links point to a *Scala-specific* location, which directly explains how to get Scala running (or runs Scala in the browser right away). The link to SDKMAN is its homepage, which only mentions Scala in passing, among a long list of all-things-Java. If I go there, honestly I have no idea what the next steps are to get Scala. I had to explore the site to finally get to https://sdkman.io/sdks#scala, which does suggest that, if I manage to install SDKMAN, then after that I can do `sdk install scala`.

Is there a Scala-specific landing page that explains how to get from zero to installing Scala using SDKMAN, which we could more easily link to than the generic home page?

marc0der:
@adriaanm yes, sdkman makes it very easy to install both the JDK, Scala and even SBT. It would be something like:

    $ sdk install java
    $ sdk install scala
    $ sdk install sbt

It manages all the `_HOME` environment variables, as well as updates your `PATH` on-the-fly.

Lastly, most of the sdkman backend is written in Scala :-)

marc0der:
@sjrd That is a very good point, users could get lost if they simply get taken to our home page.

The Kotlin community have done the following on their site, giving a [short recipe](https://kotlinlang.org/docs/tutorials/command-line.html) of how to install with sdkman, brew and a few others. Alternatively, we do have this [section](https://sdkman.io/sdks#scala) on our website that gives a deep link to Scala and how to install.

sjrd:
Perhaps an alternative formulation for that line would be:

    <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install Scala</a> with <code>sdk install scala</code>.</li>

which makes it clearer that SDKMAN per se is not Scala-specific, gives a link the Scala-specific section of the website and provides the user who already knows SDKMAN with the actual one-liner install incantation.

marc0der:
@sjrd I made the change as suggested.

SethTisue:
Well, I would prefer that if we have this, we also had more such links to various package managers (e.g. Homebrew, which is where I currently get my JDK+Scala+sbt), rather than SDKMAN only.

But perhaps we should go ahead and merge this and invite the community to follow up with more PRs in this area. "how many more alternatives should we expect to receive PRs for?" asks Adriaan — if it eventually becomes unwieldy, we can deal with it then (perhaps with more compact formatting).

PR #958, opened 2018-10-03
License updated to Apache v2

adriaanm:
Also note that we include software with other licenses.

### DIFF
--- a/license.md
+++ b/license.md
@@ -18,3 +18,5 @@ limitations under the License.
 
 Scala includes software with other licenses. See the NOTICE file distributed 
 with this work for additional information regarding copyright ownership.
+
+NOTE: Versions of Scala distributed prior to October 2018 were licensed under the [BSD 3-Clause License](http://opensource.org/licenses/BSD-3-Clause). The license change was [announced in May 2018](https://www.scala-lang.org/news/license-change.html).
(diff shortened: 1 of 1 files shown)


SethTisue:
Since 2.11.x and 2.12.x (and to some extent 2.10.x) are still in wide use, I think we probably need to include some mention (the briefest possible would be fine, I think) that versions 2.12.x and earlier were under a different license.

adriaanm:
Added a blurb.

license.md
```diff
@@ -18,3 +18,5 @@ limitations under the License.
 
 Scala includes software with other licenses. See the NOTICE file distributed 
 with this work for additional information regarding copyright ownership.
+
+NOTE: Versions of Scala distributed prior to October 2018 were licensed under the [BSD 3-Clause License](http://opensource.org/licenses/BSD-3-Clause). The license change was [announced in May 2018](https://www.scala-lang.org/news/license-change.html).
```
dwijnand: Probably best use versions rather than dates, no?  If we release 2.12.8 this will be incorrect.

adriaanm: We should update 2.12.8 to Apache 2 as well, and hopefully not have to release it this month :-)

SethTisue:
LGTM, but we should wait to merge until at least one release is out with the new license.

dwijnand:
There's also #7407

SethTisue:
tiny fixup: 1c404469abfbb9cd84e35a7f688bc8bad7d8a195

PR #963, opened 2018-10-12
Remove console statements

Philippus:
They don't belong in production code.


sjrd:
They can probably all go away, including `console.error`s. It's not going to help anyone in production.

PR #967, opened 2018-10-21
Use absolute url's in opengraph tags, fixes #964

### LINKED ISSUE #964: Sharing blog posts on Twitter
The Twitter game is less good than I remember.

Posting links to the blog has a gray icon.

Twitter post 10:17 AM - 4 Oct 2018

May require more meta tags:

https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary.html

> `twitter:image`
> A URL to a unique image representing the content of the page. You should not use a generic image such as your website logo, author photo, or other image that spans multiple pages. Images for this Card support an aspect ratio of 1:1 with minimum dimensions of 144x144 or maximum of 4096x4096 pixels. Images must be less than 5MB in size. The image will be cropped to a square on all platforms. JPG, PNG, WEBP and GIF formats are supported. Only the first frame of an animated GIF will be used. SVG is not supported.

This image is specified on the Scala site by the Open Graph meta tag:

https://www.scala-lang.org/resources/img/scala-spiral-3d-2-toned-down.png

Running it through Twitter's validator,

```
INFO:  Page fetched successfully
INFO:  10 metatags were found
INFO:  twitter:card = summary tag found
ERROR: Possibly invalid value (if specified) (og:title)
ERROR: Possibly invalid value (if specified) (twitter:text:title)
ERROR: Possibly invalid value (if specified) (twitter:title)
```

Here's Twitter's card validator:

https://cards-dev.twitter.com/validator

Twitter is trying to tell us to fix something and punishing with a bad logo?   Maybe just fix the `og:title`, and the images might start working?

Philippus:


_includes/headertop.html
```diff
@@ -13,9 +15,9 @@
     <meta name="twitter:site" content="@scala_lang"/>
     <meta name="twitter:creator" content="@scala_lang"/>
     {% if page.url %}
-    <meta property="og:url" content="{{ site.baseurl }}{{ page.url }}"/>
+    <meta property="og:url" content="{{ site.url }}{{ page.url }}"/>
     {% endif %}    
-    <meta property="og:image" content="{{ site.baseurl }}/resources/img/scala-spiral-3d-2-toned-down.png"/>
+    <meta property="og:image" content="{{ site.url }}/resources/img/scala-spiral-3d-2-toned-down.png"/>
```
ashawley: Does this work?  I'm not seeing the full URL when I build the site locally.

It seems like the issue is that `baseurl` is empty in the `_config.yml`:

https://github.com/scala/scala-lang/blob/af99785/_config.yml#L5

Changing it to www.scala-lang.org has other issues I predict.

Philippus: Seems I have to set the `site.url` according to https://jekyllrb.com/docs/variables/.
In development it's set automatically: https://jekyllrb.com/news/#3-siteurl-is-set-by-the-development-server.

`baseurl` is something different, this page has a good explanation: https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/

ashawley: Cool, this looks promising.  It should  fix the bug on deployment, but still does the right thing for local development.

_includes/headertop.html
```diff
@@ -1,9 +1,11 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>{% if page.title %}{{ page.title }} | {% endif %}{{ site.title }}</title> 
+    <title>{% if page.title %}{{ page.title }} | {% endif %}{{ site.title }}</title>
     {% if page.title %}
     <meta property="og:title" content="{{ page.title }}"/>
+    {% else %}
+    <meta property="og:title" content="{{ site.title }}"/>
```
ashawley: Not sure if this is necessary (Why is `page.title` empty in the first place...?  ), but at least it's consistent with `<title>`.  I guess this is just a thing with Jekyll.   ¯\\\_(ツ)\_/¯

_includes/headertop.html
ashawley: Doesn't look like this change will disrupt link checking by htmlproofer, so there's validity in the Travis test passing.  Doesn't seem it's getting bypassed.

SethTisue:
shall we merge?

PR #972, opened 2018-10-29
Add Scala Days 2019

Philippus:


SethTisue:
very nice, thanks for improving the historical data

Philippus:
the very first Scala Days are still in the internet archive

PR #974, opened 2018-11-15
Provide a direct link to Java 8 download page

soronpo:
This is to prevent beginners from accidentally using Java 11, since it's not yet supported on mainstream Scala releases.
See also my comments on https://contributors.scala-lang.org/t/java-version-collision-for-beginners-weird-sbt-errors/2593

### DIFF
--- a/_layouts/downloadpage.html
+++ b/_layouts/downloadpage.html
@@ -25,7 +25,7 @@ layout: inner-page-parent
               <h3>First, make sure you have the Java 8 JDK installed.</h3>
               <p>To check, open the terminal and type:</p>
               <p><code>java -version</code><span>(Make sure you have version 1.8.)</span></p>
-              <p><i>(If you don't have it installed, <a href="http://www.oracle.com/technetwork/java/javase/downloads/index.html">download Java here</a>.)</i></p>
+              <p><i>(If you don't have it installed, <a href="https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html">download Java here</a>.)</i></p>
             </div>
           </div>
           <div class="step">


nafg:
IIUC it works fine, although there are some JDK 9+ features that aren't supported yet, for instance you can't do Jigsaw modules. Also you get a warning because SBT uses protobufs which does something deprecated. I've been using JDK 10 for a while.

There's a different issue which is that from JDK 11, you should use the OpenJDK distribution. Everything it was missing it now has, and so the Oracle distribution only exists to be a proprietary version, and they've changed the license accordingly.

SethTisue:
it's true that basic Scala + sbt works fine on Java 11, but there are libraries and tooling that haven't caught up yet, so I agree that it's best to steer beginners towards Java 8 for now

nafg:
@SethTisue in what sense? What doesn't work?

SethTisue:
> @SethTisue in what sense? What doesn't work?

I don't have a list handy and this PR isn't a good place to draft one, but it could be a good Discourse thread. (A possible outcome is that I'd be pleasantly surprised at how short a list we end up with.)

soronpo:
>I don't have a list handy and this PR isn't a good place to draft one, but it could be a good Discourse thread. (A possible outcome is that I'd be pleasantly surprised at how short a list we end up with.)

Is it possible to run the community build with JDK 11 and report which libraries fail?

soronpo:
> 
> 
> Sounds good.
> 
> Fix these two other pages that exist in the other repo as well?
> 
>     * [Getting Started with Scala in IntelliJ](https://docs.scala-lang.org/getting-started-intellij-track/getting-started-with-scala-in-intellij.html)
> 
>     * [Getting Started with Scala and sbt on the Command Line](https://docs.scala-lang.org/getting-started-sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.html)

Glad to do it, but I can't seem to find its source. Any clue what generates this?

Philippus:
they're in the `docs.scala-lang`-repo:
https://github.com/scala/docs.scala-lang/blob/master/getting-started-intellij-track/getting-started-with-scala-in-intellij.md and 
https://github.com/scala/docs.scala-lang/blob/master/getting-started-sbt-track/getting-started-with-scala-and-sbt-on-the-command-line.md

SethTisue:
someone want to follow up in the other repo?

soronpo:
Added PR to docs.scala-lang repo here https://github.com/scala/docs.scala-lang/pull/1269

PR #979, opened 2018-12-04
Update an example in Java

som-snytt:
Update the example to be more plausible yet no less pitiable.


SethTisue:
maybe include both the old and new styles? or old-style only but identify it as "Java (pre Java 8)" or something? a lot of Java programmers are still writing in the old style, and the same is true of other languages. showing how the code looks in imperative style is arguably more important than showing how it looks with an only rather superficially different Java API

SethTisue:
@som-snytt wdyt?

som-snytt:
Life was too short.

SethTisue:
more time to work on that JLine 3 PR 🤩

PR #981, opened 2018-12-04
Consider the height of the new-on-the-blog header for the EPFL marker

Philippus:
Currently, if the new-on-the-blog header is visible EPFL ends up in Lake Geneva, this PR adds 25 pixels on the y-axis in that case.


SethTisue:
EPFL saved from drowning!!

PR #983, opened 2018-12-08
add lampepfl org to CoC page

SethTisue:
at least, I assume the omission wasn't intentional?


SethTisue:
followup: scala/scala#7986

PR #992, opened 2018-12-26
Bad nescala link by adding "www"

ashawley:
Doing this worked for the other links.  Seems one snuck by me or got added before the fix was merged in #990.


Philippus:
thanks!

Issue #973, opened 2018-11-05
Upgrade jekyll to version 3.8.5


SethTisue:
would someone like to take a stab at this...? perhaps it will be easy or even trivial

RichardMarto:
The docker-compose.yml are just getting the latest version of jekyll/jekyll image. Maybe I can upgrade jekyll at jekyll/jekyll repository.

Philippus:
Here's an example of an earlier upgrade that was done: https://github.com/scala/scala-lang/pull/940

PR #940, opened 2018-07-25
Upgrade Jekyll to 3.8

jan0sch:
The jekyll installation is upgraded to 3.8.3. Because bundler refused to update farther than 3.6 it was necessary to remove to old `Gemfile.lock` and run `bundle install`.

So far it looks good.

This fixes #932

PR #995, opened 2019-01-08
Upgrade jekyll to version 3.8.5

RichardMarto:
This PR comes to solve #973.
I just followed the same path as #940 and I got similar results.


SethTisue:
looks like `.travis.yml` needs a change?

```
3.94s$ bundle install --jobs=3 --retry=3 --deployment --path=${BUNDLE_PATH:-vendor/bundle}
You must use Bundler 2 or greater with this lockfile.
```

SethTisue:
I'm a Ruby/rvm/bundler idiot, but maybe https://github.com/scala/scala/pull/7599 is relevant here as a model?

RichardMarto:
> Two things:
> 
> 1. Try removing the "BUNDLED WITH" section of the `Gemfile.lock` so that building the site doesn't depend on a particular version of bundler, but also fixes the Travis error Seth mentioned.
> 2. Change the value for `rvm` from "2.3.1" to be `2.4` in `.travis.yml`, since Travis enables Bundler 2.0 with 2.4 or later, since one wants to transition to Bundler 2.0.
> 
> I suppose we only need to do (1), but might as well do both.

It worked, I just did the first trick, thanks.

SethTisue:
@ashawley shall we merge?

ashawley:
This means the bundler required is version 1.x, still.  This may need to be documented in the README.   It's also unfortunate that running `bundle install` will modify `Gemfile.lock`, so people have to annoyingly ignore its change when committing to git.   This is where we're at with multiple major versions of bundler.

propensive:
@ashawley If I'm understanding your comment correctly, I think we can document our way past annoying log messages. If that's all we need, can I give this a gentle nudge towards being merged? My understanding of the domain is certainly inferior to @SethTisue's, so I think all we need is just a brief confirmation that it's safe to merge. :)

SethTisue:
thanks @RichardMarto and reviewers!

PR #1011, opened 2019-02-22
Fix #953: Make it clearer how to contact the moderation team.

sjrd:
Also /cc @darjutak 

@adriaanm @SethTisue @darjutak Tell me if it's fine for you to be listed there.

PR #1014, opened 2019-02-28
strengthen wording about in-channel discussion of moderation

SethTisue:
```diff
-If you think a ban is unjustified
+If you think a moderator action is unjustified
-Complaints about bans in-channel are not allowed
+Complaints about moderation in-channel are not allowed
```

this is pretty minor change, but it is a change. we don't really have a process for changing the CoC, but I guess I'll ask for review for starters from: @adriaanm @sjrd 

I think the new wording better reflects the intent and goals (my understanding of them, anyway, as a co-drafter of the CoC, and a longtime community moderator) of this part of the code. The goal is to keep the channel on-topic and not allow meta-discussion to distract from the purpose of the channel. Discussion about a warning is equally off-topic as discussion of a ban.

This dovetails with the (soon-to-be-merged, I believe) #1011, which provides better guidance about where to take complaints about moderation instead.

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -48,7 +48,7 @@ the [Scala Center](https://scala.epfl.ch/).
 - If the warning is unheeded, the user will be “kicked,” i.e., kicked out of the communication channel to cool off.
 - If the user comes back and continues to make trouble, they will be banned, i.e., indefinitely excluded.
 - Moderators may choose at their discretion to un-ban the user if it was a first offense and they if they make suitable amends with the offended party.
-- If you think a ban is unjustified, please take it up with that moderator, or with a different moderator, in private. Complaints about bans in-channel are not allowed.
+- If you think a moderator action is unjustified, please take it up with that moderator, or with a different moderator, in private. Complaints about moderation in-channel are not allowed.
 - Moderators are held to a higher standard than other community members. If a moderator acts inappropriately, they should expect less leeway than others.
 
 In the Scala community we strive to go the extra step to look out for each


Ichoran:
Did we ever answer the question about whether there was a formal venue to discuss moderation, or if it is always off topic in all channels?  On the one hand, moderation _is_ a potentially important issue relevant to the community, and everything _else_ with such status has a venue.  On the other hand, it's very difficult to keep it from degenerating (one would think so theoretically, and we have empirical evidence that this has been what happens).  On the first hand again, having an attempt at something constructive is possibly better than twitter flamewars; it's not like people will shut up everywhere just because the formal venues do not allow it.  Second hand again--maybe better to have brighter flames farther away than get scorched up close.

The reason I bring this up as relevant to this PR is that the wider you cast the net of "don't talk about this", the more pressure there is for some kind of alternative.

SethTisue:
> Did we ever answer the question about whether there was a formal venue to discuss moderation, or if it is always off topic in all channels

That's an excellent question. I'm wondering that myself.

Informally, I think the status quo is that https://contributors.scala-lang.org is the place. People are free, within reason, to start additional threads such as https://contributors.scala-lang.org/t/looking-for-a-moderation-process-to-enforce-the-coc/2330/76

Personally I think it's better if these moderation discussion are disallowed in all of our chat rooms. When chat starts flying back and forth, people say things they'll regret later and make poor choices in the moment. (I've certainly done it.) Discourse threads can go south too — the one I just linked to certainly did — but I still think it's better to have these meta-discussions in a medium that encourages a slower pace of discussion and more reflection. Also, I think it's absolutely crucial that Discourse lets people mute individual threads and entire topics they're not interested in, whereas everything in a chat room is all mixed together and there's no way to mute anything.

So, I feel pretty strongly that we ought to merge this PR, and #1011, regardless. But I'd very much welcome an additional PR addressing the question you've raised.

Issue #1006, opened 2019-02-08
blog post categories are confusing and inconsistent


SethTisue:
I've never understood what the difference between "Blog", "News", and "Announcement" is supposed to be at https://www.scala-lang.org/blog/

and worse, we have published some posts that aren't in any of the three categories — in the root `_posts` directory, so that's a fourth non-category category

SethTisue:
overlaps with #849 (that ticket was opened specifically about our RSS feed, but the general issue is broader than that)

SethTisue:
@adpi2 writes at #744:

> I merged Blog and News and added the Highlights section

(Highlights is specifically, and only, for the new Scala Highlights newsletter)

adpi2:
Also we discussed this further with @darjutak and we think we should rename `Announcement` to `Releases` and move all release notes and announcements to it.

We will have:
- `Blog`: Everything else, which is not `Highlights` nor `Releases`
- `Highlights`: only for the Scala Highlights newsletter
- `Releases`: all release notes, and release announcements

SethTisue:
does "blog" need to exist as a distinct category, or can everything else besides release and highlights just be no category at all?

SethTisue:
https://github.com/scala/scala-lang/pull/1769 tries to improve the situation by standardizing on `category: blog`, `category: release`, and `category: highlights` as the only possibilities

`category: news` is deleted and the old `category: announcements` was renamed to `category: release`

SethTisue:
I think this is now complete, actually.

Old blog posts have a bewildering variety of permalinks, but oh well. At least we can be consistent going forward.

We will likely still have some trouble with blog post authors cargo-culting incorrect boilerplate. If it occurs, I'll try and notice during PR review, and/or fix it later. The problem should lessen over time.

SethTisue:
OMG I CAN'T BELIEVE WE FINALLY FIXED THIS

Issue #849, opened 2018-02-27
Stale XML feed


ashawley:
I've been meaning to report that the main feed hasn't been updating since October, but suddenly it updated when the MOOC edx announcement was made in #847 .

http://scala-lang.org/feed/index.xml

The blog feed is fine and is still updating:

http://scala-lang.org/feed/blog.xml

I predict the top-level `_posts` directory needs to be merged in to the `news/_posts` (and maybe `blog/_posts`) directories, but I'm not sure.

SethTisue:
at https://users.scala-lang.org/t/any-way-to-subscribe-to-the-official-scala-blog-via-email/3955/7?u=sethtisue, crater2150 notes:

> there are only feeds for two of the sections of the blog, “News” and “Blog”, while “Announcement” and “All” are missing (also, “All” contains not only the posts of all three other, but also some which are in neither of the subsections, e.g. the currently newest post about the community build).

SethTisue:
I've opened a separate ticket about the categories at #1006. I hadn't understood before that the mess-up wasn't specific to the RSS feeds, the confusion is also reflected on the HTML site

for RSS purposes, I think the simplest thing would be a single feed that has everything

SethTisue:
can someone review https://github.com/scala/scala-lang/pull/1750 ?

PR #1017, opened 2019-03-01
Added a news item announcing the Northeast Scala Symposium's CFP.

bmc:


bmc:
I added this as a news item. If you'd prefer that it be a blog post, instead, just reject the PR and let me know, and I'll resubmit.

SethTisue:
I have misgivings about opening the scala-lang blog up to CFPs.

https://www.scala-lang.org/events lists about 25 events a year. If we publish a CFP blog post for each one, that's a ton of blog posts — it would be half the content on the blog.

I anticipate: "But it's a 'news item' not a 'blog post'" — but I'm not convinced the distinction between "blog post" and "news item" helps much with this, given #1006 and #849 

I do want the Scala org to help conference organizers get the word out about their conferences. Currently we're doing that by maintaining https://www.scala-lang.org/events, tweeting it monthly from https://twitter.com/scala_lang, and keeping those tweets pinned

Is there something more/better we could do to help get the word out? Probably, but I have doubts this is the way to do it.

at present, I'd suggest a post on https://users.scala-lang.org instead. posts like https://users.scala-lang.org/t/call-for-papers-scala-symposium-2019/4137 are extremely welcome there

Any other opinions on this...?

bmc:
Sounds good. Thanks, Seth.

PR #1018, opened 2019-03-04
add Sphere.it 2019

SethTisue:


SethTisue:
formerly ScalaSphere. there is still a dedicated Scala track

PR #1030, opened 2019-03-29
events - add fpure conference

dos65:


SethTisue:
how Scala-heavy is it? I can't tell from the website.

we only list general FP conferences if there is a clear, strong Scala-specific presence, not just a few talks here and there

dos65:
It is indeed a Scala-specific FP conference similar to Fby and LambdaConf.
Scala community is probably onе of the largest among other FP communities in Russia. We expect lots of Scala users to attend and as a consequence, Scala talks would take big enough part of the entire conference programme.

SethTisue:
cool, thanks for confirming

PR #1062, opened 2019-08-01
remove 404 links

SethTisue:


som-snytt:
I guess pride goes before the fall.

PR #1072, opened 2019-09-09
SBTB is in Oakland this year, and the dates changed slightly

SethTisue:
as pointed out by @longcao

### DIFF
--- a/_events/2019-11-14-scale-bythebay.md
+++ b/_events/2019-11-14-scale-bythebay.md
@@ -4,7 +4,7 @@ title: Scale by the Bay
 logo: /resources/img/scalebythebay.png
 location: Oakland, CA, USA
 description: ""
-start: 14 November 2019
-end: 16 November 2019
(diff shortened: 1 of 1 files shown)


_events/2019-11-14-scale-bythebay.md
```diff
@@ -2,7 +2,7 @@
 category: event
 title: Scale by the Bay
 logo: /resources/img/scalebythebay.png
-location: San Francisco, CA, USA
+location: Oakland, CA, USA
 description: ""
 start: 14 November 2019
 end: 16 November 2019
```
Philippus: Looks like the dates should be changed to 13-15 November as well

SethTisue: good catch! I pushed a second commit with a fix

SethTisue:
merging on the fast track since scala_lang just did the monthly events tweet

PR #1079, opened 2019-09-20
Small updates to Code of Conduct language

TimMoore:
These changes were made in the [Lightbend Community Code of Conduct](https://www.lightbend.com/conduct) and are being proposed to the upstream Scala Code of Conduct as minor improvements consistent with the spirit of the existing code.

- Update introductory statement with characteristics from the [Contributor Covenant v1.4.1](https://www.contributor-covenant.org/version/1/4/code-of-conduct)
- Inline examples of unacceptable behavior from Citizen Code of Conduct
- Consistently suggest to "contact the moderation team" rather than "contact one of the moderators"
- Clarify moderation of cursing in "a hateful or aggressive" manner
- Include Citizen Code of Conduct in the Credits section

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -47,7 +47,7 @@ As a member of the community, you agree to the following:
   - Sustained disruption of community events, including talks and presentations.
   - In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
 - Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact the moderation team](#contact) immediately.
-- Likewise any spamming, trolling, flaming, baiting or other attention-stealing behaviour is not welcome.
+- Likewise any spamming, trolling, flaming, baiting or other attention-stealing behavior is not welcome.
 
 ### Moderation
 
--- a/conduct.md
+++ b/conduct.md
@@ -33,7 +33,7 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
-- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/):
+- We will exclude you from interaction if you insult, demean or harass anyone. Behavior that will lead to exclusion includes the following:
   - Violence, threats of violence or violent language directed against another person.
   - Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
   - Posting or displaying sexually explicit or violent material.
--- a/conduct.md
+++ b/conduct.md
@@ -11,9 +11,10 @@ redirect_from:
 ## Scala Code of Conduct
 
 We are committed to providing a friendly, safe and welcoming environment for
-all, regardless of level of experience, gender, gender identity and expression,
-sexual orientation, disability, personal appearance, body size, race, ethnicity,
-age, religion, nationality, or other such characteristics.
+all, regardless of age, body size, disability, ethnicity, sex characteristics,
+gender identity and expression, level of experience, education, socio-economic
+status, nationality, personal appearance, race, religion, or sexual identity
+and orientation.
 
 ### Our Standards
 
@@ -32,8 +33,20 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
-- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/). **In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.**
-- Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact one of the moderators](#contact) immediately.
+- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/):
+  - Violence, threats of violence or violent language directed against another person.
+  - Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
+  - Posting or displaying sexually explicit or violent material.
+  - Posting or threatening to post other people’s personally identifying information ("doxing").
+  - Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
+  - Inappropriate photography or recording.
+  - Inappropriate physical contact. You should have someone’s consent before touching them.
+  - Unwelcome sexual attention. This includes, sexualized comments or jokes; inappropriate touching, groping, and unwelcomed sexual advances.
+  - Deliberate intimidation, stalking or following (online or in person).
+  - Advocating for, or encouraging, any of the above behavior.
+  - Sustained disruption of community events, including talks and presentations.
+  - In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
+- Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact the moderation team](#contact) immediately.
 - Likewise any spamming, trolling, flaming, baiting or other attention-stealing behaviour is not welcome.
 
 ### Moderation
@@ -42,7 +55,7 @@ These are the policies for upholding our community’s standards of conduct. If
 you feel that a thread needs moderation, please
 [contact the moderation team](#contact).
 
-- Remarks that violate the above code of conduct, including hateful, hurtful, oppressive, or exclusionary remarks, are not allowed. (Cursing is allowed, but never targeting another user, and never in a hateful manner.)
+- Remarks that violate the above code of conduct, including hateful, hurtful, oppressive, or exclusionary remarks, are not allowed. (Cursing is allowed, but never targeting another user, and never in a hateful or aggressive manner.)
 - Moderators will warn users who make remarks inconsistent with the above code of conduct.
 - If the warning is unheeded, the user will be “kicked,” i.e., kicked out of the communication channel to cool off.
 - If the user comes back and continues to make trouble, they will be banned, i.e., indefinitely excluded.
@@ -102,3 +115,4 @@ Adapted from and/or inspired by multiple successful Codes of Conduct, including:
 * [The Contributor Covenant v1.4.0](http://contributor-covenant.org/version/1/4/)
 * [The Recurse Center's User Manual](https://www.recurse.com/manual#sec-environment)
 * [The 18F Code of Conduct](https://18f.gsa.gov/code-of-conduct/)
+* [Citizen Code of Conduct](http://citizencodeofconduct.org/)
--- a/conduct.md
+++ b/conduct.md
@@ -13,8 +13,8 @@ redirect_from:
 We are committed to providing a friendly, safe and welcoming environment for
 all, regardless of age, body size, disability, ethnicity, sex characteristics,
 gender identity and expression, level of experience, education, socio-economic
-status, nationality, personal appearance, race, religion, or sexual identity
-and orientation.
+status, nationality, personal appearance, race, religion, sexual identity
+and orientation, or other such characteristics.
 
 ### Our Standards
 


tpolecat:
It's awkward for the language's maintainer to not adopt the language's own CoC, so to the extent that this is necessary for Lightbend (due to many of their community members not being Scala programmers IIUC) I hope we can converge on common language quickly. These changes seem fine to me.

conduct.md
```diff
@@ -11,9 +11,10 @@ redirect_from:
 ## Scala Code of Conduct
 
 We are committed to providing a friendly, safe and welcoming environment for
-all, regardless of level of experience, gender, gender identity and expression,
-sexual orientation, disability, personal appearance, body size, race, ethnicity,
-age, religion, nationality, or other such characteristics.
+all, regardless of age, body size, disability, ethnicity, sex characteristics,
+gender identity and expression, level of experience, education, socio-economic
+status, nationality, personal appearance, race, religion, or sexual identity
+and orientation.
```
SethTisue: does this new wording come from somewhere specific, or is it original?

I'm not sure I understand the motivation for the changes, what problem is being solved. (It looks like somebody almost alphabetized the list, but not quite?)

"sex characteristics" seems like odd wording to me, I don't really know what it means. what does it mean that isn't already covered by "gender identity and expression, [...], sexual identity and orientation"?

conduct.md
```diff
@@ -32,8 +33,20 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
-- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/). **In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.**
```
SethTisue: I agree that it's better to inline the relevant part of the Citizen Code of Conduct, better than requiring somebody follow a link. :+1:

conduct.md
```diff
@@ -42,7 +55,7 @@ These are the policies for upholding our community’s standards of conduct. If
 you feel that a thread needs moderation, please
 [contact the moderation team](#contact).
 
-- Remarks that violate the above code of conduct, including hateful, hurtful, oppressive, or exclusionary remarks, are not allowed. (Cursing is allowed, but never targeting another user, and never in a hateful manner.)
+- Remarks that violate the above code of conduct, including hateful, hurtful, oppressive, or exclusionary remarks, are not allowed. (Cursing is allowed, but never targeting another user, and never in a hateful or aggressive manner.)
```
SethTisue: This seems like an improvement to me :+1:

SethTisue:
attn @sjrd @darjutak

conduct.md
SethTisue: :+1: to aligning with Contributor Covenant here

SethTisue:
> Consistently suggest to "contact the moderation team" rather than "contact one of the moderators"

I don't see any changes in this area in the diffs?

TimMoore:
> > Consistently suggest to "contact the moderation team" rather than "contact one of the moderators"
>
> I don't see any changes in this area in the diffs?

It's hard to spot because it got lumped in with the inlining of the Citizen Code of Conduct:

https://github.com/scala/scala-lang/pull/1079/files#diff-102cddd805ad66d3e769233af629d8b3R49

```diff
- - Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact one of the moderators](#contact) immediately.
+ - Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact the moderation team](#contact) immediately.
```

TimMoore:
To @tpolecat's point, if for some reason these changes aren't accepted, we can change the version at https://www.lightbend.com/conduct to align. I agree that keeping them as close as possible is a good thing, and more important than making minor clarifications that cause them to diverge. However these changes were all suggested by various Lightbend employees as improvements to the original wording or formatting, and I agree that they are improvements, so they seemed worth including with the intent of proposing them back here.

In the future, I think it will be better to suggest the improvements upstream here first, and then align https://www.lightbend.com/conduct with any changes that are merged into this version.

conduct.md
SethTisue: (note that the original wording was copied from the Rust CoC)

conduct.md
```diff
@@ -32,8 +33,20 @@ As a member of the community, you agree to the following:
...
-- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/). **In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.**
-- Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact one of the moderators](#contact) immediately.
+- We will exclude you from interaction if you insult, demean or harass anyone. The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/):
```
martijnhoekstra: Maybe completely drop what is or isn't included in the term Harassment. It's irrelevant and may invite rule-lawyering.
```suggestion
- We will exclude you from interaction if you insult, demean or harass anyone. Behaviour that will lead to exclusion includes the following:
```

SethTisue: this seems like a sensible adjustment, clarifying what I understand to be the intended meaning anyway. and it's within the scope of this PR. @TimMoore wdyt?

conduct.md
soronpo: What about aligning it with https://www.fossi-foundation.org/code-of-conduct ?
all, regardless of age, culture, ethnicity, genotype, gender identity or expression, language, national origin, neurotype, phenotype, political beliefs, profession, race, religion, sexual orientation, socioeconomic status, subculture and technical ability.

TimMoore: @soronpo can you explain a bit more why you think that would be better?

TimMoore: To add a bit of background, the currently published text in this section appears to be derived from [Contributor Covenant v1.3](https://www.contributor-covenant.org/version/1/3/0/code-of-conduct):

> We are committed to making participation in this project a harassment-free experience for everyone, regardless of level of experience, gender, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, or nationality.

So, my aim was to align with the text in the latest version of the same document.

conduct.md
TimMoore: Yes, I agree. I'll allow another couple of days for further comments from the community before committing this suggestion.

Ichoran: You get rules-lawyering whether or not you specify the rules.  If adding rules adds clarity, go for it.  If it is misleading, drop it.  If you're worried about rules-lawyering the only solution is to not have rules.

conduct.md
Ichoran: I'm not terribly thrilled about the upstream document.  If it's alphabetized, it's full of mistakes.  If it's in order of importance or how often it comes up, it seems really bizarre.  Alignment is good, I suppose, but this looks like we don't care.  The original list seems better ordered, and I know what all the terms mean.  It looks like they replaced "gender" with "sex characteristics", I guess because technically the term is "sex" but that's ambiguous between the morphology (which is what is meant), and the act?  Also, there are no longer "other such characteristics", so for example we're absolutely able to be unfriendly, dangerous, and unwelcoming to people on the basis of whether or not they like cats?

Anyway, I don't imagine the change will cause problems, but I think it's strictly worse.

SethTisue: @soronpo on the one hand, I'm a bit reluctant to see the Scala CoC become _even more_ of a Frankenstein, with bits and pieces assembled from _even more_ other codes. on the other hand, the FOSSi wording does seem higher quality (and Rex has done a good job of unpacking some of the reasons)

I'm fine with any of these versions (including just sticking with what we originally had)

soronpo: It seems both lists align in most ways. The notable exception is `political beliefs` that has no equivalent in the Contributor Covenant. I believe that is on purpose, and not in a good way.

SethTisue: ah, I see. in that case I think using the FOSSi text would be a more substantial change that would need a separate PR and discussion. I suggest we keep this PR scoped more modestly.

Ichoran: Both the Fossi code and the Contributor Covenant explicitly welcome _everyone_, with no exceptions.  The lists, in both cases, are just to warn people that one recognizes these things come up and that, as with every other reason, these things will not be stood for.  So I don't have much of a preference for content, as long as it covers the basics.  If we're arguing about whether a particular unpleasantness was or wasn't called out by name, things have already gone awry.

If the "for all" bit is rhetorical flourish, not intended to be taken seriously, while the list is intended to be taken seriously, then I wouldn't think "political beliefs" is the biggest difference.  People don't talk about politics much on any of these venues; it's off-topic.

"language" seems more likely to come up (e.g. is a venue English-only, and if so, what is your attitude towards people who don't respect that...or people who are upset when non-English is used extensively in a venue without a specified language).  "neurotype" is also potentially significant (again, both ways).

Again, I don't think any of these things ought to matter much if we forthrightly mean "all", "everyone", and the like.  If we mean "we feel free to make things unpleasant for Russian-speaking autistic Libertarians", we probably ought to change the "all" stuff before worrying about Fossi vs. Contributor Covenant.

conduct.md
```diff
@@ -32,8 +33,20 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
```
Ichoran: This wasn't changed, but by using "That' s so easy my grandmother could do it" as an example of micro-aggressive behavior against the person you're talking to, aren't we demonstrating age discrimination?  Maybe we want to eat our own dog food here and not insult grandmothers?

SethTisue: I thought the point was that it insults both the listener _and_ grandmothers. But if you didn't read it that way, perhaps there's a clarity problem that should be addressed. (But not in this PR, let's stay focused on the changes Tim is proposing.)

conduct.md
TimMoore: @Ichoran I agree, "for all" is the most important part of the statement.

I can see now that "other such characteristics" wasn't in the original Contributor Covenant document, and seems to be adapted from the slightly different version in the [Rust Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct):

> We are committed to providing a friendly, safe and welcoming environment for all, regardless of level of experience, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, nationality, or other similar characteristic.

Removing "or other such characteristics" was an unintended oversight on my part. I'll put it back.

As for the rest of it, I don't have a strong preference for the new vs. old wording, but this did come up from internal feedback when drafting the Lightbend variant at https://www.lightbend.com/conduct. Some felt that the list was missing characteristics that we should explicitly include. A list like this can never be fully comprehensive, but I am very reluctant to remove characteristics from the list that has already been published there, as it might be incorrectly interpreted to mean we no longer consider those characteristics to be protected.

So if the consensus here is not to change this section, I'll remove that from this pull request, but leave it as a difference between the Scala and Lightbend versions of the document.

On the other hand, if the decision is to reorder or add additional characteristics, I'd want to align the Lightbend copy to it.

conduct.md
TimMoore: @Ichoran just to make sure I understand: are you objecting to @martijnhoekstra's suggested change?

conduct.md
TimMoore: Thanks. I agree that further changes should be suggested in new PRs.

conduct.md
Ichoran: I haven't formed an opinion on the change itself.  I am objecting to the stated reasoning.

In one place, we list age and gender and so on without fear that it will cause "rules-lawyering" or that it is irrelevant given that we welcome "all".

Here, the proposal is to drop even a link to a list to avoid "rules-lawyering".

This seems inconsistent.  So if there is a reason to remove lists, the stated reason doesn't quite seem to be it.

SethTisue: I agree that "invites rule-lawyering" isn't the best defense of Martijn's tweak. I simply think that Martijn's tweak makes the wording clearer.

As for the main change here, Rex I think you may have misunderstood the nature of the change, probably because you are looking at just this one point in the diffs. If you look at the whole diff, you'll see that "the proposal is to drop even a link to a list" simply isn't what's happening here. Rather, the link is being replaced by an inlined list.

conduct.md
SethTisue: > So if the consensus here is not to change this section, I'll remove that from this pull request, but leave it as a difference between the Scala and Lightbend versions of the document.

Okay. That seems... inelegant, but not terrible. I guess if anyone asks, we can point them at this discussion and shrug.

SethTisue: > we can point them at this discussion

and/or at https://www.contributor-covenant.org/faq ("The Contributor Covenant explicitly lists a set of protected classes; does this make it acceptable to discriminate or make others feel unwelcome based on other factors?" ...)

conduct.md
Ichoran: Ah, okay.  'You're right; I'd thought the diff would show the following stuff.  Sounds reasonable to inline the list, and being clearer is enough!

However, are we giving appropriate credit if we remove the link?

SethTisue: > are we giving appropriate credit if we remove the link?

see credits at the bottom. our code already incorporates big chunks of text from the other codes, and Citizen CoC is already listed.

conduct.md
NthPortal: > It looks like they replaced "gender" with "sex characteristics", I guess because technically the term is "sex" but that's ambiguous between the morphology

hi, your local trans girl here 👋

gender (identity) and sex are not the same. the former is very important to a lot of people (myself included), and in very simplified terms is "are you nonbinary, a woman, genderqueer, a man, genderfluid, bigender, etc.?". the latter is... complicated for a variety of reasons. I don't want to discount the identity or experience of intersex folks, but for the most part, sex (in western culture) is garbage and an attempt to classify everyone as "male" or "female", regardless of the fact that those categories don't fit well and aren't useful for a lot of people.

tl;dr: "gender" and "sex" are not the same. IMO, gender is the more important one (please please *please* do not change the wording to "sex"), but "sex characteristics" may also be a useful category to explicitly specify (I'm not intersex, so I can't really speak for those who are).

TimMoore: Thanks for adding your perspective, @NthPortal!

I'd also like to point out that the new list includes: "gender identity and expression". So, I don't believe "gender" was simply replaced by "sex characteristics", but rather the new text acknowledges inclusivity across both of the distinct categories of gender and sex.

TimMoore: Here are the origins of the upstream changes. The addition of "sex characteristics" and the removal of "gender" were made in separate changes:

- Collapsing "gender, gender expression and identity" to only "gender expression and identity" was done in the initial draft of the 1.4 version: https://github.com/ContributorCovenant/contributor_covenant/pull/226. There's no explicit explanation of the change.
- Adding "sex characteristics" came in a later update with this explanation: "Adding 'sex characteristics' to the examples, as some harassment and discrimination is based on sex characteristics, including but not limited to references to menstruation, fertility, and genitalia." https://github.com/ContributorCovenant/contributor_covenant/pull/550

conduct.md
TimMoore: Thanks for clarifying this, @SethTisue. I agree that with the list inlined, the link is no longer so important, and the suggested change adds clarity. With Citizen Code of Conduct explicitly acknowledged in the credits, I believe we've done our duty of attribution.

@Ichoran, please let me know if you're satisfied with this explanation, and if so I'll accept @martijnhoekstra's change to the lead sentence, assuming there are no other objections.

As a separate issue, I noticed that the Citizen Code of Conduct is licensed under the [CC Attribution-ShareAlike license](https://creativecommons.org/licenses/by-sa/3.0/). This is a GPL-like license that requires that derived works be distributed under the same license. In my interpretation, this should have been true all along, since the Scala CoC built upon the Citizen CoC in a way, even before inlining the content. Currently, there's no explicit license documented, but the Domain section implies permission to reuse the Scala CoC, and of course in practice other projects have adopted it. So, I think the spirit of that license is embraced, but abiding strictly by its terms would require explicitly documenting the use of that license, and also requiring other users of the Scala CoC to use the same license.

TimMoore: I'll raise another pull request to add that license, and we can discuss that specific issue there.

TimMoore: https://github.com/scala/scala-lang/pull/1082 I think that will need to be resolved before this one can be merged. Should I also start a thread on Discourse?

SethTisue: I don't think a Discourse thread is necessary, I don't think there is broad interest in what the license on the Scala CoC itself is or might be

conduct.md
NthPortal: @TimMoore sorry, I wasn't objecting to the proposed wording; I was replying to Rex's comment.

I'm personally happy with just straight aligning with the Contributor Covenant here. They have a good list, and it makes maintenance more straightforward.

Ichoran: I think the changes are fine!

Also, if I object to a certain line of reasoning, it doesn't follow that I object to the changes.

As a biologist (hey, I'm your local biologist here! :wave:) who understands the regulation of sex-specific expression in mammals (and the rates of morphology that isn't clearly dimorphic in one direction or the other), I'm rather baffled by the apparent claim that classifying _almost_ everyone (~99.8%) as "male" or "female" morphologically (sans hormone therapy) isn't the only rational thing to do.  I'm also baffled by the apparent culture-bashing.  Telling people that their culture has aspects which are "garbage" isn't a very good way to be inclusive.  When something has problems, pointing out the flaws politely and constructively is generally the way to go?

But none of that matters to the CoC.  The point of the CoC is to call us to be kind and welcoming, and to specifically reject that particular kind of nastiness that often manifests when someone seems to be "different" in some substantive way.  Thanks, Tim, for digging out the justification!  I had assumed they actually just meant "sex" since it seems weird to me to not mention sex but mention sex characteristics, but sure! (Background: with the fuss over bathrooms and other semi-private places, the objection is generally voiced as something like _X is a man and I don't care how they self-identify._ It's specifically sex, not gender, not sex characteristics (isasmuch as those can be divorced from sex).)

Anyway--I'm also fine with just aligning with CoC!

SethTisue: I feel like this has become contentious in a way that's outside the boundaries of this PR, so I'd like to ask that any further background discussion about the deeper underlying issues here go elsewhere.

It sounds like everyone is at least reasonably comfortable with adopting the Contributor Covenant language, or at least doesn't consider it notably worse than what we have already, so that seems like consensus to me — within the boundaries of this PR anyway, which is where I ask that we stay.

conduct.md
TimMoore: I'll go ahead and commit the suggestion from @martijnhoekstra now.

SethTisue: I merged the PR with the new license.

SethTisue:
thanks, all, for the help getting this into final form

PR #1087, opened 2019-10-24
Move the examples of harassment to separate section in code of conduct

patriknw:
This will make the Encouraged and Discouraged sections more
balanced.

The examples are from the Citizen Code of Conduct and would be
good to clarify that to avoid the misunderstanding that these were
based on incidents in our community.

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -33,8 +33,7 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
-- We will exclude you from interaction if you insult, demean or harass anyone. See [examples of unacceptable behavior](#examples-of-unacceptable-behavior) below.
-- We don’t tolerate behavior that excludes people based on their association with social, ethnic, religious, political, or similar groups.
+- We will exclude you from interaction if you insult, demean or harass anyone. See [examples of unacceptable behavior](#examples-of-unacceptable-behavior) below. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
 - Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact the moderation team](#contact) immediately.
 - Likewise any spamming, trolling, flaming, baiting or other attention-stealing behavior is not welcome.
 
--- a/conduct.md
+++ b/conduct.md
@@ -33,19 +33,7 @@ As a member of the community, you agree to the following:
 
 - Keep unstructured critique to a minimum. We encourage sharing ideas and perspectives, so please ensure that your feedback is constructive and relevant. If you have solid ideas you want to experiment with, make a fork and see how it works.
 - Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
-- We will exclude you from interaction if you insult, demean or harass anyone. Behavior that will lead to exclusion includes the following:
-  - Violence, threats of violence or violent language directed against another person.
-  - Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
-  - Posting or displaying sexually explicit or violent material.
-  - Posting or threatening to post other people’s personally identifying information ("doxing").
-  - Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
-  - Inappropriate photography or recording.
-  - Inappropriate physical contact. You should have someone’s consent before touching them.
-  - Unwelcome sexual attention. This includes, sexualized comments or jokes; inappropriate touching, groping, and unwelcomed sexual advances.
-  - Deliberate intimidation, stalking or following (online or in person).
-  - Advocating for, or encouraging, any of the above behavior.
-  - Sustained disruption of community events, including talks and presentations.
-  - In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
+- We will exclude you from interaction if you insult, demean or harass anyone. See [examples of harassment](#examples-of-harassment) below. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
 - Private harassment is also unacceptable. No matter who you are, if you feel you have been or are being harassed or made uncomfortable by a community member's behavior, please [contact the moderation team](#contact) immediately.
 - Likewise any spamming, trolling, flaming, baiting or other attention-stealing behavior is not welcome.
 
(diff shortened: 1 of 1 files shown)


conduct.md
```diff
@@ -106,6 +94,23 @@ listed above,
...
+- Posting or displaying sexually explicit or violent material.
+- Posting or threatening to post other people’s personally identifying information ("doxing").
+- Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
+- Inappropriate photography or recording.
+- Inappropriate physical contact. You should have someone’s consent before touching them.
+- Unwelcome sexual attention. This includes, sexualized comments or jokes; inappropriate touching, groping, and unwelcomed sexual advances.
+- Deliberate intimidation, stalking or following (online or in person).
+- Advocating for, or encouraging, any of the above behavior.
+- Sustained disruption of community events, including talks and presentations.
+
```
patriknw: These were added recently in https://github.com/scala/scala-lang/commit/e45719817c0ef761ac4e75a68757e66807e6a0cb#diff-102cddd805ad66d3e769233af629d8b3L35-R49 to make it independent of external resource "Citizen Code of Conduct", as I have understood it.

conduct.md
```diff
@@ -106,6 +94,23 @@ listed above,
   * Sébastien Doeraene, Scala Center ([sebastien.doeraene@epfl.ch](mailto:sebastien.doeraene@epfl.ch))
   * Kathleen Hayes, Lightbend ([kathleen.hayes@lightbend.com](mailto:kathleen.hayes@lightbend.com))
 
+### Examples of harassment
+
+The term “Harassment” includes “Unacceptable Behavior” described in the [Citizen Code of Conduct](http://citizencodeofconduct.org/).
```
sjrd: What's the reason for reverting the changes in https://github.com/scala/scala-lang/commit/2f76f5246deff5c7423b8cf16c15a9757c420c6b? The purpose was to simplify the main text of this CoC not to depend on another document.

conduct.md
```diff
@@ -106,6 +94,23 @@ listed above,
...
+Examples from that resource, but not based on events in our community:
```
sjrd: This reads as if it wants to say "no, believe me, nothing that bad happened here". I do not think this is appropriate. I preferred the previous text which was completely neutral wrt. whether or not things did happen before or not:

> Behavior that will lead to exclusion includes the following:

conduct.md
```diff
@@ -106,6 +94,23 @@ listed above,
...
+- Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
+- Posting or displaying sexually explicit or violent material.
+- Posting or threatening to post other people’s personally identifying information ("doxing").
+- Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
+- Inappropriate photography or recording.
+- Inappropriate physical contact. You should have someone’s consent before touching them.
+- Unwelcome sexual attention. This includes, sexualized comments or jokes; inappropriate touching, groping, and unwelcomed sexual advances.
+- Deliberate intimidation, stalking or following (online or in person).
+- Advocating for, or encouraging, any of the above behavior.
+- Sustained disruption of community events, including talks and presentations.
```
sjrd: Use `*` for top-level enumerations. `-` is used for second-level enumerations.

conduct.md
adriaanm: I think not clarifying this is more likely to lead to conspiracy theories. Explicit good. Implicits bad :-)

conduct.md
patriknw: The reason is to avoid the possible interpretation that these examples were taken from historical events in our community. Since these examples were taken (verbatim) from the Citizen Code of Conduct it would be good to keep the reference and clarify that they are from that resource.

conduct.md
patriknw: I'm open to alternative formulation.

conduct.md
patriknw: fixend, thanks

conduct.md
SethTisue: I lean towards Seb's position here. I think referencing the Citizen Code of Conduct is a sufficient disclaimer.

Also, we shouldn't be too confident these things aren't happening in our community. I'm not aware of any doxing, violence, or sustained disruption of talks/events, true; but the rest of the list definitely happens in any sufficiently large tech community, and that includes ours.

conduct.md
SethTisue: Note also that it's not just a reversion. The bullets are re-added later in the document.

SethTisue:
Note that the context for this is https://github.com/akka/akka-meta/issues/120 . Some of the feedback there was about balancing positivity and negativity, to keep the document from feeling too negative.

At the time we merged the PR inlining the Citizen CoC bullets, I don't think that's an angle we fully considered.

conduct.md
martijnhoekstra: I remain thoroughly opposed to defining harassment as including those bulleted points for no apparent benefit. It doesn't matter whether those things are included in our definition of harassment. What matters is that those things are unacceptable, regardless of the term "Harassment" does or doesn't include them.

rkuhn: This is a good point, @martijnhoekstra: “harassement” is a legal term already, and its definition does not match these bullet points in a legal system I know, so should we change the title to “Unacceptable Behavior”?

rkuhn:
Thanks for linking the context @SethTisue! I have opened another PR for addressing the second half of the feedback from Akka, see #1089.

conduct.md
patriknw: updated to that

conduct.md
```diff
@@ -33,19 +33,7 @@ As a member of the community, you agree to the following:
...
-  - Posting or threatening to post other people’s personally identifying information ("doxing").
-  - Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
-  - Inappropriate photography or recording.
-  - Inappropriate physical contact. You should have someone’s consent before touching them.
-  - Unwelcome sexual attention. This includes, sexualized comments or jokes; inappropriate touching, groping, and unwelcomed sexual advances.
-  - Deliberate intimidation, stalking or following (online or in person).
-  - Advocating for, or encouraging, any of the above behavior.
-  - Sustained disruption of community events, including talks and presentations.
-  - In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
+- We will exclude you from interaction if you insult, demean or harass anyone. See [examples of harassment](#examples-of-harassment) below. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
```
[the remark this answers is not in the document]
patriknw: updated to that

conduct.md
patriknw: changed to "Examples of unacceptable behavior"

patriknw:
I have incorporated feedback from @rkuhn .
Let me know if there are other opinions.

SethTisue:
In its current form, this PR contains three changes:

Change 1a: move the examples-of-bad-behavior down in the document, to their own section.

Change 1b: describe the examples as "unacceptable behavior" rather than as "harassment".

I think both of these are clear improvements. They are modest refinements to #1079 that bring the actual text in better alignment with what I understand to be the code's intent. Based on the discussion so far, there doesn't seem to be any objections, either, so this much is now mergeable, I think (though I suggest we wait until a total of at least two weeks have passed, to see if any further feedback comes in).

Change 2 replaces "In particular, we don’t tolerate behavior that excludes people in socially marginalized groups" with "We don’t tolerate behavior that excludes people based on their association with social, ethnic, religious, political, or similar groups".

I'm less comfortable with this change, and I might even be opposed to it. I'm not sure yet. It seems to me like a real change to the substance of the code, rather than a matter of improving how the Coe is expressed. Perhaps we could find _some_ change somewhat along these lines that I would be comfortable with? In any case, I would prefer that this part go in a separate PR, with a separate discussion.

conduct.md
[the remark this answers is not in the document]
SethTisue: (See my comment below where I've suggested moving this change to a separate PR.)

patriknw:
reverted Change 2 (marginalized groups)

rkuhn:
Sounds reasonable, let’s address one thing after the other. I’ll open a new PR to discuss the “In particular” issue once this one is merged (otherwise there would be conflicts).

SethTisue:
@milessabin @rossabaker @neko-kai objections or thoughts on the current form of this PR?

SethTisue:
thanks, all, for the thoughtful input on this

PR #1089, opened 2019-10-27
Make opening of Code of Conduct more friendly

rkuhn:
In the spirit of #1087 and the context of akka/akka-meta#120 I propose to reword the preamble and the _Encouraged_ section of the Code of Conduct to be more welcoming: the aim is to address the reader in a warm and friendly tone, using more explanations and less imperatives.

In the same vein I propose to remove the term “micro-aggressive” from the _Discouraged_ section since it is not well-defined — the examples given in that bullet point and the reference to the Recurse Center’s manual make it clear already what is meant, and even the Recourse Center does not use this term. Finding such a term in the instructions for interaction repels everyone who does not have a clear understanding of this term, it is exclusive.

**Update:** The second part has been removed, only the _Encouraged_ section improvement remains.


conduct.md
```diff
@@ -24,15 +21,14 @@ As a member of the community, you agree to the following:
...
-- Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
+- Avoid aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
```
SethTisue: I don't have a strong feeling about the word "micro-aggressive" either way, but I would like to know what the other original drafters of the CoC (@heathermiller, @robinske, @stewSquared — apologies if I'm forgetting anybody, it's been a while) think.

"Micro-aggressive" as a term is a rather recent invention and is arguably jargon. In that sense, I wouldn't mind losing it. But before losing it, I think we should consider whether the word was doing some work in the sentence that we ought to try and accomplish in some other way. Or maybe the rest of the paragraph already does that work? I'm not sure.

The reason the term "micro-aggressive" even exists is because of a dynamic which I believe is important and real. Namely, blatant open aggression isn't the only way an environment can be hostile. It can also be hostile if more subtle forms of aggression are repeated often enough over time. By simply deleting "and micro-aggressive", that shade of meaning is lost.

SethTisue: I don't have a strong feeling about the word "micro-aggressive" either way, but I would like to know what the other original drafters of the CoC (@heathermiller, @robinske, @stewSquared — apologies if I'm forgetting anybody, it's been a while) think.

"Micro-aggressive" as a term is a rather recent invention and is arguably jargon. In that sense, I wouldn't mind losing it. But before losing it, I think we should consider whether the word was doing some work in the sentence that we ought to try and accomplish in some other way. Or maybe the rest of the paragraph already does that work? I'm not sure.

The reason the term "micro-aggressive" even exists is because of a dynamic which I believe is important and real. Namely, blatant open aggression isn't the only way an environment can be hostile. It can also be hostile if more subtle forms of aggression (that, taken one example at a time, might seem not all that bad in themselves) are repeated often enough over time. By simply deleting "and micro-aggressive", that shade of meaning is lost.

conduct.md
```diff
@@ -24,15 +21,14 @@ As a member of the community, you agree to the following:
...
+- **Be kind and courteous.** Keep in mind that all public communication is received by many people you don’t know, so before sending a message please ask yourself whether someone from a different context would misunderstand it. Do not attack other persons, in particular consider that age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, or religion have no bearing on the respect you owe another person.
+- **Respect differences of opinion** and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer, we will find the best solutions by engaging in constructive discussion, with everybody bringing their unique viewpoint and experience to the table.
+- **Remember that everyone was new to Scala at some point.** We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Always assume good intentions and a willingness to learn, just as you are willing to evolve your own opinion as you gain new insights.
```
SethTisue: I like the addition of bolding here, it makes the code easier to read.

conduct.md
```diff
@@ -24,15 +21,14 @@ As a member of the community, you agree to the following:
 
 **Encouraged:**
 
-- Be kind and courteous.
-- Respect differences of opinion and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer.
-- Remember that everyone was new to Scala at some point. We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Assume competence.
-- Show empathy towards other community members.
```
SethTisue: What's the reason to drop "Show empathy towards other community members."?

conduct.md
```diff
@@ -10,11 +10,8 @@ redirect_from:
 
 ## Scala Code of Conduct
 
-We are committed to providing a friendly, safe and welcoming environment for
-all, regardless of age, body size, disability, ethnicity, sex characteristics,
-gender identity and expression, level of experience, education, socio-economic
-status, nationality, personal appearance, race, religion, sexual identity
-and orientation, or other such characteristics.
+We are committed to providing a friendly, safe and welcoming environment.
+We expect the following standards to be observed and upheld by all participants in any community forum (Discuss, GitHub, gitter, etc.).
```
SethTisue: There are multiple distinct changes happening in this part of the diff.

I like having "We are committed to providing a friendly, safe and welcoming environment" stand as a sentence by itself.

I am _not_ comfortable with reducing the emphasis on diversity so sharply. 

I don't understand why it's important to include a list of forums here (the "Discuss, GitHub, gitter" parenthetical). Why is it important to foreground that so prominently? (If we do include it, s/Discuss/Discourse` and `s/gitter/Gitter`. But I suggest we don't include it here. If it's missing from the document as a whole, is there a place further down it could go?)

conduct.md
```diff
@@ -24,15 +21,14 @@ As a member of the community, you agree to the following:
 
 **Encouraged:**
 
-- Be kind and courteous.
-- Respect differences of opinion and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer.
-- Remember that everyone was new to Scala at some point. We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Assume competence.
-- Show empathy towards other community members.
+- **Be kind and courteous.** Keep in mind that all public communication is received by many people you don’t know, so before sending a message please ask yourself whether someone from a different context would misunderstand it. Do not attack other persons, in particular consider that age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, or religion have no bearing on the respect you owe another person.
```
Ichoran: I think it would be better if the examples of encouraged behavior would actually be positive, inspiring examples instead of lists of things to avoid.  Alas, I don't have any good ideas for what to put here.  But as it is, this is "encouraged" in title only; it's mostly a list of inappropriate but all too common motivations for discouraged acts.

conduct.md
```diff
@@ -24,15 +21,14 @@ As a member of the community, you agree to the following:
...
+- **Respect differences of opinion** and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer, we will find the best solutions by engaging in constructive discussion, with everybody bringing their unique viewpoint and experience to the table.
```
Ichoran: The comma between `answer` and `we` should be a semicolon (or a period and a new sentence); otherwise it's a run-on sentence.  (You can't join independent clauses with a comma.)

conduct.md
Ichoran: The problem with listing micro-aggressive behavior is that, much like a stampede, the individual behaviors are relatively innocuous; it is only in aggregate that there can be a problem.  Trying to enforce "zero micro-aggressions" seems undesirable and impractical to me, since the battle over whether or not one thing was a micro-aggression would do way more damage to all involved than the micro-aggression would do to anyone (kind of by definition, or it isn't micro-).

But we already need to be kind and courteous, so we can catch the systematic patterns of behavior as a violation of that edict.

Instead, I'd say something like "Avoid behaviors that lead to strife and tension, such as unconstructive criticism, ...".  Wording like this puts the emphasis more on the overall picture (strife, tension) than a 200-post-long argument about whether an expression of surprise was feigned or genuine, and whether or not feigned surprise is actually a micro-aggression or not.

Ichoran: Also, phrasing it in terms of the consequence to mood involves the targets and onlookers, not just the person producing the behavior; so someone might be able to claim "I didn't intend comparing you to Microsoft Bob to be prejudicial!", but it's much harder to claim "everyone in the room who says that this felt hostile and uncalled-for is mischaracterizing how they feel".

conduct.md
Ichoran: It wasn't removed, just moved to the "kind and courteous" section.  Do you mean that you think it should be at the top for emphasis?  (I don't think it works all that well in the "encouraged" section TBH.)

SethTisue: > It wasn't removed, just moved to the "kind and courteous" section. Do you mean that you think it should be at the top for emphasis?

Yes, that's what I meant.

SethTisue: > I don't think it works all that well in the "encouraged" section TBH

and yes, that's a good additional point. I was primarily objecting to the reduction in emphasis, but even apart from that, I agree that the new location doesn't work

conduct.md
dwijnand: Was "sexual identity and orientation" dropped from this list on purpose?  I believe those to be distinct from anything else in the list, so if they were dropped because they're subsumed, I think that's a mistake.

conduct.md
Ichoran: After reading the discussion point from @rkuhn I understand the goal of the move: get the friendly stuff up, as that presents a more welcoming environment than a list of rules that we're scrutinizing you for.  (If you want to be friendly, "show, don't tell".  We should sound friendly!)

However, I don't think it works very well as implemented.  I think we can keep the positive stuff up front; on reflection it does sound awfully hostile and/or scary in the first few lines.  ("You're gonna discriminate against someone on the basis of personal appearance, aren't you?  Aren't you!  Don't you try it, buddy, we're watching your every move, with our fingers over the ban button." or "We say we're welcoming, but TBH all this horrible stuff happens every day on our forums, and we're soooooo fed up with it!  But, uh, you, um, will be fine, totally fine here now, cause uh, yeah, it's all good from now on; how are you?  (Fires blaster into mic.)")

We can reflect that we're serious about being welcoming to people of diverse backgrounds (who are often left out of "all" when being welcomed) by strengthening our wording later on when we've already established, by acting the part, that we're serious about being welcoming.  Then readers are more likely to feel that we've welcomed them in, and these things are as protections to keep things welcoming in practice.

We just need to find the right spot for it.

conduct.md
otto-dev: That would extend the list to:
- gender expression
- gender identity
- sexual identity
- sexual orientation
- sex characteristics

Effective and impactful communication is achieved by a few, carefully chosen words, not by many. May it sound counter-intuitive to some: If you want to communicate all of the points more strongly, I recommend to reduce the same to from five to two, namely "sexuality and gender". This, firstly, reads more welcoming than the alternative, secondly, communicates the points concisely and does not take away from the message, and thirdly, gives you equal grounds to enforce the same.

martijnhoekstra: I agree with the gist here, but if you want to capture the full list, you need one more: sexuality, gender and sex characteristics.

dwijnand: Could you explain why you think it reads more welcoming?

otto-dev: Sure. The language being used communicates not just objective facts, but also mind-set and atmosphere. Firstly, the atmosphere of the paragraph in question has already been well described by @Ichoran:

> it does sound awfully hostile and/or scary in the first few lines. "You're gonna discriminate against someone on the basis of personal appearance, aren't you? Aren't you! [...]"

I think this could in part be remedied by being less verbose in detailing inacceptible behaviour - while retaining the original meaning.

Secondly, to the mind-set. As I'm sure you are well aware, questions surrounding views on gender identity are highly controversial and political. For example, the controversy includes questions about the relationship between gender and sex, and is influenced by factors such as religion and culture. By making a bullet point list that distinguishes in detail between "gender expression", "gender identity", "sexual identity", "sexual orientation" and "sex characteristics" the document is making an implicit, yet unambiguous statement which side of the debate the reader is to hold, and which view is accepted in the community.

The Code of Conduct should prohibit discrimination on controversial grounds, and preclude this community from being a platform for such political debates, but it's language should not be prescriptive of, or siding with, political views, either implicitly, or explicitly.

conduct.md
rkuhn: Very well put, @Ichoran! In order to avoid github hiding this discussion as “outdated” I’ll wait a bit for the original drafters to chime in before applying your suggestion.

conduct.md
rkuhn: It felt like a duplication of the above points, now that they spell out what kindness and respect entail. If that is not the case, then please suggest contents for an accompanying paragraph as for the others.

conduct.md
rkuhn: Again, @Ichoran is spot on, and I agree that the current placement is not ideal — I copied it from the Idris community standards. I’ll try out different placements and make a new suggestion.

conduct.md
rkuhn: Indeed, sorry about that (in German this comma is just fine).

conduct.md
rkuhn: @dwijnand My removal was based on “seeing a duplicate”, an oversight. But I’m glad that I did it, because this discussion has shown further room for improvement, as @otto-dev aptly put it. Let’s make the CoC as precise and at the same time as little political as possible by shortening the list to what @martijnhoekstra suggests.

conduct.md
rkuhn: @SethTisue The list of forums was also inspired by the Idris community standard, but I agree that it detracts from the main focus. How about the updated proposal that only reminds people that there are online and offline forums?

conduct.md
```diff
@@ -24,15 +21,15 @@ As a member of the community, you agree to the following:
...
+- **Be kind and courteous.** We treat our fellow community members with the respect and dignity all humans deserve. Keep in mind that public communication is received by many people you don’t know, so before sending a message please ask yourself whether someone from a different context would misunderstand it.
+- **Respect differences of opinion** and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer; we will find the best solutions by engaging in constructive discussion, with everybody bringing their unique viewpoint and experience to the table.
+- **Remember that everyone was new to Scala at some point.** We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Always assume good intentions and a willingness to learn, just as you are willing to evolve your own opinion as you gain new insights.
 
 **Discouraged:**
 
+- Do not exclude people or reject their contributions based on age, body size, disability, ethnicity, sex characteristics, sexuality, gender, level of experience, education, socio-economic status, nationality, personal appearance, race, or religion.
```
rkuhn: I moved the anti-discrimination list to the top of the _Discouraged_ section, this should give it the right amount of emphasis and it also makes the _Encouraged_ section entirely positive.

conduct.md
rkuhn: I’ll add “empathy” to the first bullet point explicitly, then I feel that we definitely have it covered.

conduct.md
SethTisue: I think changing this list again is a distinct change that would need its own PR with its own discussion. I ask that we leave it out of this PR.

Note that we only just quite recently went around about this at https://github.com/scala/scala-lang/pull/1079#discussion_r326688217 . (The stopping point of that discussion was to use wording taken verbatim from the Contributor Covenant, a respected and widely adopted upstream source, rather than try to wordsmith our own.)

Ichoran: We can achieve the reorganization and make the document feel friendlier and more welcoming without changing the contents of this list, so I agree with Seth's point here.

I also think that otto has a good point about the implicit message you're sending by what exactly you put on a list.  But a separate PR would be a better place to discuss what the implications are both in mood and revealed political preference (of the original list, not the modification), and whether they're acceptable; and if not, how we should change them.

That we already talked about whether to adopt the Contributor Covenant language verbatim doesn't mean we shouldn't do so again, but it's a good idea to read the reasoning there so we don't need repeat it on a new PR (or here).

rkuhn: Good point, I’ll move the original list verbatim, then we can discuss in a separate PR.

conduct.md
rkuhn: @SethTisue Can we consider this resolved?

hepin1989: ~~regardless of age, body size, disability, ethnicity, sex characteristics,	We expect the following standards to be observed and upheld by all participants in any community forum, be that online or offline.
gender identity and expression, level of experience, education, socio-economic	
status, nationality, personal appearance, race, religion, sexual identity	
and orientation, or other such characteristics.~~

SethTisue: @rkuhn no, I'm afraid not. I repeat:

I am _not_ comfortable with reducing the emphasis on diversity so sharply.

In the original text, it's prominent up front that we care about this. in the new text, it's.... not _completely_ buried, but is certainly substantially less prominent than it was. I oppose this strongly.

I do not think the current placement is hostile or scary.

I definitely do not agree at all that this change can be characterized as making the CoC "more friendly". I think it makes it less friendly for groups of people who are underrepresented and discriminated against in tech. Being clear that we want Scala to be a friendly place for those people is one of the main goals of this code.

Ichoran: Can we switch the text under "our standards" to be first?  I think it's important to have a positive, uplifting message as the first full thought, and a list of "regardless of" groups isn't very effective at doing that.  If you want people to not worry about age discrimination, mentioning that you will accept them regardless of age _still makes them think about age in the context of discrimination_ which is not such a positive thing.

I don't think it needs to be pushed all the way into the don't-do-this category, as it can be phrased kinda-positively (and was in the original).

But as the very first thing we encounter, as a positive opening statement, it isn't very good.

stewSquared: > I don't think it needs to be pushed all the way into the don't-do-this category, as it can be phrased kinda-positively (and was in the original).

Agreed. Let's not make it a list of negatives. We could add another sentence before it that's more active, positive, and welcoming. Such a sentence could build motivation for the list. At present, it come up abruptly, but it would feel more natural if we made it more obvious that this opening paragraph is not about rules, but our goals/mission (things like expanding the community and providing space where members can flourish, etc).

> list of forums

I agree with Seth that it's best to leave this open. It's not just online and offline forums either. Conferences, meetups, and other events use this Code of Conduct. Typelevel and ny-scala use it verbatim.

@rkuhn  Consider for a second the audience for this piece. Who really needs to see this? Who is actively looking for the Code of Conduct? If that person is looking for this list, it should probably be closer to the beginning. It should read more like a goal and less like a rule or afterthought.

martijnhoekstra: If we want to keep an emphasis on us valuing diversity, maybe instead of the long list, we can say so explicitly here, in the form of "We are committed to providing a friendly, safe and welcoming environment that is accepting of diversity."

If we value diversity, we should at least provide an accepting environment, and say so explicitly.

otto-dev: Numerous valid concerns have been raised in regard to putting the list at the top in its current form. The concern for emphasis of diversity, may it be given a high priority, has to be incorporated into the argumentation that went before instead of simply dismissing the same.

If there are reasons to dismiss previous argumentation, or add to the same, those reasons themselves need to be communicated. As soon as the authority for what is permissible is based what one person or another feels comfortable with, instead of sound reason that can be displayed, justified and understood, a collaborative due process is no longer possible.

Ichoran: @otto-dev - If the list were the only thing that was improved here, then we should have a deeper discussion here about the content.  But it isn't the only thing, and that discussion is potentially a long one, so it's probably more effective to conservatively consolidate the other improvements we've made so far and save the real discussion of the list for yet another PR or issue.

You've made a reasonable point about lists potentially putting off people by what they include or exclude, and I see no evidence that this has been given careful consideration.  It should be.

However, ultimately I think there is not anywhere near enough evidence to unambiguously support a particular course of action, especially in light of people typically having a non-identical set of core values.  (Despite being non-identical, people's values are typically quite similar; the similarities tend to go unstated, leading us to think that they're rather different.  Operationally they're usually not.  But it can be enough to matter.)  So the point of creating robust argumentation cannot here be to deliver a decision that is independent of personal opinion.  Nonetheless, the opinion(s) can be informed by careful reasoning and incorporation of a variety of perspectives, rather than whatever people happened to feel about it before careful consideration.  (What they feel also is probably due in part to reasoning and consideration of some perspectives, but perhaps not as many as would be appropriate.)  And the reasoning thought to be most important in this case can be spelled out rather than leaving it ambiguous, implicit, or otherwise unstated.

otto-dev: Thanks, and agreed. I don't expect an unbiased outcome, but that there is an active *effort* to facilitate an unbiased process. The way this issue is handled thus far makes me question if possible biases that are contained in the document itself will be removed. Anyway, the future will show, let's leave this discussion alone, and focus on the issues at hand.

conduct.md
```diff
@@ -24,15 +21,15 @@ As a member of the community, you agree to the following:
...
+- **Respect differences of opinion** and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer; we will find the best solutions by engaging in constructive discussion, with everybody bringing their unique viewpoint and experience to the table.
+- **Remember that everyone was new to Scala at some point.** We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Always assume good intentions and a willingness to learn, just as you are willing to evolve your own opinion as you gain new insights.
 
 **Discouraged:**
 
+- Do not exclude people or reject their contributions based on age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, sexual identity and orientation, or other such characteristics.
```
xingyif: I disagree with this phrase because instead of saying "we are dedicated to create a safe community", it says "we shall not exclude anyone". They are very different concepts.
Diversity is never about tolerating everyone without limit. See https://en.wikipedia.org/wiki/Paradox_of_tolerance

The original phrase is sufficient.
>We are committed to providing a friendly, safe and welcoming environment for all, regardless of age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, sexual identity and orientation, or other such characteristics.

Since it is the goal of the Code of Conduct and the mission of the community, it should remain on top.

conduct.md
```diff
@@ -24,15 +21,15 @@ As a member of the community, you agree to the following:
...
-- Remember that everyone was new to Scala at some point. We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Assume competence.
-- Show empathy towards other community members.
+- **Be kind and courteous.** We treat our fellow community members with the empathy, respect and dignity all humans deserve. Keep in mind that public communication is received by many people you don’t know, so before sending a message please ask yourself whether someone from a different context would misunderstand it.
+- **Respect differences of opinion** and remember that every design or implementation choice carries a trade-off and numerous costs. There is seldom a single right answer; we will find the best solutions by engaging in constructive discussion, with everybody bringing their unique viewpoint and experience to the table.
+- **Remember that everyone was new to Scala at some point.** We want to encourage newcomers to join our community and learn the Scala language and ecosystem. Always assume good intentions and a willingness to learn, just as you are willing to evolve your own opinion as you gain new insights.
```
TimMoore: I also like the bolding and the expanded descriptions.

The changes to the "Encouraged" section don't seem to be controversial, and don't relate directly to the other changes. Maybe they could be merged first if they were split out to an independent pull request, while further discussion continues.

conduct.md
TimMoore: I agree with @xingyif. To me, this change doesn't seem to support the goal of making the document more friendly and welcoming. Some of the comments above said that the currently published text sound like they're scolding the reader for discrimination, but I feel like this change makes it sound _more_ like that. The original text is welcoming a diverse population into the community. This change makes it specifically about prohibiting discrimination, and only in a very narrow way (exclusion or rejection).

conduct.md
```diff
@@ -24,15 +21,15 @@ As a member of the community, you agree to the following:
...
-- Avoid aggressive and micro-aggressive behavior, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
+- Avoid behaviors that lead to strife and tension, such as unconstructive criticism, providing corrections that do not improve the conversation (sometimes referred to as "well actually"s), repeatedly interrupting or talking over someone else, feigning surprise at someone’s lack of knowledge or awareness about a topic, or subtle prejudice (for example, comments like “That’s so easy my grandmother could do it.”). For more examples of this kind of behavior, [see the Recurse Center's user manual](https://www.recurse.com/manual#sec-environment).
```
TimMoore: This change alters the meaning in a way I don't think was intended. "Micro-aggressive" is meant to describe subtle exclusion, whereas "strife and tension" might be interpreted to mean active conflict. That could lead people to think that as long as some behavior doesn't cause a flame war to break out, it's fine. We want people to also think about how their words and behavior might cause other problems such as people simply choosing not to engage in the community anymore because they were treated disrespectfully by other members.

I appreciate that many people might not be familiar with the term "microaggression". What does everyone think about this wording? "Avoid aggressive, condescending, or dismissive behavior..."

Ichoran: As I mentioned before, I think it's easier to police overall mood (strife, tension) than individual actions that depend in large part on the (difficult-to-know) intent of the speaker.  But I think "aggressive, condescending, or dismissive" is acceptable too.

Also, strife is active, but tension isn't, particularly.  How likely is it, do you think, that people would think it only applied to active conflict, especially given the examples?  Maybe the two of those are not enough, though; one can be excluded in ways that are low-strife and not very tense, but still very exclusionary (e.g. "cold shoulder").  On the other hand, every example causes either strife or tension or both.

I don't think unfamiliarity with the term "micro-aggression" is the problem with using the term.  The problem is that (1) it carries a connotation among some, deservedly or not, that it is expected that people will go into hysterics over a single tiny unintended or unknowable perceived slight; and (2) at least in my opinion it is implicitly demeaning to the targets of condescending etc behavior by suggesting that what's happening to them is "micro" (even though the originators of the term AFAICT intended the opposite).  Social exclusion and status-lowering behaviors can feel really bad (comparable to moderate levels of physical aggression, or people wouldn't be willing e.g. to get into fights over it, which they are), which is entirely reasonable given how critical social standing is to social creatures like us.  Anyway, we have lots of other better-established words that can convey pretty much anything we want to convey, both in literal meaning and in implication.  The only thing we'd miss is, "Hey, we're hip to this new terminology that's in fashion now."

SethTisue: I like Rex's unpacking of what's less-than-ideal about using the term "micro-aggression", thanks for that. So I think a good replacement wording is still worth trying for.

I had initially liked the "Avoid behaviors that lead to strife and tension" wording, but Tim's remarks give me pause.

maybe a slight rewording to "tension or strife" would help improve that version?

@TimMoore your version isn't bad, but it changes the emphasis in a way I think is probably not desirable. in "tension or strife", the emphasis is on the effect behavior has on others. But "aggressive, condescending, or dismissive" puts the emphasis on the behavior in itself, and I think that could encourage arguing about the definitions. It's better during enforcement to say things like "you're bothering others" than "your behavior was aggressive"; people will more readily accept the former, whereas the latter puts them on the defensive, in my experience.

conduct.md
rkuhn: @xingyif We agree on the goal of the CoC: to form the reference around which we create a welcoming community. With the proposed change, this mission statement remains front and center in the preamble, in fact it is strengthened by removing additional wording that says something else, namely the “regardless of <characteristics>” appendix.

The reason for moving that anti-discrimination list to a different part of the CoC is twofold: listing these legal-sounding terms does _not_ create a welcoming atmosphere for the reader, I posit that it has the opposite effect on most people. And as it is negative by nature, it belongs with the list of things to avoid, which is the _Discouraged_ section.

The precise wording of that sentence is up for debate. My reasoning behind the current proposal is that the unwanted effect of discrimination based on the listed characteristics is the rejection of exclusion of individuals or their contributions. If there are other effects that I am missing, please let me know.

I am also not happy with the list itself, as it is both very specific about certain parts while also being open-ended — but that is a debate we decided to postpone to a later PR.

@TimMoore The _Discouraged_ section has multiple bullet points and this is just the first of them — I assume that we do intend to make it clear that discrimination will not be tolerated? If so, then I don’t follow your argument. We also discussed earlier that the goal of this PR is to make the first part of the CoC purely positive to get people excited about the community; an anti-discrimination list does not contribute to this positivity.

conduct.md
rkuhn: @SethTisue I like your proposal of “tension or strife”, also because it is listed in escalating order.

One issue (perhaps the main one bothering me and others over here) is that “micro-aggression” is a foreign concept in Europe AFAIK — we tend to use established words to describe the effects. This makes me uncomfortable because I do not have a good intuition for what “micro-aggression” really means.

conduct.md
TimMoore: @rkuhn what I meant in my comment above is that the current wording is not an "anti-discrimination" list, it's an "explicitly welcome" list. It's primarily addressing the people who exist in various points of the multi-dimensional space it creates, particularly ones in areas that are frequently underrepresented in open source communities, or generally more subject to exclusion. Secondarily, it's reminding everyone to be sensitive to these differences.

This list was taken from the latest [Contributor Covenant](https://www.contributor-covenant.org/version/1/4/code-of-conduct). Here's how the Contributor Covenant FAQ [explains the list](https://www.contributor-covenant.org/faq#the-contributor-covenant-explicitly-lists-a-set-of-protected-classes-does-this-make-it-acceptable-to-discriminate-or-make-others-feel-unwelcome-based-on-other-factors):

> The Contributor Covenant explicitly lists a set of protected classes; does this make it acceptable to discriminate or make others feel unwelcome based on other factors?
>
> No. The Contributor Covenant explicitly lists protected classes for many reasons, such as reminding people to give them appropriate consideration, and assuring people in those protected classes that they are welcome. However, this is not an invitation for rules lawyers to seek loopholes, or to discriminate against others or make people feel unwelcome based on criteria not listed here. ([With the notable caveat that those who discriminate or make others feel unwelcome are themselves not welcome](https://en.wikipedia.org/wiki/Paradox_of_tolerance).)

To be honest, I consider it a small flaw in the Contributor Covenant that you need to also read the FAQ to fully understand the main document. I think we can improve on that here.

I'm open to the idea of reformulating the long list into something else that conveys the same intent in a different way. I'm not a huge fan of the list either: it will never be fully comprehensive, invites a lot of debate about what should or should not be listed, and distracts from the main point of the sentence. However, I think it is important to acknowledge the value of diversity and explicitly welcome people from underrepresented groups. That intent is what I don't want to lose.

If we can come up with a better wording for the introduction, then explicitly discouraging discrimination is also fine with me.

conduct.md
TimMoore: @SethTisue: good points. I agree that the motivation here should be to emphasize the effects on others in the community.

@rkuhn: also good points. I think it's fair to say that the term is probably not widely known anywhere, and people who do understand the term are likely to be more self-aware about it already. I acknowledge that it can be a controversial term, and can easily be misinterpreted, so I'm positive on being more explicit about the intent.

One thing that I like about this code of conduct is that it isn't only about "policing" egregious harm by people with bad intent, but also about encouraging members of the community to be aware of more subtle forms of exclusion and to take an active role in creating a more welcoming community.

People who have suffered a history of exclusion might approach a new community by looking for patterns of communication or behavior that fit their previous bad experiences. If they see indications of this, they might immediately turn away from the community, having no more patience for repeating that history. This is a bad outcome for the community. Does it qualify as tension or strife? I think we're missing something that captures that nuance. Maybe "marginalization, tension or strife"?

Unfortunately, @Ichoran I think your comment mischaracterizes the intent of the concept of microaggression and in doing so demonstrates some of the problem. From what I understand, the original intent was to talk about incidents that are individually small enough that those affected think "it's not worth making a big deal out of it", but have a cumulative effect of making them feel like second-class citizens. These problems are rarely raised to attention, especially because the people affected by them expect to be dismissed, or further stigmatized, if they do. Even the phrase "go into hysterics" is loaded with a history of exaggerated and discriminatory use that some people will interpret as a cue that reporting small problems in the community will be unwelcome, disrespected, and provoke a defensive overreaction. I would also say that "Hey, we're hip to this new terminology that's in fashion now" could be more generously interpreted as "Hey, we're educated about and sensitive to common causes of subtle harm that have been raised by people who have experienced them."

Ichoran: @TimMoore - I'm intentionally mischaracterizing the intent of the concept of micro-aggression because intent is not the only thing that determines how language is understood; among other things, the literal meaning, the cultural context, and the existing implicit associations of the words used, also have a major impact on how languages convey meaning.

People have been marginalizing, dismissing, mocking, teasing, stigmatizing, goading, maligning, negatively stereotyping, demeaning, disrespecting, etc. etc. each other for a very long time, and we consequently have a rich vocabulary with which to describe all manner of inappropriate and/or negative treatment.  The thing that "micro-aggression" adds to the vocabulary which is possibly useful is a general term to describe the category of individual acts that in aggregate cause a substantially negative experience.

However, it's not clear to me that there's any benefit for us to talk about the individual acts as opposed to the overall impact, and there are drawbacks to the terminology: unfamiliarity, and connotations that the term carries for some that we probably don't actually intend.  If we're talking about overall impact, we don't need the kind of generified quantization that "micro-aggression" provides.

Anyway, I think for now we should put the beginning of the document back _exactly the way it was_ and discuss the content of the list and the phrasing of the first few sentences in another place, so as not to indefinitely postpone the other helpful changes in this PR.

conduct.md
rkuhn: I’ve reverted the change to the preamble, to be improved in a follow-up PR.

conduct.md
rkuhn: I’ve reverted this change as well, to be improved in a follow-up PR.

rkuhn:
The PR now only contains the uncontentious part, I’ll open separate follow-ups for the two postponed topics.

rkuhn:
@mpilquist @xingyif @neko-kai With this reduced scope, do you maintain your thumbs-down votes?

mpilquist:
Latest revision looks good to me. :+1:

SethTisue:
thanks, all, for the thoughtful input on this

PR #1100, opened 2019-12-20
Code of Conduct: "Well actually"

JustinPihony:


SethTisue:
_well actually_ :-), I think the original is probably correct

it's a pretty unusual construction, taking a phrase in quotes and turning it into a noun, so I'm not completely confident what the "correct" orthography is, but any rate, it isn't a typo

JustinPihony:
I agree, and ran it past others. I guess typo is not quite right, but it sure looks awkward...even if maybe more accurate. I'm game one way or the other. Let me know if you prefer closing instead?

SethTisue:
I don't think I have ever heard "well actually" used as a noun phrase, and a little Googling seems to indicate that it isn't a common usage

so perhaps it would be best to simply remove the entire parenthetical? I don't think it adds much value if any; it may be confusing to non-native speakers, or others who aren't familiar with the "well actually" joke/meme; and in general, keeping the code short is better

lrytz:
> probably confusing to non-native speakers, or others who aren't already familiar with the "well actually" phrase/joke/meme

agree

Ichoran:
According to normal rules of English grammar, you can't pluralize a quoted statement at all (even if "Hello, world!"s is clear to pretty much any programmer, and our audience is programmers).  The rules aren't entirely consistent about quoting-with-apostrophe, but when it is done, it's done with a single word (not a phrase) and no quotes.  The advice, instead, is to reword to avoid such an awkward construct, and only use apostrophe to pluralize single letters.  (E.g. m's rather than ms, though of course programmers probably understand 'm's better!)

Since here, the awkward grammar is coupled with not-widely-understood-meaning, we should take the rewrite advice.  Probably by deleting the parenthetical phrase entirely, though it does form a nice cognitive break from an otherwise dull list, and provokes thought where otherwise people might just go, "yeah, whatever, so much blah blah blah".

SethTisue:
closing in favor of #1103, then

PR #1110, opened 2020-01-30
Replace references to SLIP with SIP

Philippus:
Applying this PR will replace the references to SLIP with SIP, as the SLIP process is no more I believe.


dwijnand:
You're right that the SLIP process is no more, but I'm not sure if the SIP process replaces it...

SethTisue:
yeah, the SIP committee doesn't take standard library proposals. (certainly not until after Scala 3.0 is out, anyway, at which point all of this might get revisited)

standard library changes are currently handled just by PRs. and anyway bincompat constraints rule out almost any controversial change

for now, maybe about the best we could do is link to https://github.com/scala/scala-dev/issues/661

SethTisue:
thanks Philippus!

PR #1115, opened 2020-02-23
Remove Scala IDE For Eclipse from homepage

louwers:
Scala IDE has not been updated for [two years](https://github.com/scala-ide/scala-ide/tree/master) and it is not compatible with the current Scala version. The [mailing list](https://groups.google.com/forum/#!forum/scala-ide-user) also does not show much activity.

It is not a good idea to advertise it on the homepage anymore, lest newcomers (like me) be disillusioned by Scala from the outset. So I remved it with this PR. See also [the discussion](https://users.scala-lang.org/t/why-is-scala-ide-mentioned-on-the-scala-homepage/5656) on the forum.


SethTisue:
Definitely agree with the change. Does it render all right?

Philippus:
I tried it out and it renders fine, thank you and welcome @louwers !

PR #1117, opened 2020-02-24
Announce new version of progfun2

julienrf:
Ready to review but we need to wait for Coursera to effectively publish the new version before we merge the PR :)


julienrf:
@SethTisue Thanks for your review, I’ve added a commit accordingly: https://github.com/scala/scala-lang/pull/1117/commits/406c4d0adb41930835921119be3ce5fc3f89b40b

PR #1120, opened 2020-03-02
Announce the developer survey results

julienrf:
I’m not sure whether we should also write a small analysis of the results or not.
I must say that I don’t have much to say except paraphrasing the charts…

### DIFF
--- a/_posts/2020-03-04-developer-survey-results.md
+++ b/_posts/2020-03-04-developer-survey-results.md
@@ -1,0 +1,12 @@
+---
+category: news
+by: Julien Richard-Foy, Scala Center
+title: "Results of the 2019 Scala Developer Survey"
+---
+
+You have been about 4 thousands Scala developers to fill the survey and we are grateful
+to you! The survey results help us identify the trends in how you use the language in
+your day-to-day workflow, and what could be improved in the tooling or ecosystem.
+
+We are happy to share the results with you. They are available at
+[https://scalacenter.github.io/scala-developer-survey-2019/](https://scalacenter.github.io/scala-developer-survey-2019/).


_posts/2020-03-04-developer-survey-results.md
```diff
@@ -0,0 +1,12 @@
+---
+category: news
+by: Julien Richard-Foy, Scala Center
+title: "Results of the 2019 Scala Developer Survey"
+---
+
+You have been about 4 thousands Scala developers to fill the survey and we are grateful
+to you! The survey results help us identify the trends in how you use the language in
```
SethTisue: The first sentence reads awkwardly to a native speaker, and there isn't a spot fix; it needs rearranging. Perhaps simply:

> Four thousand Scala developers filled out the survey. We are grateful to you all!

I would also suggest a paragraph break at this point.

_posts/2020-03-04-developer-survey-results.md
```diff
@@ -0,0 +1,12 @@
...
+your day-to-day workflow, and what could be improved in the tooling or ecosystem.
```
SethTisue: This is more of a nitpick, but after a comma, you must begin a new independent clause (with its own verb). It's fixable by omitting the comma, but then the sentence as a whole becomes somewhat hard to parse. It's a bit hard even with the comma. Perhaps aid the reader with bullets:

> The survey results help us identify:
> * trends in how you use the language in your daily work
> * what could be improved in tooling or the ecosystem

(note also the slight wording adjustments)

_posts/2020-03-04-developer-survey-results.md
```diff
@@ -0,0 +1,12 @@
...
+
+We are happy to share the results with you. They are available at
+[https://scalacenter.github.io/scala-developer-survey-2019/](https://scalacenter.github.io/scala-developer-survey-2019/).
```
SethTisue: Especially since you haven't chosen to share your own thoughts about the result, I would suggest starting a users.scala-lang.org topic with a link to the results, and then close the post with a link to that topic as a suggested place for discussion.

It would be nice to have the discussion in one place and not spread around Twitter — not that that won't happen too :-)

Not entirely sure which Discourse is better, users or contributors. A thread like this is likely to turn into a big, diffuse pile-on, not unlike the notorious "What could make Scala more popular?" thread on contributors. There are downsides either way. A big pile-on might be off-putting to newcomers, especially if it gets contentious. But it's also awfully noisy for contributors. Sometimes I actually miss having a separate scala-debate group :-/

Regardless, I think we should have the thread, discussions are good despite sometimes getting big and/or contentious. I lean towards using users this time, but I'm okay either way.

SethTisue:
Is it expected that the results link is 404 at present?

julienrf:
> Is it expected that the results link is 404 at present?

Yes, I didn’t publish the website.

_posts/2020-03-04-developer-survey-results.md
julienrf: I also think users.scala-lang.org is a great place. Now I’m thinking that we should just write something there, and not publish the news on scala-lang. Also, the website currently just shows the raw results, we might publish a deeper analysis in a month or so. If we do that, then I find little value in publishing a blog on scala-lang.org with just a link to the users forum, with then another link to the website that actually sows the results.

What do you think of 1) closing this PR, 2) publishing an answer to the users thread with a link to the results? If we do a deeper analysis of the results then we could publish it as a blog here.

SethTisue: Hmm, tricky. Let's discuss in person!

SethTisue:
https://twitter.com/scala_lang/status/1234815280800509952

PR #1127, opened 2020-03-19
htmlproofer: ignore 401 and 429 results

SethTisue:
bringing this repo in line with docs.scala-lang

prompted by a lot of 429 errors turning up at
https://travis-ci.org/github/scala/scala-lang/builds/664541189

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-# opt-in to Travis new infrastructure
-sudo: false
-
 language: ruby
 rvm:
   - 2.3.1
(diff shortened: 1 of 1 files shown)


SethTisue:
merging on the fast track to get CI builds green, but post-merge feedback is welcome

ashawley:
FWIW, I recall the ignoring of 401 status codes on docs.scala-lang.org was done in desperation to fix the strange errors with www.oracle.com that may have not worked.

ashawley:
Is the 429 because GitHub is complaining about getting crawled?  If so, I guess it's good it's only done as a scheduled cron job.

SethTisue:
> FWIW, I recall the ignoring of 401 status codes on docs.scala-lang.org was done in desperation to fix the strange errors with www.oracle.com and it may have not worked.

thx, 492c144235ddc886b32bcd4d67abed0635e7b4f6 cuts it back to just ignoring 429

> Is the 429 because GitHub is complaining about getting crawled?

beats me

ashawley:
One route would be asking GitHub in a support ticket.  However, they'd probably say go fly a kite.

PR #1130, opened 2020-03-20
One click install for Scala

sjrd:
Update the download page with the brand new one-click install developed by @alexarchambault for the Scala Center, and add a blog post to give it visibility :)


_layouts/downloadpage.html
```diff
@@ -16,60 +16,65 @@
...
+              <p>The two most popular IDEs for Scala are IntelliJ and VS Code with Metals.</p>
             </div>
           </div>
           <div class="download-options">
             <div class="download-intellij">
               <span class="or">or</span>
               <div class="description">
                 <img src="/resources/img/download/arrow-left.png" alt="">
-                <p>Best if you prefer a full-featured IDE (recommended for beginners)</p>
+                <p>Best if you prefer to start without build tool (recommended for beginners)</p>
```
Philippus:
```suggestion
                <p>Best if you prefer to start without a build tool (recommended for beginners)</p>
```

sjrd:
Here is a rendering of the download page for Windows:
image

_layouts/downloadpage.html
```diff
@@ -16,60 +16,65 @@
...
               </div>
-              <a href="" class="btn-download" id="download-sbt-link">
+              <a href="" class="btn-download" id="download-vscode-link">
                 <i class="fa fa-download"></i>
-                <span>Download Sbt</span>
+                <span>Download VS Code</span>
               </a>
               <ul>
-                {% include tutorial-list.html column=0 %}
+                <li><a href="https://scalameta.org/metals/docs/editors/vscode.html"><i class="fa fa-file-text-o"></i>Getting Started with Scala in VS Code</a></li>
```
julienrf: I would not qualify this tutorial as “getting started in Scala” tutorial. It seems to target developers already familiar with Scala. I think eventually we need to create a proper tutorial, like the ones below.

julienrf: So maybe it’s simpler to leave this section as it used to be and do the changes that you suggest when we have a proper “getting started” tutorial with VS Code + Metals?

_posts/2020-03-23-one-click-install.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+which will install all the following software, if not yet installed:
+
+* a JDK
+* the build tools [sbt](https://www.scala-sbt.org/) and [mill](https://www.lihaoyi.com/mill/)
+* the [Ammonite](https://ammonite.io/) enhanced REPL
+* [scalafmt](https://scalameta.org/scalafmt/), the Scala formatter
+* the [coursier](https://get-coursier.io/) CLI, to install further Scala-based applications
+* the `scala` and `scalac` command-line tools
```
julienrf: Should we add some details about how we manage specific versions of `scala` and `scalac`, and how to update the `cs` tool itself?

lrytz:
For me it's the first time I looked at the `cs` native launcher a little bit closer, also the first time I looked at `cs setup`, `cs install` and `cs java`. Since the download page also targets beginners, maybe mostly beginners, that's probably a good starting point for a review 🙂

First of all, `cs` looks really great, I'm looking forward to this change!

I worry about how "magic" it all seems when just looking at the scala-lang webiste, and coursier's CLI install page that is linked. The "one click" / "one command" to set up everything is the perfect starting point, but after doing that, people will wonder
  - where does it install all the stuff?
  - how does it modify my environment?
  - what java (edition, version) does it install?
  - ...

These things matter, even to begginers - a Scala beginner is not necessarily programmer beginner. Even more, `cs` is not only for beginners, so experts should feel confident to start using it!

Of course we cannot explain these things on the download page, but it should be clear that `cs setup` is not a one-shot script that does some stuff to your system and then leaves you alone. We should explain what `cs` is and roughly how it works. Maybe also improve the documentation on the coursier website, because if scala-lang->download links to https://get-coursier.io/docs/cli-overview.html#install-native-launcher, then that page really becomes an entry point into the coursier website.

Things that are unclear to me after browsing through the coursier website
  - How to install a specific version, and switch between versions? I think this is important at least for scala, not everyone wants 2.13 right now for example.
  - What is `latest.release`? Does cs always automatically upgrade to the latest release? Is that a good idea?
  - `cs` can install `cs`, right? Can i delete the native launcher after `cs setup`? What if I used `brew install`?

Finally, for me the download of the native launcher was extremely slow, i tried on mac and windows (it wasn't due to my connection).

image

image

_layouts/downloadpage.html
```diff
@@ -16,60 +16,65 @@
...
+              <p>The two most popular IDEs for Scala are IntelliJ and VS Code with Metals.</p>
             </div>
           </div>
           <div class="download-options">
             <div class="download-intellij">
               <span class="or">or</span>
               <div class="description">
                 <img src="/resources/img/download/arrow-left.png" alt="">
-                <p>Best if you prefer a full-featured IDE (recommended for beginners)</p>
+                <p>Best if you prefer to start without a build tool (recommended for beginners)</p>
```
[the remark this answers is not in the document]
sjrd: Reverted

_layouts/downloadpage.html
sjrd: Reverted to the old content with sbt.

sjrd:
@alexarchambault I guess @lrytz's comment is best addressed by you?

_posts/2020-03-23-one-click-install.md
sjrd: Done as well.

sjrd:
Here is a new rendering:
image

SethTisue:
I worry a bit if this stuff — which is still rather new — has been tested enough, on a wide enough variety of setups, to be safe enough to recommend to newcomers.

Should I worry, or relax?

sjrd:
At least for Windows, it works better than any of the individual tools' install mechanisms. It's really just double-clicking on the exe, and it installs everything, and then everything just works.

It may not have been tested by a large number of people, true. We specifically made a call for testers for Windows a week or two ago. For other systems, I don't know.

tpolecat:
What do non-Windows users see?

julienrf:
Here is a screenshot of what it looks like from a Linux machine:

Screenshot from 2020-03-23 17-38-14

tpolecat:
I agree with comments from @lrytz. I want to know what version of what is installed where … when I type `java` is it the same VM that's used when I call `scala`? How do I switch versions? Will all my tooling also switch versions with me (Ammonite for example, which is tied to a specific Scala version)? And so on. This seems great but it will prompt a lot of questions.

sjrd:
@julienrf Hum that's not normal. Those are the generic instructions that it falls back on if it can't tell the OS. It should display something like this:

    $ curl -Lo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup

Did you get any error in the browser's console?

julienrf:
No, but it seems that the os detection is not very good. It is based on `window.navigator.appVersion`. On my OS its value is “5.0 (X11)”, which is then identified to be “unix”, that’s why I get these instructions rather than the “linux” instructions.

Maybe that function is more robust? https://stackoverflow.com/questions/38241480/detect-macos-ios-windows-android-and-linux-os-with-js With this one, my system is correctly identified as Linux.

julienrf:
@alexarchambault Would it be possible to update the documentation of coursier to answer all the questions raised by Lukas? Eventually, we need to copy parts of this documentation in the blog post and in the getting started page.

_layouts/downloadpage.html
```diff
@@ -16,23 +16,24 @@
...
         <div class="install-steps">
           <div class="step">
             <img src="/resources/img/download/arrow-asset.png" alt="">
             <div class="number-step">1</div>
             <div class="text-step">
-              <h3>First, make sure you have the Java 8 JDK (or Java 11 JDK) installed.</h3>
-              <p>To check, open the terminal and type:</p>
-              <p><code>java -version</code><span>(Make sure you have version 1.8 or 11.)</span></p>
-              <p><i>(If you don't have it installed, download Java from <a href="https://www.oracle.com/java/technologies/javase-jdk8-downloads.html">Oracle Java 8</a>, <a href="https://www.oracle.com/java/technologies/javase-jdk11-downloads.html">Oracle Java 11</a>, or <a href="https://adoptopenjdk.net/">AdoptOpenJDK 8/11</a>. Refer <a href="https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html">JDK Compatibility</a> for Scala/Java compatiblity detail.</i></p>
+              <h3>First, install Java, Scala, and all related command-line tools.</h3>
```
exoego: I think the link to JDK compatibility, that was recently added in https://github.com/scala/scala-lang/pull/1123, should be preserved, so new comers do not get trouble when they use latest JDK and older Scala combination.

alexarchambault:
Thanks for your comments everyone. I'd like to give some more context before answering @lrytz's and @tpolecat's points in more detail.

### About `cs setup`

The `cs setup` command is just some "repackaging" of features of the `cs java` and `cs install` commands. One doesn't have to run `cs setup`, it's just a convenience - for beginners in particular. By relying on commands that more advanced users can benefit from, `cs setup` should be kept up-to-date and more maintained in the future.

The `cs java` command (managing JVMs) and `cs install` (managing applications) are quite independent of each other. One can use one and not the other. One can keep managing JVMs another way, while using `cs install`. Or one can use JVMs from `cs java` while not using `cs install`. It's also possible to use these two commands while not installing anything globally - not changing profile files / user environment variables, and keeping everything in the coursier cache, see below.

### About `cs java`

`cs java` manages JVMs. For example, running `cs java --jvm 11 -version` will automatically download the latest AdoptOpenJDK 11 (in the coursier cache), unpack it (in the [managed JVM directory](https://get-coursier.io/docs/cli-java.html#managed-jvm-directory)), and run it with `-version`:
```bash
$ cs java --jvm 11 -version
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.6+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.6+10, mixed mode)
```
 It uses the [index](https://github.com/shyiko/jabba/blob/8c8e6be29610a3d5ea505087a791e9a57f6e48a6/index.json) of jabba to know where to download JVM archives, and assumes AdoptOpenJDK if only a version is passed.

`cs java-home` prints the Java home of a JVM, like
```bash
$ cs java-home --jvm 9
~/Library/Caches/Coursier/jvm/adopt@1.9.0-0/Contents/Home
```

One can [update](https://get-coursier.io/docs/cli-setup.html#how-it-sets-environment-variables-globally) profile files (Linux / macOS) or user environment variables (Windows) for a specific JVM with `--setup`, like
```bash
$ cs java --jvm openjdk:1.14 --setup
```
(this updates `JAVA_HOME` and `PATH`)

`cs setup` does basically that, for the default JVM (AdoptOpenJDK 8).

But you don't have to update these environment variables globally. For example, one can just get a Java home with
```bash
$ cs java-home --jvm graalvm:20
~/Library/Caches/Coursier/jvm/graalvm@20.0.0/Contents/Home
```
or set environment variables only for the current session with
```bash
$ eval "$(cs java --env --jvm 11)"
$ java -version
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.6+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.6+10, mixed mode)
```

These two commands only write things in the coursier cache, and the [managed JVM directory](https://get-coursier.io/docs/cli-java.html#managed-jvm-directory), which is a cache too (both of these directories live under `~/.cache/coursier` on Linux, `~/Library/Caches/Coursier` on macOS).

## About `cs install`

The `cs install` command creates launchers for applications, in the [installation directory](https://get-coursier.io/docs/cli-install.html#installation-directory) (`~/.local/share/coursier/bin` on Linux). One can pass `--install-dir` to `cs install` and `cs setup` to install launchers in another directory.

One can run `cs install --setup` to [update](https://get-coursier.io/docs/cli-setup.html#how-it-sets-environment-variables-globally) profile files (Linux / macOS) or user environment variables (Windows) (this adds the installation directory to `PATH`). `cs setup` does that too.

But one can also rely on applications… while not installing them, with `cs launch`, like
```bash
$ cs launch scala:2.12.10
Welcome to Scala 2.12.10 (OpenJDK 64-Bit Server VM, Java 11.0.6).
Type in expressions for evaluation. Or try :help.

scala> :q
```
or
```bash
$ cs launch ammonite:2.0.4 --scala 2.12.8
cs launch ammonite:2.0.4 --scala 2.12.8
Loading...
Welcome to the Ammonite Repl 2.0.4 (Scala 2.12.8 Java 1.8.0_121)
@
```

Lastly, one can run applications with specific JVMs with
```
$ cs launch --jvm 11 --scala 2.12.9 ammonite:2.0.4
…
```
(this particular command suffers from [a regression](https://github.com/coursier/coursier/issues/1635) as of writing this, should be fixed soon).

lrytz:
Thank you @alexarchambault for the summary! I did manage do get most of what you descibe by reading/skimming through the coursier website. For someone that's completely new to coursier, or coming from the scala download page, this kind of summary should be the first thing that's shown.

A few specific questions:
  * Is it a good idea that `cs install scala` doesn't support picking a version, and that it auto-upgrades? My intuition is that this should be more manual (like what I'm used to from homebrew).
  * After running `cs setup` for the first time, `cs` installs itself right? Should we recommend to delete the download?
  * Running just `cs` on mac does nothing - should it display `--help`? On windows, just running `cs` re-runs `cs setup`. This should probably not be the default once you have it installed on the command line. So maybe it would be a good idea if the one-click-installer was a separate executable that does `cs setup` when double-clicked, but then the `cs` version installed by `cs` doesn't do that, but shows `--help`?
  * The native launcher downlaods are really slow - can this be improved?

alexarchambault:
> Is it a good idea that cs install scala doesn't support picking a version, and that it auto-upgrades?

You can pick version with `cs install`, like
```bash
$ cs install scala:2.12.10
```

For now, running `cs update` will bump it to the latest version though. That needs to be fixed. (Is is what you meant by "auto-upgrades"?)

> After running cs setup for the first time, cs installs itself right? Should we recommend to delete the download?

We can, yes, and maybe should actually.

> Running just cs on mac does nothing - should it display --help?

Good point, I guess it should, that needs to be fixed.

> On windows, just running cs re-runs cs setup. This should probably not be the default once you have it installed on the command line. So maybe it would be a good idea if the one-click-installer was a separate executable that does cs setup when double-clicked, but then the cs version installed by cs doesn't do that, but shows --help?

That's a good point. It's the same launcher for now only because it's easier (and faster) to generate and upload one launcher rather than two, from the coursier CI (via [this script](https://github.com/coursier/coursier/blob/99082e2aeb5ed0c40d39477475367b41f1d39499/launcher.sc#L210)).

> The native launcher downlaods are really slow - can this be improved?

Maybe we should use `curl` rather than `bitsadmin` in the instructions to download it. `bitsadmin` is supposed to download things with a low priority somehow. And recent versions of Windows are supposed to have `curl` pre-installed.

alexarchambault:
To answer @tpolecat's points:

> I want to know what version of what is installed where …

Currently, this is kept in some metadata files written alongside launchers (the launchers are actually zip files, these metadata are under `META-INF/coursier/` in them, even for the native applications). Maybe we could add a command to list installed applications with their versions.

> when I type java is it the same VM that's used when I call scala?

Yes. The installed application launchers, and the installed JVMs, are quite independent from each other. They aren't even aware of each other. The app launchers will pick whatever version of java is installed on the current machine (via `PATH`, `JAVA_HOME`, …) They will pick a JVM installed by `cs setup` or `cs java` only if it's already setup via `PATH` or `JAVA_HOME`.

> How do I switch versions?

For the `scala` launcher, it may be possible to offer to change the scala version by relying on the scripts from @dwijnand's [scala-runners](https://github.com/dwijnand/scala-runners). For now, the `scala` launcher starts the version that was choosen upon installation, which is the latest by default.

But it's possible to run specific versions via `cs launch`, like
```bash
$ cs launch scala:2.12.0
…
$ cs launch scala:2.12.0 -- …options…
…
```

> Will all my tooling also switch versions with me (Ammonite for example, which is tied to a specific Scala version)?

No. The launchers are independent of each others. In more detail, they're tiny Java applications, with a list of URLs in a resource. Upon launch, they read that URL list, find the corresponding JARs in the coursier cache, and start that. So you can have `scala` start scala `2.12.10` (if you did `cs install scala:2.12.10`) and Ammonite use scala `2.13.1`. Both pick their JARs in the coursier cache, just like sbt or mill do for library dependencies.

tpolecat:
@alexarchambault :+1: thanks for the explanations. As @lrytz suggests, I think it would be helpful to display this information prominently, and/or make it available via `--help` for specific commands. Because this is a new way to install and manage things it's important for people to understand what exactly is going on.

alexarchambault:
> I think it would be helpful to display this information prominently, and/or make it available via --help for specific commands.

Agreed… I'll try to fix that.

lrytz:
> You can pick a version with `cs install`

Great, I didn't realize.

> For now, running `cs update` will bump it to the latest version though. That needs to be fixed. (Is is what you meant by "auto-upgrades"?)

No, I thought the only possibility was to say `cs install scala` (without version), and that the installed `scala` would always dynamically pick the latest release on maven. Since that's not the case, that's fine. `cs update` updating to the latest makes sense I think. Maybe the possibility to pin a version would be useful.

Another feature that would be useful is `cs list` to show what's installed, the versions (similar to `brew list`).

> Maybe we should use `curl` rather than `bitsadmin` in the instructions to download it

If that helps, sure. But I have to say also on my mac the download was really slow (using `curl`).

julienrf:
@sjrd @alexarchambault I believe that in order to move forward you need to enrich the documentation of coursier and the announce here with the additional details you’ve provided in the discussion.

alexarchambault:
FYI, I reworked the CLI docs on the coursier web site. You can have a preview of it [here](https://get-coursier.io/docs/next/cli-overview). I'm planning on tweaking things a bit more (add more details about the `install` command on that page).

julienrf:
@alexarchambault Thanks for the improvements! On my side, I still find it a bit confusing that we use a different name for the the “native” launcher vs the jar-based launcher. Why not always call it `cs`? Also, it would be good to know the exact list of [applications](https://get-coursier.io/docs/next/cli-setup.html#applications) that are installed by `cs setup`. Is `cs` installed by `cs setup`? If that’s the case, maybe add a note about it in the [Installation](https://get-coursier.io/docs/next/cli-installation) instructions so that users know that they don’t need to keep the `cs` file they’ve just downloaded?

sjrd:
@lrytz @tpolecat Does the new documentation pointed out by @alexarchambault at https://github.com/scala/scala-lang/pull/1130#issuecomment-611448013 addresses the issues you had with it?

tpolecat:
@sjrd yes, the doc is greatly improved, thanks! :+1:

_posts/2020-03-23-one-click-install.md
```diff
@@ -0,0 +1,33 @@
...
+For example, on Linux, all we now need is:
+
+```bash
+$ curl -Lo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup
+```
+
+which will install all the following software, if not yet installed:
+
+* a JDK
+* the build tools [sbt](https://www.scala-sbt.org/) and [mill](https://www.lihaoyi.com/mill/)
```
eed3si9n, line 22: Regarding sbt, would this install the same launcher script and launcher JAR as what one would get using downloading the installer or using SDKMAN?

alexarchambault: This installs a launcher for the [coursier-based sbt launcher](https://github.com/coursier/sbt-launcher), based on [this "app descriptor"](https://github.com/coursier/apps/blob/d21889a73797cbef3d0f8b0742945b054ac58c0c/apps/resources/sbt-launcher.json).

You can generate one locally with `cs bootstrap sbt-launcher -o sbt` for example.

eed3si9n: I think it would be very misleading to call any alternative packaging of sbt as `sbt` on novice user's PATH since inevitably they will run into some incompatibilities of handling SBT_OPTS environment variable or dependency resolution difference etc, and then whoever in the position of supporting the person in-house, or in open source would have to reverse-engineer "ok. what `sbt` script are you using, and how did it get to your PATH?"

I'm more than happy to use Coursier for all three layers (launcher, plugins, proper build), but I think it should happen in conjunction with the official (preloaded) distribution.

eed3si9n: Here's an example in Bash script incompatibility (likely parsing of SBT_OPTS):

```
$ ./sbt
[info] Loading settings for project global-plugins from idea.sbt,vimquit.sbt,plugins.sbt,pgp.sbt ...
[info] Loading global plugins from /Users/eed3si9n/.sbt/1.0/plugins
[info] Loading project definition from /Users/eed3si9n/work/sbt-coursier/project/project
[info] Loading settings for project sbt-coursier-build from plugins.sbt ...
[info] Loading project definition from /Users/eed3si9n/work/sbt-coursier/project
[info] Loading settings for project sbt-coursier-root from build.sbt ...
[info] Set current project to sbt-coursier-root (in build file:/Users/eed3si9n/work/sbt-coursier/)
[error] Expected ';'
[error] -Xms2048M
[error]          ^
```

alexarchambault: The output of `cs bootstrap sbt-launcher -o sbt` (basically, what is installed by `cs setup` or `cs install sbt-launcher`) doesn't read `SBT_OPTS`. Not sure how you can get that error.

If you used the [custom sbt-extras script](https://github.com/coursier/sbt-extras/blob/master/sbt) linked from the [README](https://github.com/coursier/sbt-launcher/blob/master/README.md), it should basically be the same handling as [sbt-extras mainline](https://github.com/paulp/sbt-extras).

eed3si9n: So there are now 3 flavors of `sbt` script: official, sbt-extra, and coursier/sbt-launcher? If it's not byte-for-byte equivalent to the official distribution, please call it something else on the PATH, like `csbt` to avoid confusion.

In some future timing, like sbt 1.4.x or whatever, the official one can switch over to something that uses Coursier by default, so if someone runs into a bug / incompatibility we can address it together. sbt releases make changes not only to the core artifacts, but also to Bash / Batch files.

For example, `sbt` supports `--supershell=never` as Bash option. Passing that to `csbt` you'd get:

```
[error] Expected ';'
[error] --supershell=never
[error]                   ^
```

I ❤️ the overall initiative to make CLI setup easier, but putting forked implementation on PATH is not helpful.

sjrd:
Version RC6-11 (second to last atm) works.
/cc @alexarchambault

alexarchambault:
@sjrd https://github.com/coursier/coursier/pull/1681 should fix that. You'll need to re-download `cs.exe` once a release with https://github.com/coursier/coursier/pull/1681 has been cut (and run `.\cs install cs` if ever you had run `cs update` and the `cs` binary in your PATH doesn't work anymore).

sjrd:
Thanks! Was this a Windows-only issue? What would be the cost of making sure this doesn't happen anymore? The link for the download automatically point to the latest release. Once we make it the default, we cannot afford the released installer to be broken like that.

alexarchambault:
> Was this a Windows-only issue? 

Yes, the code that crashed is only run on Windows (it's related to encoding some powershell code in base64 prior to running it…)

> What would be the cost of making sure this doesn't happen anymore?

Right now, the native launchers are only generated upon release, as these take time to generate. I'm thinking about running some tests before pushing them, and / or maybe run those on a daily basis, so that we can catch regressions not too long after they occurred.

_posts/2020-03-23-one-click-install.md
alexarchambault: The coursier-based launcher works in a majority of cases, and also allows to be installed just like any other JVM apps here, with no ad-hoc handling. To me, it's fine if there's a discrepancy between the coursier-based launcher and the others for a short amount of time…

Plus it shouldn't differ that much from sbt-extras, apart from the Java options handling (it only accepts those via `JAVA_OPTS`, just like the other apps installed by `cs setup` or `cs install`). And I would expect those discrepancies to shrink even more if the coursier-based launcher were to use native-image - as it would run the `java` command itself, it could handle Java options too for example, kind of like what you described in https://github.com/sbt/sbt/issues/5406#issuecomment-615401865.

alexarchambault:
About 
> On windows, just running `cs` re-runs cs setup. This should probably not be the default once you have it installed on the command line. So maybe it would be a good idea if the one-click-installer was a separate executable that does cs setup when double-clicked, but then the cs version installed by cs doesn't do that, but shows --help?

this should be addressed in the upcoming release. It'll show the help message when run with no arguments, rather than not printing anything.

On Windows, the installed `cs` will also show the help message when run with no arguments. Some environment variables are set right before the installed native launcher starts, that allows it to detect if it's `install`-ed (show the help) or not (assume the user double-clicked on it, and run `setup`).

_posts/2020-03-23-one-click-install.md
SethTisue: it came up on Gitter just now that somebody was using the coursier-based launcher and wasn't getting their `.jvmopts`, in case there's any doubt this will bite people in practice

SethTisue: ah, the user in question opened a ticket: https://github.com/coursier/sbt-launcher/issues/167

alexarchambault: FYI, `cs setup` now installs an sbt launcher based on [sbt/launcher](https://github.com/sbt/launcher). If you already installed `sbt` via `cs setup` or `cs install`, remove it and install `sbt` again, like
```text
$ cs uninstall sbt
$ cs install sbt
```

It still doesn't read `.jvmopts` though. It handles a few of the options of sbt-extras via [coursier/sbt-runner](https://github.com/coursier/sbt-runner) (that parses some options upfront, then calls the main class of sbt/launcher).

lrytz: This is great!

Is there a reason `cs update sbt` would not work?

alexarchambault: Yes, that's because the name of the "app descriptor" changed from [`sbt-launcher`](https://github.com/coursier/apps/blob/8e124db6395253e5ece202d6bf1dd47cae922170/apps/resources/sbt-launcher.json) to [`sbt`](https://github.com/coursier/apps/blob/8e124db6395253e5ece202d6bf1dd47cae922170/apps/resources/sbt.json). The uninstall-then-install steps allow to pull the newer app descriptor.

Just `sbt` now works, as if one uses it with an explicit version, it does correspond to an actual sbt version as expected. This gives `cs launch sbt:1.3.10 -- new scala/scala-seed.g8`. This works because the new app descriptor references `org.scala-sbt:sbt-launch` directly. With the former one, the explicit version would have applied to just coursier/sbt-launcher, [whose versions](https://github.com/coursier/sbt-launcher/releases) are not in sync with sbt. `cs launch sbt-launcher:1.2.21` felt slightly less odd than `cs launch sbt:1.2.21`. That problem went away now.

alexarchambault:
Just a heads up, to tell that I think most issues around the `sbt` launcher were addressed.
- It now relies on [sbt/launcher](https://github.com/sbt/launcher). It's much closer to the standard sbt launcher now.
- It now supports options like `--supershell=…`. This relies on some custom option parsing logic, via [this project](https://github.com/coursier/sbt-runner), that runs before the main class from sbt/launcher. This logic doesn't come straight from [sbt/sbt-launcher-package](https://github.com/sbt/sbt-launcher-package), but it can be evolved along sbt/sbt-launcher-package. coursier/sbt-runner has been discussed [here](https://github.com/coursier/coursier/issues/1714).
- It supports reading java options from `.jvmopts`.

You need to install / update `sbt` with the latest `cs` for that to work, like
```text
$ cs update cs
$ cs
Coursier 2.0.0-RC6-21
…
$ cs update sbt
Updated sbt
```
(If the last step found no new update, do `cs uninstall sbt && cs install sbt` to re-generate the `sbt` launcher.)

alexarchambault:
We've been gathering feedback about the Windows launcher [here](https://twitter.com/alxarchambault/status/1239931282127273985), and introduced the `setup` command to a wider audience [here](https://contributors.scala-lang.org/t/easier-scala-installation/4326).

_posts/2020-03-23-one-click-install.md
```diff
@@ -0,0 +1,33 @@
...
+layout: blog-detail
+post-type: blog
+by: Sébastien Doeraene
+title: "One-click install for Scala"
+---
+
+Installing Scala has always been a task more challenging than necessary, with the potential to drive away beginners.
+Should I install Scala itself? sbt? Some other build tools? What about a better REPL like Ammonite? Oh and before all that I need to install Java?
+
+To solve this problem, the Scala Center contracted Alexandre Archambault in January 2020 to add a one-click install of Scala to [coursier](https://get-coursier.io/docs/cli-overview).
```
julienrf, line 11: I’m not sure about “one-click install of Scala **to** coursier”. Should it be “through coursier” instead?

sjrd:
Following all the discussions and a suggestion of @julienrf, I have removed the changes to the download page, and only kept the blog post, so that we can move on by announcing the existence of the one-click install, encourage some people to test it more widely, but keep from recommending it by default until it's been tested more.

SethTisue:
Perhaps the blog post could open with some kind of indication of what lies ahead. And/or some kind of call-to-action for the community; what do we want people to do with this information?

_posts/2020-03-23-one-click-install.md
```diff
@@ -0,0 +1,32 @@
+---
```
julienrf, line 1: Before we merge we need to update the date :)

julienrf:
> Perhaps the blog post could open with some kind of indication of what lies ahead. And/or some kind of call-to-action for the community; what do we want people to do with this information?

Actually, I think the first paragraph already does that.

SethTisue:
scala_lang tweet?

julienrf:
Yes, good point. Fixed :) https://twitter.com/scala_lang/status/1278249350523748352

martijnhoekstra:
> Following all the discussions and a suggestion of @julienrf, I have removed the changes to the download page, and only kept the blog post, so that we can move on by announcing the existence of the one-click install, encourage some people to test it more widely, but keep from recommending it by default until it's been tested more.
> 
> I have kept the changes to the download page in a separate branch for safekeeping: https://github.com/sjrd/scala-lang/tree/one-click-install-by-default

This is the future calling. Maybe the time is now.

bishabosha:
https://alexarchambault.github.io/posts/2020-09-21-cs-setup.html, very good information about the usage

PR #1148, opened 2020-06-30
upgrade rack and other dependencies

SethTisue:
as usual with Ruby stuff I have no idea what I'm doing, but we got
a security alert about our rack version, CVE-2020-8184

so I ran `bundle update rack-jekyll` and this is the result


SethTisue:
Ima hit "merge" because security (even if theoretical), we can always revert

SethTisue:
this doesn't seem to have broken anything, but didn't fix the underlying issue either; see #1150

PR #1155, opened 2020-07-24
describe Discourse primarily as a forum

SethTisue:
this brings the wording at the top of
https://www.scala-lang.org/community/ in line with the other wordings
farther down the page


SethTisue:
(I would also be happy to just say "forums" everywhere, since the body text has "You can participate via the web, or you can use “mailing list mode”, where you receive posts in your inbox and can reply to them via email." I'm not sure it needs more emphasis than that. But for now anyway I've chosen the more modest change.)

PR #1157, opened 2020-08-19
Jekyll 3.9.0 upgrade, also kramdown upgrade

SethTisue:
this is comparable to scala/docs.scala-lang#1756.  the motivation is
the same: we got a security alert about kramdown.

but it went a bit differently in this repo. if I did `bundle update
jekyll` that went all the way to some 4.x version. I don't know why.
but it seems safer to do a smaller upgrade first. `bundle update
--minor jekyll` got us to 3.9.0, but then kramdown didn't get updated,
so I followed it with `bundle update kramdown` (after adding
`kramdown-parser-gfm` to `Gemfile`, as in the other PR)

since I didn't do a full `bundle update`, I didn't need to revert
any version bump of jekyll-redirect-from

so.... it's not great that this doesn't bring the two repos exactly
in sync, but oh well. I'm trying to resolve the security issue without
putting excess effort into it.

I wouldn't oppose followup PR(s) that reduce the deltas between the
repos.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'rack-jekyll'
 gem 'rake'
 gem 'jekyll-redirect-from'
 gem 'html-proofer'
+gem 'kramdown-parser-gfm'
 
 # group :jekyll_plugins do
 #   gem 'hawkins'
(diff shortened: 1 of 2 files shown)


SethTisue:
site seems okay — happy to respond to any post-merge feedback

PR #1163, opened 2020-10-08
remove incorrect bit about partest

SethTisue:


som-snytt (approved):
A true Hacktoberfest contribution. Happy Hackoween!

PR #1186, opened 2020-11-30
Write blog post about new tuples in Scala 3

vincenzobaz:


_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+This behavior is desirable in many cases, for example when
+we want a function to return two or more values or when we want
+to use types to decide which transformation to apply to each
+value.
+
+# How do tuples evolve in Scala 3 ?
+
+## Size limit
+
+Probably the most well know limitation of tuples in Scala 2 was the
```
sjrd, line 55:
```suggestion
Probably the most well known limitation of tuples in Scala 2 was the
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+```scala
+scala> (1, "2", 3.0, List(4))
+val res0: (Int, String, Double, List[Int]) = (1, 2, 3.0, List(4))
+```
+This behavior is desirable in many cases, for example when
+we want a function to return two or more values or when we want
+to use types to decide which transformation to apply to each
+value.
+
+# How do tuples evolve in Scala 3 ?
```
sjrd, line 51:
```suggestion
# How do tuples evolve in Scala 3?
```
In most languages using the latin script, there are no spaces before punctuation marks. French is an exception.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+for (i <- 0 until someStuff.size)
+  println(someStuff(i))
+```
+This was possible in Scala 2 with the `productIterator` although this
+produced a value of type `Iterator[Any]` which means that we had to pattern
+match or eventually cast the type of the elements.
+
+This brings us to the conceptual change that we will explore in the
+next change: tuples become a collection of data that we can manipulate
+and program against.
```
sjrd, line 95: I don't really understand this sentence. We could already program with tuples and manipulate them before.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+as Shapeless in Scala 2, which was a complicated task for new
+Scala developers.
+
+These operations are now available in the standard library, they
+are safe and preserve the individual types of each element.
+
+The first one was already introduced: `.size` retrieves the number
+of elements in the tuple.
+
+Let's see:
```
sjrd, line 111: ?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Tuples can grow
+
+We can add an element to a tuple using the `*:` operator,
+which is very similar to the `::` combinator available on `List`.
+
+```scala
+val fourWeirdNumbers = (1, "2", 3.0, List(4))
```
sjrd, line 119: Calling them numbers, even weird ones, is a bit distracting, because `List` and `Set` don't fit there. Consider not naming the concept. Just use `fourElements`.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+
+These operations are now available in the standard library, they
+are safe and preserve the individual types of each element.
+
+The first one was already introduced: `.size` retrieves the number
+of elements in the tuple.
+
+Let's see:
+
+### Tuples can grow
```
sjrd, line 113: Not really. That would imply mutability. But we can build larger tuples from smaller ones.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+val thisIsTrue = fourWeirdNumbers == evenWeirder // true
+
+val fiveWeirdNumbers = Set(0) *: evenWeirder // (Set(0),1,2,3.0,List(4))
+```
+
+When we use a tuple as argument of `*:`, it is prepended as a single element:
+```scala
+val notGood: ((Int, Int), Int, Int) = (1, 2) *: (3,4) // ((1, 2), 3, 4)
+```
+So how can we concatenate two tuples ?
```
sjrd, line 131:
```suggestion
So how can we concatenate two tuples?
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+forgets the type of the elements.
+It is also possible to use `.toIArray` which has exactly the same behavior
+but produces an `IArray` where the `I` stands for immutable.
+```scala
+scala> List(1, "2").toArray
+val res0: Array[Object] = Array(1, 2)
+```
+
+I believe however that the most interesting conversion is `toList`
+which produces yes a list, but not a `List[Any]`.
```
sjrd, line 178: That sounds weird. Consider dropping the "yes a list but not ...", and instead directly says it produces a `List[U]`.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+Instead it creates `List[U]` where `U` is the union type of the types
+of the elements of the tuple. 
+That is:
+
+```scala
+(1, "2", 3.0)
+val res0: List[Int | String | Double] = (1, "2", 3.0).toList
```
sjrd, line 185: Was that supposed to be a REPL transcript? It seems some things have truncated a bit too far.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+// The code compiles without errors or warning
+// the compile verified that we handled all possible cases
+twoWeirdNumbers.toList.map {
+  case i: Int => (i + 2).toString
+  case j: String => j
+}
+```
+Read more about [Union types here](https://dotty.epfl.ch/docs/reference/new-types/union-types.html).
+
+We can also transform a tuple by appling a function to each element.
```
sjrd, line 209:
```suggestion
We can also transform a tuple by applying a function to each element.
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+With tuples each element has a different type!
+How can we generalize the concept of a function whose argument type is
+not fixed ?
+We can use a **`PolyFunction`**. This is a more advanced syntax:
+
+```scala
+twoWeirdStrings.map[[X] =>> String]([T] => (t: T) => t match {
+  case i: String => i.toUpperCase
+  case j: Int => j.toString
+})
```
sjrd, line 223: I don't think that's a convincing example, because that would work just as well with `.toList.map`. Consider a version where you put every element in a container such as a `Some`:
```scala
twoWeirdStrings.map[[X] =>> Option[X]]([T] => (t: T) => Some(t))
```
and show that the result type is `(Option[Int], Option[String], Option[Double])`.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+
+```scala
+def malicious[L <: Tup, R <: Tup](left: L, right: R): Concat[L, R] = left
+// This does not compile!
+```
+
+We can use an extension method to allow users to write `(1, 2) ++ (3, 4)` instead
+of `concat((1, 2), (3, 4))`, I believe that you now know how to do this too.
+
+We can use the same approach for other functions on tuples, I invite you to have
```
sjrd, line 372:
```suggestion
We can use the same approach for other functions on tuples, I encourage you to have
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+a look at the source code of the standard library to see how the other operators are
+implemented.
+
+# Conclusion
+
+We had a look at the new operations that the are available on tuples in Scala 3 and at
```
sjrd, line 378:
```suggestion
We had a look at the new operations that are available on tuples in Scala 3 and at 
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+of `concat((1, 2), (3, 4))`, I believe that you now know how to do this too.
+
+We can use the same approach for other functions on tuples, I invite you to have
+a look at the source code of the standard library to see how the other operators are
+implemented.
+
+# Conclusion
+
+We had a look at the new operations that the are available on tuples in Scala 3 and at 
+how a more flexbile type system provides the the fundamental tools to implement safer
```
sjrd, line 379:
```suggestion
how a more flexbile type system provides the fundamental tools to implement safer
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+and readable code.
```
sjrd, line 380:
```suggestion
and more readable code.
```
?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+
+This shows how advanced type combinators in Scala 3 enable to create
+APIs that benefit developers no matter their level of proficiency in the language:
+an expert-oriented feature such as dependently match types enable to build a safe
+and simple operation such as tuple concatenation.
```
sjrd, line 385:
```suggestion
This shows how advanced type combinators in Scala 3 allow to create
APIs that benefit developers no matter their level of proficiency in the language:
an expert-oriented feature such as dependently match types allow to build a safe
and simple operation such as tuple concatenation.
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+As most of indexes are 0 based in Scala, this brings more consistency
+to codebases. It also provides more flexibility. We can, for example,
+*iterate* on any tuple to print each element on a line:
+
+```scala
+val someStuff = (1, "2", 3.0, List(4))
+for (i <- 0 until someStuff.size)
+  println(someStuff(i))
```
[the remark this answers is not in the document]
julienrf: By the way, we should stress that supplying a wrong argument (e.g. `someStuff(-1)`) fails at compile-time.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+How can we generalize the concept of a function whose argument type is
+not fixed ?
+We can use a **`PolyFunction`**. This is a more advanced syntax:
+
+```scala
+twoWeirdStrings.map[[X] =>> String]([T] => (t: T) => t match {
+  case i: String => i.toUpperCase
+  case j: Int => j.toString
+})
+```
```
julienrf, line 224: I think a parametric example would be more appropriate here. For instance a function that turns a value of any type `A` into a value of type `Option[A]`.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+That is a tuple is either empty, or an element `head` which precedes
+another tuple. Using this recursive definition we can create
+a tuple as follow:
+
+```scala
+import Tup._
+
+val myTup = TCons(1, TCons(2,  EmpT))
+```
+It is not very pretty, but it can be easily be adapted to provide
```
julienrf, line 270:
```suggestion
It is not very pretty, but it can easily be adapted to provide
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,385 @@
...
+Again, similarly to conversion methods on collections, it is possible to
+transform a tuple into another collection.
+
+We have to pay attention to the type of the resulting collection.
+Let's start with the simple case: as its name might hint,
+`toArray` produces an array. The type of its elements will always be
+`Object`. This makes it easy to reason about this method although it
+forgets the type of the elements.
+It is also possible to use `.toIArray` which has exactly the same behavior
+but produces an `IArray` where the `I` stands for immutable.
```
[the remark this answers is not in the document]
sjrd: It's called `IArray` in the Scala stdlib. If you think it should be changed, you should submit an issue to lampepfl/dotty.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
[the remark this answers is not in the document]
vincenzobaz: I did try it out, you can play with it https://scastie.scala-lang.org/UkruyR1nSdClyyJPMoxynQ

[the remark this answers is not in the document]
bishabosha: I think if you do not supply a constant then the result of indexing will just be the unreduced match type

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: I am trying to say that in Scala 2 tuples are static: we cannot add terms, we cannot drop terms, it is hard to iterate over them etc... while in Scala 3 they are more similar to collections

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: I do not master the syntax of `PolyFunction` unfortunately.
I cannot seem to get the following snippet to compile:
```scala
val value = (1, 'a', "dog", 3.0).map[[X] =>> Option[X]]([T] => (t: T) => t match {     
  case i: Int => Some(i+1)                                               
  case i: Double => Some(i+1)                                               
  case _: _ => None                                             
}
```
@bishabosha do you have an idea?

julienrf: The `=>>` type operator is for “type functions” (or type constructors), not polymorphic function types. See https://scastie.scala-lang.org/FLUV7345T6KR2WNCGkJqlw

~~~ scala
val f: [A] => A => Option[A] =
  [A] => (a: A) => Some(a)
val items: (Int, String, Boolean) = (5, "foo", true)
val items2: (Option[Int], Option[String], Option[Boolean]) = items.map(f)
~~~

bishabosha: It seems that it is a problem that with the map to option example that the branches do not unify `T` with each branch, such as `Int` or `String`, this is the same in a normal def

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+This is better explained with an example:
+```scala
+scala> List(1, "2", 3.0, List(4))
+val res0: List[Any] = List(1, 2, 3.0, List(4))
+```
+We see that the compiler tries to infer a common supertype for the elements of the list,
+in this case `Any`.
+
+If do the same with tuples, the elements maintain their individual and specific type:
```
Philippus, line 41:
```suggestion
If we do the same with tuples, the elements maintain their individual and specific type:
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+```scala
+scala> (1, "2", 3.0, List(4))
+val res0: (Int, String, Double, List[Int]) = (1, 2, 3.0, List(4))
+```
+This behavior is desirable in many cases, for example when
+we want a function to return two or more values or when we want
+to use types to decide which transformation to apply to each
+value.
+
+# How do tuples evolve in Scala 3?
```
SethTisue, line 51: slightly unidiomatic — “did” instead of “do” would fix it, but using “evolve” at all emphasizes the transformation process rather than emphasizing the end result. I think simply “How are tuples better in Scala 3” would be more direct

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+In Scala 3, we can use the `apply` method with a 0-based argument:
+
+```scala
+("First", "Second")(1) // "Second"
+```
+
+As most of indexes are 0 based in Scala, this brings more consistency
+to codebases. It also provides more flexibility. We can, for example,
+*iterate* on any tuple to print each element on a line:
```
SethTisue, line 82: iterate over, not iterate on

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+("First", "Second")._2 // "Second"
+```
+
+In Scala 3, we can use the `apply` method with a 0-based argument:
+
+```scala
+("First", "Second")(1) // "Second"
+```
+
+As most of indexes are 0 based in Scala, this brings more consistency
```
SethTisue, line 80: will `_1` etc be deprecated in the future? if not, why not?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+to codebases. It also provides more flexibility. We can, for example,
+*iterate* on any tuple to print each element on a line:
+
+```scala
+val someStuff = (1, "2", 3.0, List(4))
+for (i <- 0 until someStuff.size)
+  println(someStuff(i))
+```
+
```
SethTisue, line 89: I'm  a bit uncomfortable with this example because it seems like something one shouldn't do. What is the _purpose_ of allowing indexed access like this? Is the main purpose consistency? Or is the main purpose that we actually expect such looping-over-a-tuple to be common? I doubt it will be common, and so using it too prominently in an example runs the risk of confusing readers about what tuples are for.

If iterating over a tuple is uncommon but something one might _occasionally_ want to do, then I think it would be good to clearly signal that.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+The argument provided to `apply` is checked at compile time. This means that
+**`someStuff(-1)` or `someStuff(4)` will result in a compilation error**.
+
+This was possible in Scala 2 with the `productIterator` although this
+produced a value of type `Iterator[Any]` which means that we had to pattern
+match or eventually cast the type of the elements.
+
```
SethTisue, line 96: I don't understand the "eventually" here

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: Since the birth of Scala, we've been emphasizing to people learning the language that lists and tuples are completely different things used for completely fferent purposes. Now it seems like the story has changed and they're becoming more interchangeable, more overlapping in purpose. _Why, and under what circumstances, is that even good?_,  is the main thing I feel is missing from this blog post.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+### Tuples can shrink
+
+Similarly to operators available on lists, we can retrieve a subset of
+a tuple. Here is a quick overview:
+
+ - `drop` allows to ignore the first *n* elements of the tuple, returning
+   an empty tuple when the number of elements is smaller than *n*:
+```scala
+(1, "2", 3.0, List(4)).drop(2) // (3.0, List(4))
```
SethTisue, line 147: Can I `.drop(n)` or can I only use a literal integer?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+   of the original tuple and the second contains the remaining elements
+```scala
+(Set(0), 1, "2", 3.0, List(4)).splitAt(3) // ((Set(0), 1, "2", 3.0), (3.0, List(4)))
+```
+
+### Tuples can transform
+
+Again, similarly to conversion methods on collections, it is possible to
+transform a tuple into another collection.
+
```
SethTisue, line 166: s/another collection/a collection/

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+We have to pay attention to the type of the resulting collection.
```
SethTisue, line 167: Do you mean "we must choose the type of the resulting collection", or do you mean something else?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+Let's start with the simple case: as its name might hint,
+`toArray` produces an array. The type of its elements will always be
+`Object`. This makes it easy to reason about this method although it
```
SethTisue, line 170: better to say `AnyRef` than `Object`, IMO

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+but produces an `IArray` where the `I` stands for immutable.
+```scala
+scala> (1, "2").toArray
+val res0: Array[Object] = Array(1, 2)
+```
+
+I believe however that the most interesting conversion is `toList`
+which produces yes a list, but not a `List[Any]`.
+Instead it creates `List[U]` where `U` is the union type of the types
+of the elements of the tuple.
```
SethTisue, line 182: I'd suggesting linking to the documentation of union types, since they aren't in Scala 2 and not all readers will be familiar with them yet.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+(1, "2", 3.0)
+val res0: List[Int | String | Double] = (1, "2", 3.0).toList
+```
+This is interesting because the type information is somehow maintained.
+We can iterate over `list` and use pattern matching to apply the
+correct transformation, knowing exactly how many and what cases to
+treat:
+
+```scala
+// The compiler tells he cannot help with checking:
```
SethTisue, line 195: s/tells he/says it/

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+We can also transform a tuple by appling a function to each element.
+The method, similarly to what we are used to with collections, is called
+`map`. The difference from collections (or functors) is however
+that they expect a `f: A => B` where `A` is the type of the elements
+of the collection.
+With tuples each element has a different type!
+How can we generalize the concept of a function whose argument type is
+not fixed ?
+We can use a **`PolyFunction`**. This is a more advanced syntax:
```
SethTisue, line 219: there should be a link here to something that would explain to me what a PolyFunction is

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+```scala
+(1, 'a', "dog", 3.0).map[[X] =>> Option[String]]([T] => (t: T) => t match {
+  case i: String => i.toUpperCase
+  case j: Int => j.toString
+})
+```
+
+## Tuples can compose
+
+The last operator allows to pair the elements of two tuples.
```
SethTisue, line 230: s/operator/operation/

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+```scala
+(1, 'a', "dog", 3.0).map[[X] =>> Option[String]]([T] => (t: T) => t match {
+  case i: String => i.toUpperCase
+  case j: Int => j.toString
+})
+```
+
+## Tuples can compose
+
```
SethTisue, line 229: personally, I don't think of `zip` as "composing" anything

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+val numbers = (1, 2, 3, 4, 5)
+val letters = ('a', 'b', 'c')
+
+numbers.zip(letters) // ((1, 'a'), (2, 'b'), (3, 'c'))
+```
+
+# Under the hood: new type operators of Scala 3
+
+I believe that the core new features that allows such a flexible
+implementation of tuples are **match types**.
```
SethTisue, line 245: Earlier in the blog post it said "compiler feature", and I wondered what that meant.

Now, some pages later, I'm finding out that it refers mainly to match types. I'd suggest simply saying so in the intro, rather than keep the reader in suspense.

Also I would call match types a language feature, not a "compiler feature". (`-Xlint` and `-rewrite` are examples of compiler features.)

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+a tuple as follow:
+
+```scala
+import Tup._
+
+val myTup = TCons(1, TCons(2,  EmpT))
+```
+It is not very pretty, but it can be easily adapted to provide
+the same ease of use as the previous examples.
+To do so we can use another exciting new Scala 3 feature: [extension methods](http://dotty.epfl.ch/docs/reference/contextual/extension-methods.html)
```
SethTisue, line 274: Couldn't I have done the same thing in Scala 2 with an `implicit class`? It's fine to call attention to the nice new syntax, but as written, it leads the reader to believe this wouldn't have even been possible in Scala 2.

If it's just a syntax difference, it isn't "exciting" in this context.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+```scala
+def concat[L <: Tup, R <: Tup](left: L, right: R): Concat[L, R] =
+  left match
+    case _: EmpT.type => right
+    case cons: TCons[head, tail] => TCons(cons.head, concat(cons.tail, right))
+```
+
+We use here a combination of match types and a form of dependent types called
+*dependently match types*. There are some quirks to it as you might have noticed:
```
SethTisue, line 360: isn't it just "dependent match types"? (I don't actually know — but “dependently match types” isn't correct English)

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+of `concat((1, 2), (3, 4))`, I believe that you now know how to do this too.
+
+We can use the same approach for other functions on tuples, I invite you to have
+a look at the source code of the standard library to see how the other operators are
+implemented.
+
+# Conclusion
+
+We had a look at the new operations that are available on tuples in Scala 3 and at
+how a more flexbile type system provides the fundamental tools to implement safer
```
SethTisue, line 381: s/flexbile/flexible/

SethTisue (commented):
I've made a number of specific nitpicks/corrections, but really my main feedback here is that I feel that the reader should be helped to understand why it would even be desirable to iterate over a tuple, or take the tail of a tuple, and so forth. I always thought of these operations as things one should not even _want_ to do with a tuple. Why is it good that I can do them now?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
julienrf: I don’t have this feeling. I have seen multiple blog articles, talks, or libraries about `HList`s (which is what Scala 3 tuples really are), and I think their usage is quite common in type-level programming. But I agree that we should explain that better in the motivation part of this blog article.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
+---
+layout: blog-detail
+post-type: blog
+by: Vincenzo Bazzucchi, Scala Center
+title: Flexible and safe tuples in Scala 3
+---
+
+# Flexible and safe tuples in Scala 3
+
+Tuples are completely revisited in Scala 3.
```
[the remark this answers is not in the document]
Philippus: maybe "completely revised"?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: No, the repl output is too verbose.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: Good example, I will use it

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: I am already using enums and match types, I did not think it would be necessary to mention `implicit`

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: I totally agree, compiler feature is not the appropriate expression. Thanks for catching it.
Some suspense is nice in all stories no ? :smile:

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: Great idea

vincenzobaz: I can't seem to find a nice explanation in the docs. Did you have a specific resource in mind ?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: I wrote `Object` because it is the output I got in the REPL. Why do you think that `AnyRef` is better ?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz:
```
scala> val n = 1
val n: Int = 1

scala> val tuple = (1, "2", 3.0)
val tuple: (Int, String, Double) = (1,2,3.0)

scala> tuple.drop(n)
val res3: Tuple.Drop[(Int, String, Double), n.type] = (2,3.0)
```

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,387 @@
...
+
+In this post we will explore the new capabilities of tuples before
+looking under the hood to learn how the improvements in the Scala 3
+compiler enable to implement type safe operations on tuples.
+
+# The basics: what are tuples?
+
+In the Python programming language, tuples are a simple concept:
+they are immutable collections of objects. As such, they are opposed
+to lists, which are mutable.
```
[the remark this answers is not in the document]
vincenzobaz: I started with Python because it is the only main stream language providing tuples. I could have started with Rust or Haskell but probably they are less well known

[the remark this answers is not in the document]
SethTisue: (It's not so much the choice of Python specifically I'm reacting to, but the choice to lead with a comparison to any other programming language at all. Anyway, no big deal.)

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: (I think simply dropping the words "exciting new" would address my concern. Anyway, no big deal.)

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
```diff
@@ -0,0 +1,382 @@
...
+
+# Flexible and safe tuples in Scala 3
+
+Tuples are revisited and completely rethought in Scala 3.
+They are more **flexible**, more dynamic and support a **wider range of operations**.
+This is enabled by new and powerful language features.
+
+In this post we will explore the new capabilities of tuples before
+looking under the hood to learn how the improvements in the Scala 3
+type system enable to implement type safe operations on tuples.
```
SethTisue, line 16: Oh, I just noticed: "enable to implement" doesn't work grammatically, you need either "enable us to implement" or "enable implementing".

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: Ok, so you _can_ do it, but instead of getting an ordinary tuple, I get something called a `Tuple.Drop` — what is that?

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: `AnyRef` is Scala, `Object` is Java. On the JVM, we have `type AnyRef = Object`, but that's an implementation detail. Scala books, the Scala spec, and so forth all talk about `AnyRef`, not `Object. 

It is unfortunate that the Scala 3 REPL sometimes does this dealiasing and shows you `Object`. The Scala 2 REPL does so much less often; it's a regression.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: At https://dotty.epfl.ch/docs/reference/overview.html it says:

> Polymorphic Function Types generalize polymorphic methods to dependent function values and types. Current status: There is a proposal, and a prototype implementation, but the implementation has not been finalized or merged yet.

and then it just links to https://github.com/lampepfl/dotty/pull/4672 rather than to a proper documentation page

So it seems the documentation is out of date. I don't actually know what the current status of the feature is, actually — I guess you'd have to ask someone on the Dotty team.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: I leave this same feedback about "suspense" on basically everyone's blog posts :-)

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
SethTisue: Is the difference that application authors won't normally need to manipulate tuples in these new ways, but library authors might?

julienrf: Yeah, that's probably more common to library authors.

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: `Tuple.Drop <: Tuple` is a match type describing the result of the drop operation. This is the pattern that I show in the second part of the post, reimplementing `++`

vincenzobaz: https://github.com/lampepfl/dotty/blob/f57055cfb60682bf34858bd06b7e65d97a77399c/library/src/scala/Tuple.scala#L180

_posts/2020-11-30-flexible-and-safe-tuples-in-scala3.md
vincenzobaz: Okay, I will use `AnyRef`, hoping that the repl will be fixed

vincenzobaz:
Thank you @SethTisue @Philippus and @julienrf for the thorough and attentive review.

I understand that the posts needs to be reworked to better motivate *why* the features are interesting and to better define the target audience.
Any pointer or idea is of course very appreciated.

In the meanwhile I will close the PR.

SethTisue:
I wonder (I genuinely don't know) if looking at how https://underscore.io/books/shapeless-guide/ motivates the analogous material in Shapeless might help.

PR #1191, opened 2020-12-09
in footer, point GitHub link to whole org

SethTisue:
with Scala 3 arriving soon, and the dotty repo likely moving
to the Scala org, this seems better to me

scala/scala remains the top pinned repo behind the org link,
so I don't think people will have any trouble finding it

and this change increases the chance that people will find
scala/bug, which is also pinned

### DIFF
--- a/_data/footer.yml
+++ b/_data/footer.yml
@@ -49,7 +49,7 @@
   class: social
   links:
     - title: GitHub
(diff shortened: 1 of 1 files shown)


faith (approved):
Seems reasonable, people should be able to locate the language's itself repo easily if needed and will be able to find more content made by scala team with less effort.

PR #1194, opened 2020-12-15
Use HTTPS for links

Marcono1234:
Currently there are multiple links in this project which use HTTP as protocol instead of HTTPS despite the target site supporting it. Even though the browser then upgrades the connection to HTTPS, it first sends an unencrypted request which could be read and whose response could be manipulated by an adversary in the same network.

Note that these creation of these changes was mostly automated, but only for links where a HTTPS connection did not appear to cause any issues.
Dead links (there are a few) and pages which returned a non-200 or 300 HTTP status code have not been changed; however I did not verify that every link still points to the correct page (in case a 200 status code was returned).

This also modifies blog posts and release notes. Please let me know if those should remain unchanged.
Any feedback is appreciated!

----

Could you please update the "About" section of this GitHub repository to use an HTTPS link as well?
GitHub repo "About" section screenshot


SethTisue:
wow, fantastic, thank you so much for tackling this!! much appreciated!

Marcono1234:
No problem. Could you please update the link in the "About" section of this GitHub repository as well (as pointed out above)?

SethTisue:
done!

PR #1198, opened 2020-12-26
Fix Atom namespace in feeds

### LINKED ISSUE #1197: Atom XML namespace must be http not https
Commit https://github.com/scala/scala-lang/commit/80b2437dc43bf34da8e226ec9e72335f314c05dd changed the Atom `xmlns` in `feed/blog.xml` and `feed/index.xml` to from `http://www.w3.org/2005/Atom` to `https`://www.w3.org/2005/Atom`.

This breaks my RSS reader because the namespace is no longer valid; Atom must be `http://www.w3.org/2005/Atom`, see e.g. <https://tools.ietf.org/html/rfc4287#section-1.2>

lunaryorn:
Must be HTTP not HTTPS.

Closes #1197

### DIFF
--- a/feed/blog.xml
+++ b/feed/blog.xml
@@ -1,7 +1,7 @@
 ---
 ---
 <?xml version="1.0" encoding="utf-8"?>
-<feed xmlns="https://www.w3.org/2005/Atom">
+<feed xmlns="http://www.w3.org/2005/Atom">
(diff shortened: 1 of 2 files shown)


SethTisue:
thank you!

lunaryorn:
Thanks 🙏

PR #1201, opened 2021-01-09
fix bad links to Dotty site

SethTisue:


SethTisue:
this is live now on https://www.scala-lang.org and removing Atom didn't break the layout

PR #1204, opened 2021-01-15
Blog about tuples in scala 3

vincenzobaz:
This PR follows https://github.com/scala/scala-lang/pull/1186

I have rewritten the first part of the post to better motivate why the new functions are useful.

The structure now looks like the following: 

 - tuples in Scala 3 are similar to HLists used in generic programming
 - why generic programming is useful
 - example of generic programming in scala 3: csv encoder
 - what other functions are there in the stdlib
 - how match types allow to implement the above functions with very precise types.

I hope that mentioning these topics and that summarizing the content at the beginning of the post also clarifies the target audience, which should be composed of intermediate and advanced Scala users.

I have also applied many corrections and suggestions thanks to @mlachkar @julienrf @MaximeKjaer


_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+even if I need a handle for the instances because it concentrates all the
+constraints in the type parameter list (and I do not need to come up with any
+name).  After this personal preference disclaimer, let's see the two cases:
+
+```scala
+object TupleEnc:
+  // Base case
+  given [T: Enc]: Enc[T *: EmptyTuple] with
+    def toCsv(oneElement: T *: EmptyTuple) =
+      summon[Enc[T]].toCsv(oneElement.head)
```
julienrf, line 113: I wonder if using an extension method for `toCsv` would provide a leaner syntax like
```suggestion
      oneElement.head.toCsv
```

julienrf, line 113: For reference, with the `ColumnEncoder`/`RowEncoder` suggestion, this given definition would have the following signature:
```suggestion
  given [T: ColumnEncoder]: RowEncoder[T *: EmptyTuple] with
    def encode(oneElement: T *: EmptyTuple) =
      List(summon[ColumnEncoder[T]].encode(oneElement.head))
```

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+
+  // Inductive case
+  given [H: Enc, T <: NonEmptyTuple: Enc]: Enc[H *: T] with
+    def toCsv(tuple: H *: T) =
+      summon[Enc[H]].toCsv(tuple.head) ++ summon[Enc[T]].toCsv(tuple.tail)
```
julienrf, line 118: And this one would look like this:

```suggestion
  given [H: ColumnEncoder, T <: NonEmptyTuple: RowEncoder]: RowEncoder[H *: T] with
    def encode(tuple: H *: T) =
      summon[ColumnEncoder[H]].encode(tuple.head) :: summon[RowEncoder[T]].encode(tuple.tail)
```

I personally think the symmetry between tuples and lists is better highlighted with this encoding.

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+```
+When recursion hits the last element of the tuple, we use its encoder,
+otherwise we invoke the encoder for the first element and for the tail of the
+tuple and combine the the two lists using the concatenation operator.
```
julienrf, line 122:
```suggestion
tuple and combine the two lists using the concatenation operator.
```

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+
+We can create an entrypoint function and test this implementation:
+```scala
+def tupleToCsv[X <: Tuple: Enc](tuple: X): List[String] =
+  summon[Enc[X]].toCsv(tuple)
+
+tupleToCsv(("Bob", 42, false)) // List("Bob", 42, false)
+```
+
```
julienrf, line 131: I suggest that we show explicitly the argument summoned by the compiler:

~~~ scala
tupleToCsv[String *: Int *: Boolean *: EmptyTuple](("Bob", 42, false))(using
  string,
  cons[String, Int *: Boolean *: EmptyTuple](using
    int,
    cons[Int, Boolean *: EmptyTuple](using
      boolean,
      singleton[Boolean]
    )
  )
)
~~~

Hmm, maybe not… Or maybe show it for simple cases? Also, it needs to use non-anonymous givens…

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+| `apply`    | `(1, 2, 3)(2)`                                              | `3`                                                  |
+| `splitAt`  | `(1, 2, 3, 4, 5).splitAt(2)`                                | `((1, 2), (3, 4, 5))`                                |
+| `zip`      | `(1, 2, 3).zip(('a', 'b'))`                                 | `((1 'a'), (2, 'b'))`                                |
+| `toList`   | `(1, 'a', 2).toList`                                        | `List(1, 'a', 2) : List[Int | Char]`                 |
+| `toArray`  | `(1, 'a', 2).toArray`                                       | `Array(1, '1', 2) : Array[AnyRef]`                   |
+| `toIArray` | `(1, 'a', 2).toIArray`                                      | `IArray(1, '1', 2) : IArray[AnyRef]`                 |
+| `map`      | `(1, 'a').map[[X] =>> Option[X]]([T] => (t: T) => Some(t))` | `(Some(1), Some('a')) : (Option[Int], Option[Char])` |
+
+
+# Under the hood: Scala 3 introduces match types
```
julienrf, line 177: I feel that this section could be a separate article. What you wrote above is already pretty dense :smile: 

At the same time, it makes me wonder… Do we have any guarantees that the two given instances you wrote for the transformation are total? Do we cover all the cases? Instead, could we implement it as a single recursive given instance and a match type (so that we would benefit from the exhaustivity checking of the compiler)?

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+```scala
+object BaseEnc:
+  given Enc[Int] with
+    def toCsv(x: Int) = List(x.toString)
+
+  given Enc[Boolean] with
+    def toCsv(x: Boolean) = List(if x then "true" else "false")
+
+  given Enc[String] with
+    def toCsv(x: String) = List(x)
```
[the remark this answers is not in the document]
vincenzobaz: What would be the most readable approach ?

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+us to define recursive operations on tuples.
+
+## Set up
+
+Let's define the `Enc[A]` type-class, which describes the capability of values
+of type `A` to be converted into `List[String]`:
+
+```scala
+trait Enc[A]:
+  def toCsv(a: A): List[String]
```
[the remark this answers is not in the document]
vincenzobaz: It is the simplest approach in my opinion. Introducing less names and types reduces the number of information the reader has to memorize to read the article.

I will try locally to decouple the two concepts and I will get back to you with my conclusions

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
vincenzobaz: I only marginally mention extension tuples in this article, I think that using only implicits makes the article simpler to read. Moreover if you remember we had some surprising behavior with extension resolution...
I prefer not to include extension in this critical part

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
vincenzobaz: I did not think about splitting the article.
I think both "halves" are short enough and are closely linked.

But it is an appropriate remark, what do the others think ?

vincenzobaz: I share your doubt about totality. The only case that does not seem to be covered to my eyes is `EmptyTuple`, for which we can add a simple encoder creating an empty list, but it is not that exciting.

What I can also do is having all the cases and maybe the type alias and split row/column encoders in a repo or gist that is linked in the article for completeness.
The idea would be to show the simplest and shortest approach in the article but provide a place to find a more complete solution. What do you think?

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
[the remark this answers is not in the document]
julienrf: Maybe add a comment?
```suggestion
    def toCsv(x: String) = List(x) // Ideally, we should also escape commas and double quotes
```

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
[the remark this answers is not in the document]
vincenzobaz: I have tested and it does look nice! So I upgraded the PR

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
sjrd: > I did not think about splitting the article.
> I think both "halves" are short enough and are closely linked.
> 
> But it is an appropriate remark, what do the others think ?

I don't have a strong opinion, but I would slightly lean towards splitting into two articles.

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+layout: blog-detail
+post-type: blog
+by: Vincenzo Bazzucchi, Scala Center
+title: Tuples bring generic programming to Scala 3
+---
+
+Tuples allow developers to create new types by associating existing types. In
+doing so, they are very similar to case classes but unlike them they retain
+only the structure of the types (e.g., which type is in which order) rather
+than giving each element a  name.
```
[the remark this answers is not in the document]
vincenzobaz: The previous iteration of the post used the collection analogy to introduce hlist, but feedback was mixed so I thought I would take another approach this time.

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
```diff
@@ -0,0 +1,326 @@
...
+
+This post focuses on how tuples in Scala 3 allow to address generic programming
+challenges, without external libraries or macros. It will also provide a short
+cheat sheet for new tuple operations as well as showing how a new language
+feature, dependent match types, allows the implementation of these operations.
+
+# Why generic programming ?
+
+When considering type-safety, HList offer the same guarantees as case classes,
+without having to declare class or field names.  This makes them more
```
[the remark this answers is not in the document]
vincenzobaz: Maybe it is not the correct terminology. I mean that similarly to case classes, tuples preserve the type of their elements

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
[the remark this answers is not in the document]
MaximeKjaer: Alternatively, the implementation can be quite short, and a well-named `def` can remove the need for a comment.

```scala
def toCsv(x: String) = List(escape(x))
def escape(x: String): String = x.replace(",", "\\,").replace("\"", "\\\"")
```

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
vincenzobaz: I think it is more confusing than helpful, especially with anonymous givens as you say

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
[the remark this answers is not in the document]
SethTisue: fwiw, I like the comparison to case classes. (I wouldn't be against _also_ including a comparison to collections.)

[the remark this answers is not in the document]
vincenzobaz: I have included a sentence so that the introduction contains analogies to both case classes and collections

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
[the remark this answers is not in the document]
vincenzobaz: Changed

_posts/2021-01-15-tuples-bring-generic-programming-to-scala-3.md
vincenzobaz: I thought a bit about it, but I prefer to keep the content together and not to split it in two articles.
The two subsections are linked and I believe that separating them will make each part look less "impressive".

sjrd:
I've tried taking a fresh look at this blog post, with the new structure, and having had time to forget the initial round.

I don't really know how to explain this. It's something very subjective and abstract that I'm trying to put into words, so I'm bound to make mistakes in how I express myself. Please bear with me.

Somehow I don't get the right "feeling" from this blog post. The feeling I get here is similar to what I get after reading random blog posts on the Internet: I see the thought process of the author as they go through their examples, and they clearly accompany me in that thought process to get me to the same understanding that they have. In a sense, the vibe I get is "here is what Scala 3 brings to me, and that I'd like to share with you". This is *good*, but somehow this is not the kind of vibe I expect from scala-lang.org. Rather, I expect a more "authoritative", "here is what we bring to you" vibe.

I am not sure what exactly makes me feel that way, but here are at least some concrete elements:

* Perhaps the clearest: separate the author from the reader. Use "we" as meaning the Scala core team (author included) and "you" as the reader. Avoid "we" as "the author and the reader together". For example, instead of "Let's consider", use "Consider". Drop things like "Let's see". In general, drop "Let's".
* Perhaps even better, in addition to using "we" as above, don't use "you" at all, if you can avoid it. Describe how things *are*. A blog post that does this well is https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html
* Bring useful stuff to the reader, instead of discovering useful stuff together with the reader. Start with the most easy-to-use, little-explanation-required stuff: start with the new operations that are available on tuples. For example, start with *using* `++` in a short example (not how to implement it). Show that the result type is the most precise one. Briefly show 2 or 3 more operations. Only move to generic programming, and finally how some operations are implemented.
* Instead of discussing how `concat` *can be* implemented, discuss how it *was* implemented. This brings the feeling that the author (author group, "we") is the one who improved the stdlib, and they are telling how. As opposed to how the author-and-reader could replicate what a third party did.

vincenzobaz:
I understand your comment and I think that our disagreement stems from the fact that you have a clear idea of the type of content that should be published on the blog. As you mention,

> This is not the kind of vibe I expect from scala-lang.org

I feel that this is reinforced by the points that you mention:
 - I am deliberately using "we" to designate me and the reader together, to  take the reader by the hand and show them my mental flow, to guide them through these new features
 - The explanation of `++` is motivated by the fact that I use tuple operations in the previous section, therefore after using them, I am interested in how they are implemented. The flow is problem=>feature to solve it=>how the feature could be implemented.  This approach is used throughout many courses and trainings which motivate useful abstractions, explain and then dissect
   them.  The previous version of the post contained a list of tuples operators with a quick example for each of them, But the comments on the PR were right, this style felt dull and unjustified ("hey look at all the fancy methods we now have")

> Instead of discussing how concat can be implemented, discuss how it was implemented

I do not agree with this idea. The actual implementation of tuple operators with arrays and general types is specific to tuples and to them only. It would not only be confusing but it would show a pattern that can hardily be reused by the reader. In contrast, I feel that the "nicer" and "naive" implementation of `concat` provides an example of usage of match types that readers could apply to other problems.

The problem of this post boils down to the type of content that we want on the blog.  I agree with you in that it shows a thinking process, trying to understand things step by step, similarly to an education tutorial. This was my intention because I like this approach and I believe it to be effective.

However if the goal of the blog is only to announce releases and features, to bring knowledge to the reader with "authority" as you say, then this post does not belong here and I will close the PR (maybe post it on a less "official" outlet). I could then start from scratch on a third version that satisfies this style or work on another topic.

vincenzobaz:
I understand your comment and I think that our disagreement stems from the fact that you have a clear idea of the type of content that should be published on the blog. As you mention,

> This is not the kind of vibe I expect from scala-lang.org

I feel that this is reinforced by the points that you mention:
 - I am deliberately using "we" to designate me and the reader together, to  take the reader by the hand and show them my mental flow, to guide them through these new features
 - The explanation of `++` is motivated by the fact that I use tuple operations in the previous section, therefore after using them, I am interested in how they are implemented. The flow is problem=>feature to solve it=>how the feature could be implemented.  This approach is used throughout many courses and trainings which motivate useful abstractions, explain and then dissect
   them.  The previous version of the post contained a list of tuples operators with a quick example for each of them, But the comments on the PR were right, this style felt dull and unjustified ("hey look at all the fancy methods we now have")

> Instead of discussing how concat can be implemented, discuss how it was implemented

I do not agree with this idea. The actual implementation of tuple operators with arrays and general types is specific to tuples and to them only. It would not only be confusing but it would show a pattern that can hardily be reused by the reader. In contrast, I feel that the "nicer" and "naive" implementation of `concat` provides an example of usage of match types that readers could apply to other problems.

The problem of this post boils down to the type of content that we want on the blog.  I agree with you in that it shows a thinking process, trying to understand things step by step, similarly to an education tutorial. This was my intention because I like this approach and I believe it to be effective.

However if the goal of the blog is only to announce releases and features, to bring knowledge to the reader with "authority" as you say, then this post does not belong here and I will close the PR (maybe post it on a less "official" outlet). I could then start from scratch on a third version that satisfies this style or work on another topic.

sjrd:
Your point of view makes sense. I don't have the final word on what should go in here and how. If you're confident with your style despite my point of view, it can go in as far as I'm concerned. ;)

SethTisue:
Maybe the introduction could do more to set readers’ expectations appropriately?

vincenzobaz:
Thank you @sjrd for your opinion. I personally like this style, so I am biased.
My reaction was driven by the fact the I have been working on this post about tuples since November and I would like to put it behind me either by getting accepted or rejected.
I will not insist for publishing if the style is not appropriate for the blog of if the content is not good enough, and focus on a new post if need be.

Thank you @SethTisue for your comment. I was not sure at all about how to address it but I managed to push a small change in the introduction, I hope that it solves the issue.

SethTisue:
@vincenzobaz user question here: https://users.scala-lang.org/t/error-compiling-tupletocsv-from-the-scala-blog/7275

SethTisue:
perhaps some of the material in the blog post  could make its way into a permanent doc page; the ticket on the nonexistence of that doc page is https://github.com/lampepfl/dotty/issues/9258

PR #1208, opened 2021-01-26
update Ruby to 2.5.8 (was 2.3.1) and nokogiri to 1.11.1

SethTisue:
as per CVE-2020-26247

also: https://github.com/scala/docs.scala-lang/pull/1909


SethTisue:
```
nokogiri-1.11.1-x86_64-linux requires ruby version < 3.1.dev, >= 2.5, which is
incompatible with the current version, ruby 2.3.1p112
```

I'm getting a feeling of deja vu, like this isn't the first time we've tried this :-/

sjrd (approved):

sjrd:
> ```
> nokogiri-1.11.1-x86_64-linux requires ruby version < 3.1.dev, >= 2.5, which is
> incompatible with the current version, ruby 2.3.1p112
> ```
> 
> I'm getting a feeling of deja vu, like this isn't the first time we've tried this :-/

I think we hadn't tried to upgrade Ruby, last time we encountered the issue.

PR #1209, opened 2021-02-02
Align download page with the getting started page

### LINKED ISSUE #1183: "Download" and "Getting Started" pages describe disjoint realities
The **Getting Started** link (which is good) looks like a caption for the **Download** button (which is bad). They target pages don't even reference each other. I thought the Download page was going away ... ?

image

Anyway there needs to be exactly one (1) supported way to set up Scala and it should be `cs setup`. This way we can document it and support it and not leave users fumbling around. We need to go all-in on this.

mlachkar:
Fixes #1183

image

image


_data/downloads.yml
```diff
@@ -1,29 +1,21 @@
...
+      <code class="hljs">$ curl -Lo cs https://git.io/coursier-cli-linux && chmod +x cs && ./cs setup </code> <br>
   - os: unix
-    text: "<p><code>java -version</code><span>(Make sure you have version 1.8.)</span></p>"
+    text: >-
+            <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a> then run:</p>
+            <p><code>$ ./cs setup</code></p>
   - os: macos
-    text: "<p><code>java -version</code><span>(Make sure you have version 1.8.)</span></p>"
+    text: >-
+      <div class="highlight">
```
sjrd, line 11: Why `highlight` here but not in the linux and unix sections?

_data/downloads.yml
```diff
@@ -1,29 +1,21 @@
...
+        <code class="hljs">$ brew install coursier/formulas/coursier && cs setup </code> <br>
+        </div>
+        <p>Alternatively, if you don't use Homebrew</p>
+        <div class="highlight">
+        <code class="hljs">$ curl -Lo cs https://git.io/coursier-cli-macos && chmod +x cs &&  (xattr -d com.apple.quarantine cs || true) && ./cs  setup </code> <br>
+      </div>
```
sjrd, line 17: The indentation is misleading.
```suggestion
        <code class="hljs">$ brew install coursier/formulas/coursier && cs setup </code> <br>
      </div>
      <p>Alternatively, if you don't use Homebrew</p>
      <div class="highlight">
        <code class="hljs">$ curl -Lo cs https://git.io/coursier-cli-macos && chmod +x cs &&  (xattr -d com.apple.quarantine cs || true) && ./cs  setup </code> <br>
      </div>
```

_layouts/downloadpage.html
```diff
@@ -16,48 +16,52 @@
...
-              <p><i>(If you don't have it installed, download Java from <a href="https://www.oracle.com/java/technologies/javase-jdk8-downloads.html">Oracle Java 8</a>, <a href="https://www.oracle.com/java/technologies/javase-jdk11-downloads.html">Oracle Java 11</a>, or <a href="https://adoptopenjdk.net/">AdoptOpenJDK 8/11</a>. Refer <a href="https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html">JDK Compatibility</a> for Scala/Java compatiblity detail.</i></p>
+              <h3>First, install Java, Scala, and all related command-line tools.</h3>
+              <p>Download the <code>cs</code> tool and execute the <code>setup</code> command</p>
+                <div id="download-step-one">
+                  <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a>, then run:</p>
+                  <p><code>$ ./cs setup</code></p>
+                </div>
+              <p>(See <a href="https://get-coursier.io/docs/cli-installation">the instructions for other operating systems.</a>)</p>
             </div>
+
```
sjrd, line 33: Spurious blank line?

_layouts/downloadpage.html
```diff
@@ -16,48 +16,52 @@
...
-              <p>To check, open the terminal and type:</p>
-              <p><code>java -version</code><span>(Make sure you have version 1.8 or 11.)</span></p>
-              <p><i>(If you don't have it installed, download Java from <a href="https://www.oracle.com/java/technologies/javase-jdk8-downloads.html">Oracle Java 8</a>, <a href="https://www.oracle.com/java/technologies/javase-jdk11-downloads.html">Oracle Java 11</a>, or <a href="https://adoptopenjdk.net/">AdoptOpenJDK 8/11</a>. Refer <a href="https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html">JDK Compatibility</a> for Scala/Java compatiblity detail.</i></p>
+              <h3>First, install Java, Scala, and all related command-line tools.</h3>
+              <p>Download the <code>cs</code> tool and execute the <code>setup</code> command</p>
+                <div id="download-step-one">
+                  <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a>, then run:</p>
+                  <p><code>$ ./cs setup</code></p>
+                </div>
```
sjrd, line 30:
```suggestion
              <p>Download the <code>cs</code> tool and execute the <code>setup</code> command</p>
              <div id="download-step-one">
                <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a>, then run:</p>
                <p><code>$ ./cs setup</code></p>
              </div>
```

index.md
```diff
@@ -9,10 +9,10 @@ headerButtonUrl: "/what-is-scala/"
...
 # Links of the Download / API Docs sections
 gettingStarted:
-  mainTitle: "Download"
-  mainUrl: "/download"
-  subtitle: "Getting Started"
-  subtitleLink: "https://docs.scala-lang.org/getting-started.html"
+  mainTitle: "Getting Started"
+  mainUrl: "https://docs.scala-lang.org/getting-started.html"
+  subtitle: "Download"
+  subtitleLink: "/download"
```
sjrd, line 15: Hum, the new Download page is more beautiful/more engaging than the Getting Started page. IMO we should leave it as the bigger button, so keep this unchanged.

sjrd (changes requested):
Very nice! Thank you for doing this.

index.md
mlachkar: It's in fact more beautiful but not "beginners' oriented". It contains release notes, Scala archives.
Putting the getting started first, shows that we give more importance to the learning process. 
Maybe we can have the same style of CSS in the getting started, and remove duplicated information from the download page?

_data/downloads.yml
mlachkar: It's a mistake

_data/downloads.yml
mlachkar: ok. thanks

_layouts/downloadpage.html
mlachkar: ok

_layouts/downloadpage.html
mlachkar: ok

index.md
sjrd: > It's in fact more beautiful but not "beginners' oriented". It contains release notes, Scala archives.
Putting the getting started first, shows that we give more importance to the learning process.

I guess that makes sense. Let's keep it as is, then.

sjrd (approved):

julienrf:
@SethTisue do you want to have a look?

vincenzobaz:
Can we have jekyll or another tool doing the HTML formatting?

_data/downloads.yml
```diff
@@ -1,29 +1,24 @@
...
+    text: >-
+      <p>Follow <a href="https://get-coursier.io/docs/cli-overview.html#install-native-launcher" target="_blank">the instructions to install the <code>cs</code> launcher</a> then run:</p>
+      <div class="highlight">
+        <p><code class="hljs">$ ./cs setup</code></p>
+      </div>
   - os: macos
-    text: "<p><code>java -version</code><span>(Make sure you have version 1.8.)</span></p>"
+    text: >-
+      <div class="highlight">
+        <code class="hljs">$ brew install coursier/formulas/coursier && cs setup </code> <br>
```
eed3si9n, line 16: The current download page links to the official sbt page to setup sbt. I am concerned about `cs setup` because it would install a fake application called `sbt` that is actually not the Bash script maintained by sbt team. This end up creating needless confusion because inevitably some behavior will be different and the user would complain to the sbt project, not Couriser. See for example https://github.com/sbt/sbt/issues/6299.

julienrf: Would it be possible to unify them? @alexarchambault

SethTisue: seconded that the coursier-based sbt launcher is not yet ready for this level of exposure

SethTisue:
> @SethTisue do you want to have a look?

I share Eugene's concern that the coursier-based sbt launcher isn't ready.

Other than that, I don't have time this week to give this a really good review, but y'all should go ahead anyway as you see fit.

julienrf:
Thank you Eugene and Seth for your review.

> I share Eugene's concern that the coursier-based sbt launcher isn't ready.

This is what we recommend on the getting started page on docs.scala-lang.org. The goal of this PR is to deliver a consistent message in both places.

So, I see two paths forward:

1. We find a way to make `cs setup` install the official sbt launcher
2. Or, is there a concrete list of issues that we could address to make the coursier-based launcher ready?

SethTisue:
> is there a concrete list of issues that we could address to make the coursier-based launcher ready?

I don't know, but the biggest one I'm personally aware of is that it ignores `.jvmopts`.

eed3si9n:
As a maintainer of the `sbt` bash script, there's only one official script that sbt project adds features, accept contributions for fix bugs, released to SDKMAN, Bintray deb repo etc, and the rest are fake ones. Similar to the sbt 1.3.x swapping the default dependency resolver to Coursier, if we wanted to use Coursier for the launcher (I'm in support of it), we should unify the launcher, and not fork with the same name.

I don't keep track of the missing features and subtle bugs like difference in environment variables, but for example, it should support [repository override](https://www.scala-sbt.org/1.x/docs/Proxy-Repositories.html) for corporate users behind the proxy. If not backward compatible out-of-box, at least provide a way to fall back to Ivy similar to `useCoursier := false`.

alexarchambault:
About installing the official sbt launcher with `cs setup`, I guess it should be possible to install it just like native applications are installed today, like `sbtn` already. That is, by picking a file from a package on GitHub releases. 

This might need a (very) little bit of tweaking, so that we don't add an indirection when we install those (currently, native apps are written in a file like `.sbtn.aux`, and `sbtn` is a script that starts the `.sbtn.aux`, we should just get rid of that). But that's mostly cosmetic, it might even already work today…

julienrf:
@alexarchambault Would you be interested in making this change (`cs setup` installs the official sbt launcher) happen?

alexarchambault:
> Would you be interested in making this change (cs setup installs the official sbt launcher) happen?

I just tried, and it turns out it's a little bit more complicated. The scripts in the sbt archive seems not to work in a standalone fashion. The `sbt` and `sbt.bat` scripts need `sbt-launch.jar` and `java9-rt-export.jar` alongside them. Which breaks what I had in mind and could have mostly worked right now, that is just picking `sbt` or `sbt.bat` from the archive, and put that in the `PATH`.

(I'm not sure what was the point of keeping `java9-rt-export.jar` separate. Its 3 classes could just have been folded in `sbt-launch.jar`… And `sbt-launch.jar` could actually be appended to `sbt` and `sbt.bat`, so that these scripts don't need to look for external files.)

To work around that, `cs` could maybe extract the full archive, instead of picking just one file, and have the script it generates itself (the one actually put in the `PATH`) call the right file from the extracted archive. But that needs a bit of work…

smarter:
I'm a big fan of `cs setup`, but as I mentioned in the corresponding issue, getting it to work reliably on all systems is not that trivial, see https://github.com/scala/scala-lang/issues/1183#issuecomment-766926804 and in particular https://gitlab.epfl.ch/lamp/cs206/-/blob/master/labs/tools-setup.md#step-2-installing-the-java-development-kit-jdk-and-sbt-via-coursier which lists all the steps necessary on all systems that I've encountered.

alexarchambault:
@smarter It seems there are only problems on Windows, all related to shelling out to powershell, right? Maybe replacing the few powershell calls in directories-jvm and coursier by JNI calls could help here…

smarter:
In the issues I've linked to in https://github.com/scala/scala-lang/issues/1183#issuecomment-766926804, there's one Windows/powershell issue, but I don't think the rest are powershell-specific (e.g., there's some issues related to using graal native), and some are not Windows-specific (~/.bash_profile having the wrong access rights / not being read on terminal startup on all systems).

b-studios:
What is the current state of this PR @mlachkar? I started added scala 3 instructions in #1246, maybe the efforts can be joined?

julienrf:
AFAICT we are a bit blocked by that https://github.com/scala/scala-lang/pull/1209#issuecomment-775331757

PR #1217, opened 2021-03-04
fix 404s

SethTisue:


_events/2015-10-08-applyscala.md
```diff
@@ -6,5 +6,5 @@ location: Amsterdam
 description: "Scala experience from the trenches"
 start: 8 October 2015
 end: 9 October 2015
-link-out: http://applyconf.com/scala/
+link-out: https://www.applyconf.com
```
Philippus, line 9: @raboof do you remember if this event happened in the end? I think the link is not valid anymore.

raboof: The event that we planned never happened - and interestingly that website now seems to describe a completely different event put on by completely different people ;)

Philippus: Thanks, follow up: https://github.com/scala/scala-lang/pull/1218.

PR #1226, opened 2021-03-19
Fixed typo in Views section of Collections Rework

tsmanner:
There's an extra ' a ' in the body of the description for Views.  I noticed it while reading the docs looking for more information about the 2.12->2.13 collections changes, and figured I might as well PR it over.  This is not related to any open issues, and I didn't think it was worth opening a new issue for correcting a typo that doesn't affect the meaning of the passage.

That said, I totally understand if this is closed to keep the archival view of the post unchanged.  Cheers


SethTisue:
thank you! fixes like these, no matter how small, are definitely appreciated

PR #1227, opened 2021-03-23
delete orphan file

martijnhoekstra:
cross-reference: https://github.com/scala/scala-dev/issues/759#issuecomment-803617396

https://www.scala-lang.org/documentation/reference.html is where it ends up. Does anything link there? Should anything link there? grep doesn't find anything relevant for `reference\)` or `reference\]` or (after building the site) `reference.html`. Deleting it doesn't seem to break anything.


SethTisue:
thanks :-)

PR #1239, opened 2021-05-10
Fix Docker setup

julienrf:
The Docker setup was not working for me, most probably because it was not using a specific version of the jekyll image, and maybe the latest images have broken something?

This PR introduces our own Dockerfile, which uses the same version of ruby as on the CI (2.5), and installs Jekyll from there. When this Docker image is built, the website dependencies are fetched and cached (ie, we run `bundle install` during the Docker image build time).


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ruby:2.5
+
+RUN gem install bundler jekyll
+
+RUN mkdir /srv/jekyll
+
+WORKDIR /srv/jekyll
```
vincenzobaz, line 7: You can just use `WORKDIR path`, if `path` does not exist, then it will be created.
Why do we need `/srv/jekyll`? can't we just use `/jekyll` or `/site`?

vincenzobaz (approved):

Dockerfile
julienrf: We don’t need `/srv/jekyll` specifically, it’s the path used by the official `jekyll` image, so I’ve used the same.

julienrf:
@vincenzobaz I’ve simplified the Dockerfile according to your advice, and I’ve also updated the README and removed unused (I think) files.

adpi2 (commented):
Thanks @julienrf it's been easier for me to make the site work with this docker setup.

It seems the docker-compose config in https://github.com/scala/docs.scala-lang is different. So which one is best?

julienrf:
> It seems the docker-compose config in https://github.com/scala/docs.scala-lang is different. So which one is best?

Good point. Vincenzo also raised this point. We should probably try to unify them. On my side, my experience with docs.scala-lang.org was that I was unable to get incremental rendering work properly. What is your experience?

adpi2 (approved):
I have same experience than you with docs.scala-lang: incremental building does not seem to work at all. It would be way better with the setup of this PR.

PR #1244, opened 2021-05-17
Scala 2.13.6

lrytz:


dwijnand (approved):
Can this be merged?

lrytz:
we usually wait a bit before announcing to give some time to compiler plugins, scala.ja

SethTisue:
Seb said on Discourse that Scala.js is published.

Scalameta released an hour ago: https://github.com/scalameta/scalameta/releases/tag/v4.4.18

kind-projector hasn't published (https://github.com/typelevel/kind-projector/issues/191) because Lars is ill, but I don't think we need to wait.

I will suggest at team meeting, an hour from now, that we go ahead and announce.

PR #1246, opened 2021-05-18
Scala3/download

b-studios:
Draft PR to add a separate Scala 3 download page.

The overall structure is:

- `/download` now is a landing page that provides a choice between Scala 2 and Scala 3.
image

- `/download/scala2.html` contains the old instructions. Only difference is the background color and the color of the tag in the top right: 
image

- `/download/scala3.html` contains the new instructions. 
image

Release pages use either the Scala 2 install instruction page or Scala 3 instruction page, depending on the first number in the release:

image
image

- The homepage now mentions both Scala 2 and Scala 3. The buttons point to the respective download page. 
image


_downloads/2021-05-04-3.0.0.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Scala 3.0.0
+start: 14 May 2021
```
julienrf, line 3: Typo in the file name, it should be `2021-05-14-3.0.0.md`

download/index.md
```diff
@@ -1,6 +1,7 @@
 ---
-layout: downloadpage
+layout: download
 title: Download
+<<<<<<< HEAD
```
julienrf, line 4: You need to fix this conflict.

_includes/downloads-scala3.html
```diff
@@ -0,0 +1,136 @@
...
+      <span>Download Coursier</span>
+    </a>
+    <div class="text-variant">
+      <p>Best if you want to get <strong>started quickly</strong>.</p>
+      <p><i>Coursier will install all the necessary dependencies for you, so after
+         installing Coursier you are ready to go.</i></p>
+    </div>
+  </div>
+  <div class="download-right">
+    <a href="https://www.scala-sbt.org/download.html"" class="btn-download">
```
julienrf, line 26:
```suggestion
    <a href="https://www.scala-sbt.org/download.html" class="btn-download">
```

_includes/downloads-scala3.html
```diff
@@ -0,0 +1,136 @@
...
+      <p>Open a terminal and run the following commands:</p>
+      <p><code>cs install scala3-compiler</code><span>to install the Scala 3 compiler</span></p>
+      <p><code>cs install scala3-repl</code><span>to install the Scala 3 interactive read-eval-print loop</span></p>
+    </div>
+  </div>
+  <div class="step">
+    <div class="number-step">3</div>
+    <div class="text-step">
+      <h3>Use Scala 3</h3>
+      <p>To use the Scala 3 compiler to compile a file <code>test.scala</code> simply run<br><code>scala3-compiler test.scala</code> in your terminal.</p>
```
julienrf, line 71: I don’t think this is useful advice, this is very far from real-world usage where projects have multiple files and third-party library dependencies.
I don’t think it’s a good idea to teach this approach to the users, since they will have to de-learn it after a few hours and re-learn how to work on proper Scala projects.
I would start with a simple project seed, whose build definition is minimalist and not intimidating, but that scales to a “regular” project (ie, a project with more than one file, and possible some library dependencies).

julienrf (changes requested):
I need to render it locally to better feel the end result, but I’ve left a few comments while reading the diff.

_includes/downloads-scala3.html
b-studios: Sorry -- I fixed it, but forgot to save before adding... My bad.

download/index.md
b-studios: Sorry -- I fixed it, but forgot to save before adding... My bad.

_includes/downloads-scala3.html
b-studios: I just wanted to somehow mention that these commands now exist after installing them above. Otherwise it is near difficult to newcomers to guess the command names. What would you suggest?

abgruszecki (commented):
Looks good overall!

_includes/downloads-scala3.html
b-studios: > otherwise it's not obvious how to actually get it to run.

Agreed!

_includes/downloads-scala2.html
```diff
@@ -0,0 +1,87 @@
+<h2>The most popular way to get Scala is either using Scala through sbt, the Scala build tool, or to use Scala through an IDE.</h2>
+<div class="install-steps">
+  <div class="step arrow">
+    <img src="/resources/img/download/arrow-asset.png" alt="">
+    <div class="number-step">1</div>
+    <div class="text-step">
+      <h3>First, make sure you have the Java 8 JDK (or Java 11 JDK) installed.</h3>
+      <p>To check, open the terminal and type:</p>
+      <p><code>java -version</code><span>(Make sure you have version 1.8 or 11.)</span></p>
+      <p><i>(If you don't have it installed, download Java from <a href="https://www.oracle.com/java/technologies/javase-jdk8-downloads.html">Oracle Java 8</a>, <a href="https://www.oracle.com/java/technologies/javase-jdk11-downloads.html">Oracle Java 11</a>, or <a href="https://adoptopenjdk.net/">AdoptOpenJDK 8/11</a>. Refer <a href="https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html">JDK Compatibility</a> for Scala/Java compatiblity detail.</i></p>
```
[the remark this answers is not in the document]
b-studios: To simplify reviews, I wanted to leave the Scala 2 page untouched, yet.

b-studios:
> I need to render it locally to better feel the end result, but I’ve left a few comments while reading the diff.

I added a few screenshots to the PR message.

_includes/downloads-scala3.html
```diff
@@ -0,0 +1,136 @@
...
+    </div>
+  </div> -->
+  <div class="step arrow">
+    <img src="/resources/img/download/arrow-asset.png" alt="">
+    <div class="number-step">2</div>
+    <div class="text-step">
+      <h3>Then, install Scala 3</h3>
+      <p>Open a terminal and run the following commands:</p>
+      <p><code>cs install scala3-compiler</code><span>to install the Scala 3 compiler</span></p>
+      <p><code>cs install scala3-repl</code><span>to install the Scala 3 interactive read-eval-print loop</span></p>
```
smarter, line 64: We could skip all of this if scala 3 became the default for scala/scalac in coursier as discussed in https://github.com/coursier/coursier/issues/1992 /cc @alexarchambault

_includes/downloads-scala3.html
b-studios: I slightly rephrased it to: 
image

Is there a good resource to point people to after this paragraph? The getting started page is not more detailed on how to use the REPL.

_includes/downloads-scala3.html
b-studios: Good point @smarter!

b-studios:
It also remains to decide, what to do with the frontpage: 
image

- Should `Download` point to the download landing page? (I guess yes)
- What resources and links should be mentioned below Download and ApiDocs?
- What should we do about the version number in the middle? Delete, have both?

b-studios:
As far as I am concerned, this PR is ready for review / merge. The point raised by @smarter is valid and we need to adapt the explanations once the coursier CLI tools are renamed. However, we could also do that in a separate PR.

_includes/downloads-scala3.html
abgruszecki: I think it might be best simply to say something like "If experimenting with the above leaves you interested for more, head on to the next section in which we explain what you need to work with Scala on larger projects".

abgruszecki:
I think the screenshots look great, both the download pages and the main page!

_includes/downloads-scala3.html
alexarchambault: I agree Scala 3 should become the default for these apps. I need to make sure the `_3` suffix is handled correctly in coursier, to enable that.

_layouts/frontpage.html
```diff
@@ -75,7 +75,6 @@ <h1>{{page.headerTitle}}</h1>
               <img src="/resources/img/frontpage/scala-spiral.png" alt="">
               <div class="scala-version">
                 <span>Scala</span>
-                <span>{{page.currentScalaVersion}}</span>
```
julienrf, line 78: Why removing the current scala version?

download/scala3.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: downloadpage
+title: Download Scala 3
+
+release_version: 3.0.0
+release_date: "May 14, 2021"
+---
```
julienrf, line 7: Don’t you need to add the `license`?

julienrf (approved):
I feel like we will have to rework the instructions anyway because there is a lot of redundancy between the download and the getting started pages. But at the same time, I see the urgency in showing the instructions to try Scala 3.0.0 now.

_layouts/frontpage.html
b-studios: Which one should we show? Scala 2 or Scala 3, maybe both? I just deleted it since the buttons now already show the versions.

download/scala3.md
b-studios: You should, yes. But then a lot of the other release notes also didn't so I followed the lead. Probably a bad example.

b-studios: Wait, I added it to the release notes in a later commit. I should also add it here, I guess.

b-studios:
> I feel like we will have to rework the instructions anyway 

Yes, I agree. This PR is just a step into that direction by structurally supporting Scala 2 and Scala 3 at all.

_layouts/frontpage.html
julienrf: OK, I didn’t see that. If it’s easy to show both Scala 2 and Scala 3 version I would go for that, otherwise I think we should show the Scala 3 version number.

b-studios: Just to be clear: you think showing them in the buttons only is not enough? 

image

If no then I can also add the versions here tomorrow.

_includes/downloads-scala3.html
b-studios: Great! Still, I would like to change this in a separate PR. We probably also need to change all instructions for Scala 2 to deal with this changed behavior.

b-studios:
Thanks for approving @julienrf! I wait until tomorrow with the merging so we can clarify the remaining open points, first.

_layouts/frontpage.html
julienrf: No, OK it’s enough to have the version on the buttons.

_includes/downloads-scala3.html
b-studios: I was actually thinking about resources on the REPL. Of course sbt is an alternative for larger projects (and there is a lot of docs about it) but IIUC we are missing a basic getting started guide on how to use the REPL and experiment from there on.

PR #1251, opened 2021-06-16
community page: replace Freenode with Libera

SethTisue:


sjrd (approved):

sjrd:
I assume the underlying migration was already done?

SethTisue:
the old Freenode has actually been shut down; there is a new Freenode but it seems to be roundly condemned by everybody I follow on Twitter. and notably, #haskell bailed and went to Libera. (many of the people who find #scala come from #haskell, which still has a strong community.)

and yes, we are all set up on Libera now. (or almost, I'll contact you about finding at least one more moderator besides myself)

PR #1254, opened 2021-07-07
Change the order of the getting-started with the download page for Scala 2 and Scala 3

mlachkar:
While waiting for #1209 where we will align both the getting-started and
download page, we need to provide a first entry point to users wishing to get started
with Scala that give helpful instructions. The download page is still easy to access
through the navigation bar

### DIFF
--- a/index.md
+++ b/index.md
@@ -10,7 +10,7 @@ headerButtonUrl: "/what-is-scala/"
 # Links of the Download / API Docs sections
 gettingStarted:
   mainTitle: "Scala 3.0.0"
-  mainUrl: "/download/scala3.html"
+  mainUrl: "https://docs.scala-lang.org/scala3/getting-started.html"
   subtitle: "Documentation"
   subtitleLink: "https://docs.scala-lang.org/scala3"
   links:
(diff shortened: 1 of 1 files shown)


sjrd:
Could you post a screenshot of what's changed, exactly?

mlachkar:
Screenshot 2021-07-07 at 16 01 09

sjrd:
So the new links go to
https://docs.scala-lang.org/scala3/getting-started.html
and
https://docs.scala-lang.org/getting-started/index.html
IIUC.

The good thing is that both pages agree on how to install Scala (though the wording is different). The bad thing is that:
* For Scala 3, the REPL doesn't work at the moment
* They use `cs setup`, and that still has the problems raised in #1209 about, notably, the sbt launcher.

I assume that the concerns about the second one will still apply to this PR :-(
/cc @eed3si9n, @SethTisue

mlachkar:
I agree on what you reported, but I'm linking pages that are already published on the website, which are the getting-started pages, so people have already access to them.

With this PR, I just give more importance to those pages that are less confusing than the download page as reported here #1183. I believe that the getting started is the page that people look for, when trying to discover a language, and right now the download page for Scala 2 is a mix of a getting started/ release notes/ how to download binaries. 

Of course working on `cs setup` is definitely the right way to go, but meanwhile, the welcome page for Scala 2 links pages that are outdated, not aligned with the getting started. The change for Scala 3 is just motivated by having something symmetric.

PR #1255, opened 2021-07-08
Add effective scala to courses

vincenzobaz:
Effective Scala currently does not appear on scala-lang.org.
This PR adds it along to the other courses with the same description that is used on Coursera (the same is done for other courses).

See the screenshots
before
after


julienrf (approved):
Thank you Vincenzo!

PR #1270, opened 2021-08-18
Add SDKMAN to the Scala 3 download page

marc0der:
Hi folks,

This PR adds a small section to the Scala 3 download page which is similar to that found on the Scala 2 page. Since Scala 3.0.1 is now the default for Scala on SDKMAN, I've also updated the Scala 2 download page appropriately. Please shout if there is anything you would like me to change.

Thanks!

### DIFF
--- a/_includes/downloads-scala2.html
+++ b/_includes/downloads-scala2.html
@@ -65,7 +65,7 @@
           <a id="download-binaries">Download the Scala binaries for <span id="users-os"></span></a>
           <br/><span class="install"><a href="{{ site.baseurl }}/download/install.html">Need help running the binaries?</a></span>
         </li>
-        <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install specific versions of Scala</a> on any platform with <code>sdk install scala 2.13.6</code></li>
+        <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install specific versions of Scala</a> on any platform with <code>sdk install scala {{ site.scalaversion }}</code></li>
         <li>On macOS you can also use <a href="https://brew.sh/">Homebrew</a> and existing <a href="https://formulae.brew.sh/formula/scala">Scala Formulae</a><br/><code>brew update</code><br/><code>brew install scala</code></li>
         <li>With <a href="https://www.macports.org/">MacPorts</a>, you can get Scala using <code>sudo port install scala2.x</code>command.<br/> For example to install Scala 2.12 simply use<code>sudo port install scala2.12</code></li>
         <li>Use <a href="https://scastie.scala-lang.org">Scastie</a> to run single-file Scala programs in your browser using multiple Scala compilers; the production Scala 2.x compilers, Scala.js, Dotty, and Typelevel Scala. Save and share executable Scala code snippets.</li>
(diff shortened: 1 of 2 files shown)


_includes/downloads-scala2.html
```diff
@@ -65,7 +65,7 @@ <h2>Other ways to install Scala</h2>
           <a id="download-binaries">Download the Scala binaries for <span id="users-os"></span></a>
           <br/><span class="install"><a href="{{ site.baseurl }}/download/install.html">Need help running the binaries?</a></span>
         </li>
-        <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install Scala</a> with <code>sdk install scala</code></li>
+        <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install specific versions of Scala</a> on any platform with <code>sdk install scala 2.13.6</code></li>
```
julienrf, line 68: I see a problem with this line in that we will have to maintain it for every new version of Scala. You can write this instead:

```suggestion
        <li>Using <a href="https://sdkman.io/">SDKMAN!</a>, you can easily <a href="https://sdkman.io/sdks#scala">install specific versions of Scala</a> on any platform with <code>sdk install scala {{ site.scalaversion }}</code></li>
```

But we might have another problem, which is that SDKMAN might not immediately support the next release? Maybe it’s taking a couple of weeks? I don’t know.

julienrf (commented):
Thank you @madorb for your contribution!

_includes/downloads-scala2.html
marc0der: Ha, I have that covered! :smile: We have a [vendor release API](https://sdkman.io/vendors) that you can use to publish your version the moment they land. This is the approach taken by most SDK providers like SBT, Kotlin, Gradle and many more. Let me know if this would be of interest, and I can open another issue to help set that up for you.

The other option is to do it via a [DB migration](https://github.com/sdkman/sdkman-db-migrations) PR on SDKMAN. Since I have a vested interest, the Scala versions usually land in SDKMAN no more than a day or two after the official release.

julienrf: Sure, it would be great to automatically let SDKMAN know that we published a new release. But this is not blocking this PR!

marc0der: So the problem here is that the variable holds the latest Scala version (I presume 3.0.1), but here we need the latest 2.13.x release as this is for the Scala 2 download page. Do you perhaps have another variable that tracks this? Alternatively, we could just make it `2.13.x`.

marc0der: Just to clarify and for context, the latest version of Scala is installed if you omit the version:
```
sdk install scala
```
For any other version you need to be explicit:
```
sdk install scala 2.13.6
```

Hope it makes more sense now!

julienrf: The variable `site.scalaversion` is currently assigned to 2.13.6, so this will do the job. We should introduce a `scala3version` variable, at some point, most probably, to manage the Scala 3 content. The scala-lang.org website will show information about both versions of Scala for some time.

marc0der:
@julienrf thanks for merging! On what project should I raise the automated publishing to SDKMAN?

julienrf:
Thank you @marc0der! Please use the issue tracker of that repo: https://github.com/scala/scala-dev/

PR #1272, opened 2021-08-23
Update scala 2 documentation page

aaparent:
Change the url for scala 2 documentation landing page (actual does 404)


julienrf (approved):
Good catch, thank you @aaparent!

PR #1273, opened 2021-08-24
Remove Dotty Gitter channel link

ennru:
as discussions should go to https://gitter.im/scala/contributors﻿


SethTisue (approved):
I've added a commit that removes it from another place

PR #1276, opened 2021-09-02
Migrate to Github Actions

### LINKED ISSUE #1275: switch from Travis-CI to GitHub Actions
we're looking to reduce our Travis-CI usage in scala/* (https://github.com/scala/scala-dev/issues/785)

is there a volunteer who would like to attempt the switch?

presumably after it was done in this repo, pretty much the same config could also be used in docs.scala-lang, so I won't bother opening a separate ticket there for now

exoego:
Closes #1275 

See test run on https://github.com/exoego/scala-lang/pull/1

## Changes

1. Email notifications are removed, since 
    * GitHub have built-in email notification on job failure, outside of GH Actions
    * It requires extra work to configure email notification inside GH Actions


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,31 @@
+name: scala-lang docs
+
+on:
+  pull_request:
+  schedule:
+    - cron: '00 0 * * *'
```
exoego, line 6: @SethTisue 
This schedule should be updated per your preference.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,31 @@
...
+  build-and-validate-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.5.8'
+          bundler-cache: true
+      - name: Build documents
+        run: |
```
julienrf, line 18: I see in the Travis-CI script that we also run `bundle update --bundler`. Is this step not necessary here as well?

https://github.com/scala/scala-lang/blob/a2252985ad8708b60cbda666aa58c087370f3b83/.travis.yml#L8

.github/workflows/ci.yaml
julienrf: I am not sure why we need this? (there seems to be something similar in the Travis CI settings, scheduling a build every day unless there was already an automatic build triggered by a PR)
The only thing I can think about is that this would allow us to detect broken external links sooner?

.github/workflows/ci.yaml
exoego: My understanding for that line is to update bundles (dependencies) while CI.
I do not know a motivation behind that.
Usually, CI uses dependencies as defined.

.github/workflows/ci.yaml
exoego: I think schedule is needed because validation logic is different for pull requests and scheduled run

SethTisue: > this would allow us to detect broken external links sooner

I believe that's the motivation, yes.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,31 @@
...
+          bundle exec jekyll build 2> error.log
+          cat >&2 error.log
+          ( ! grep -qie Error -e Warn error.log )
+          find ./_site \( -type d \( -name events -o -name training \) -prune -false \) -o \( -type f ! -name error.log -empty \)
+          find ./_site \( -type d \( -name events -o -name training \) -prune -false \) -o \( -type f ! -name error.log -empty \) | ( ! read )
```
vincenzobaz, line 23: Can we simplify this?
We could rely more on Github Actions logs rather than using custom text files that are not exported anywhere.

.github/workflows/ci.yaml
exoego: Added that step.
https://github.com/scala/scala-lang/pull/1276/commits/4e7ce9b406a4c57b8b7ecbe7f29d96414e01af92
This step may be replaced with other (e.g. dependabot), but I am not sure.

.github/workflows/ci.yaml
exoego: It could be, but I would like to keep this PR as a straight rewrite with minimum changes

julienrf (approved):

SethTisue:
working on test PR (#1278) :+1:

thank you @exoego!!

PR #1277, opened 2021-09-02
Scala 3.0.2

Kordyjan:


blog/_posts/2021-09-02-scala-3.0.2-released.md
```diff
@@ -0,0 +1,194 @@
+---
+layout: blog-detail
+post-type: blog
+by: Paweł Marks, VirtusLab
+title: Scala 3.0.2 released!
+---
+
+Greetings from the Scala 3 team! We are glad to announce that Scala 3.0.2 is now officially out. As no critical bugs have been found in the previously released Scala 3.0.2-RC2, it has been promoted to 3.0.2 and is the current stable scala version.
```
julienrf, line 8:
```suggestion
Greetings from the Scala 3 team! We are glad to announce that Scala 3.0.2 is now officially out. As no critical bugs have been found in the previously released Scala 3.0.2-RC2, it has been promoted to 3.0.2 and is the current stable Scala version.
```

blog/_posts/2021-09-02-scala-3.0.2-released.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+### Towards better null safety in the type system
+
+The compiler option `-Yexplicit-nulls` modifies Scala's standard type hierarchy to allow easier tracing of nullable values by performing strict checks directly on the level of the type system rather than just relying on conventions (e.g. this prevents you from writing code like `val foo: Option[String] = Some(null)`, which would be otherwise valid Scala although very likely to cause a `NullPointerException` at some further point).
+
+After the recently introduced changes with this option enabled the `Null` type becomes a subtype of `Matchable` instead of inheriting directly from `Any`, making the code below compile (this used to compile before only without strict nullability checking).
+
+```scala
+def foo[T <: Matchable](t: T) = t match { case null => () }
```
julienrf, line 56: Maybe a simpler example would be the following?

```suggestion
def foo(x: Matchable) = x match { case null => () }
```

julienrf (approved):

PR #1292, opened 2021-11-03
Move charset information before any content

s5bug:
This fixes the rendering of non-ASCII characters in the OpenGraph embeds. For example, 🎉🎉🎉 is rendering as ðŸŽ‰ðŸŽ‰ðŸŽ‰.

image

### DIFF
--- a/_includes/headertop.html
+++ b/_includes/headertop.html
@@ -1,6 +1,9 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="UTF-8"/>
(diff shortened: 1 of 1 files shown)


SethTisue:
I wasn't seeing this issue in my browser (Safari), but it seems like a safe change.

s5bug:
@SethTisue This doesn't show up in browsers as they correctly interpret the charset coming after the OGP data, however Discord's crawler does not.

PR #1293, opened 2021-11-04
Remove slashes from the end of charset tags

s5bug:
For some reason, some scrapers (most notably Discord's OGP scraper) don't respect charset tags that do contain closing slashes.

Huge thanks to @adryd325 for taking the time to poke at Discord's scraper to figure this out.

### DIFF
--- a/_includes/headertop.html
+++ b/_includes/headertop.html
@@ -1,8 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="UTF-8"/>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


SethTisue:
There is a bunch of stuff about this issue out there on the web, e.g. https://superuser.com/questions/1048013/why-some-metatags-have-in-the-end-and-some-dont , not really sure what the most authoritative source of advice is on this, but afaict from what I read the change is safe, given that our site is HTML and not XHTML.

SethTisue:
Assuming this fixes it, a corresponding PR over in the other repo (https://github.com/scala/docs.scala-lang/) would be welcome.

s5bug:
I'll wait and see, and if it does fix it I'll do so 😄

PR #1294, opened 2021-11-05
Don't overwrite the HREFs for the Download IntelliJ/SBT buttons

### LINKED ISSUE #1285: Scala 2 Download Page's "Download IntelliJ" and "Download SBT" links are broken
https://www.scala-lang.org/download/scala2.html (archive: <https://web.archive.org/web/20210813043859/https://www.scala-lang.org/download/scala2.html>)

The buttons for "Download IntelliJ" and "Download SBT" have broken `href`s, with no actual link value.

s5bug:
This should fix #1285 for good. In #1286, it was unbeknownst to me that there was JS that ended up editing the HREFs of the IntelliJ and SBT links, and so the code that was initially broken there remained broken.

I've tested this by making the relevant changes in the Chrome Debugger and verifying that the HREFs aren't overwritten.

### DIFF
--- a/resources/js/functions.js
+++ b/resources/js/functions.js
@@ -260,11 +260,7 @@ function getOS() {
 $(document).ready(function() {
     if ($(".main-download").length) {
         var os = getOS();
-        var intelliJlink = $("#intellij-" + os).text();
-        var sbtLink = $("#sbt-" + os).text();
-        var stepOneContent = $("#stepOne-" + os).html()
-        $("#download-intellij-link").attr("href", intelliJlink);
-        $("#download-sbt-link").attr("href", sbtLink);
+        var stepOneContent = $("#stepOne-" + os).html();
         $("#download-step-one").html(stepOneContent);
     }
 });


julienrf (approved):
Good catch!

That being said, it seems that `#download-step-one` is not present on the HTML page, do you know when it was removed?

s5bug:
Huh, I have no clue. Should I remove this entire `ready` block then?

SethTisue:
Should I merge this now in order to get the basic bugfix in, and then we can continue discussing further changes?

(It's possible I've misunderstood the situation.)

julienrf:
Ok to merge this now.

SethTisue:
thanks Aly!

s5bug:
What should we do about `#download-step-one`? New issue, or?

SethTisue:
If you intend to tackle it yourself, you can just PR a fix directly (or send a "should we even merge this?" draft PR, if it needs discussion). But opening an issue and hoping someone else eventually deals with it is fine too.

julienrf:
> Huh, I have no clue. Should I remove this entire `ready` block then?

I am not sure. The goal was to show specialized instructions according to the OS of the user. It would be helpful to investigate whether these specialized instruction still make sense or not, and to re-introduce them, if it makes sense. I don’t have the bandwidth to do that myself, but if you are willing to help that would be greatly appreciated!

PR #1303, opened 2021-11-14
CoC: clarify personal-information bullet

SethTisue:
I think the word "doxing" is slang; the boundaries of its meaning are disputed; and accusing someone of "doxing" is potentially a very serious accusation.

So I think it's better to be clear that something could fall under this bullet without actually necessarily qualifying as "doxing" per se.

### DIFF
--- a/conduct.md
+++ b/conduct.md
@@ -101,7 +101,7 @@ Behavior that will lead to exclusion includes the following points, inspired by 
 * Violence, threats of violence or violent language directed against another person.
 * Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
 * Posting or displaying sexually explicit or violent material.
-* Posting or threatening to post other people’s personally identifying information (including so-called "doxing").
+* Posting or threatening to post other people’s personally identifying information.
 * Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
 * Inappropriate photography or recording.
 * Inappropriate physical contact. You should have someone’s consent before touching them.
--- a/conduct.md
+++ b/conduct.md
@@ -101,7 +101,7 @@ Behavior that will lead to exclusion includes the following points, inspired by 
 * Violence, threats of violence or violent language directed against another person.
 * Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
 * Posting or displaying sexually explicit or violent material.
-* Posting or threatening to post other people’s personally identifying information ("doxing").
+* Posting or threatening to post other people’s personally identifying information (including so-called "doxing").
 * Personal insults, particularly those related to gender, sexual orientation, race, religion, or disability.
 * Inappropriate photography or recording.
 * Inappropriate physical contact. You should have someone’s consent before touching them.


SethTisue:
This seems like a harmless clarification to me that doesn't alter what I understand to be the intention behind the text.

Regardless, I suggest we not merge it until:

- We wait and see if any other such minor clarification edits arise from the current episode, where CoC issues are coming up.
- We run the batch of edits (even if ends up just being this one) by the major adopters of the code for feedback.

SethTisue:
fyi @darjutak @sjrd

conduct.md
```diff
@@ -101,7 +101,7 @@ Behavior that will lead to exclusion includes the following points, inspired by
 * Violence, threats of violence or violent language directed against another person.
 * Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
 * Posting or displaying sexually explicit or violent material.
-* Posting or threatening to post other people’s personally identifying information ("doxing").
+* Posting or threatening to post other people’s personally identifying information (including so-called "doxing").
```
eed3si9n, line 104: I don't know if adding the word "including" clarifies here since as far as I know "Posting or threatening to post other people’s personally identifying information" perfectly describes "doxing".

SethTisue: maybe it’s better to omit the parenthetical altogether and just not use the word “doxing” at all.

SethTisue:
in response to review feedback, I have pushed an additional commit that removes the word "doxing" altogether

conduct.md
```diff
@@ -101,7 +101,7 @@ Behavior that will lead to exclusion includes the following points, inspired by
 * Violence, threats of violence or violent language directed against another person.
 * Sexist, racist, homophobic, transphobic, ableist or otherwise discriminatory jokes and language.
 * Posting or displaying sexually explicit or violent material.
-* Posting or threatening to post other people’s personally identifying information ("doxing").
+* Posting or threatening to post other people’s personally identifying information.
```
Daenyth, line 104: I think it would help to clarify how this relates to public information

```suggestion
* Posting or threatening to post other people’s (non-public) personally identifying information.
```

Ichoran: I like the spirit of the change, but if personal identifying information is public but hard to find, and someone digs it out of an old usenet archive or something and posts it in the context of a heated argument, that is still very bad because it is frightening and potentially endangering them.  Likewise, if someone happens to have a listed telephone number, tracking it down and posting it is bad.  So I'm not 100% sold on this change.  It might admit too much abuse for not enough clarification that sharing e.g. their twitter handle is fine.

nafg: Why do we want to add wording that suggests that it's okay to 'post or threaten to post' other people’s **public** personally identifying information?

SethTisue:
I am voluntarily withdrawing the suggestion, as it’s proving more contentious than I had anticipated.

conduct.md
Ichoran: @nafg - Well, if I am saying something like, "Hey I really like Li Haoyi's requests web server libraries" and someone else replies, "Oh really?  That sounds good to me, but I have a few tricky requirements. I'd love to talk to him, if you think he'd not mind some questions" then I'm kind of in a bind if I can't mention any identifying information at all.  You couldn't ever refer anyone to anyone if literally no personally identifying information could be posted.

Issue #1318, opened 2021-12-21
Deal with non-English-language Gitter rooms


SethTisue:
context: #1256, #1312, #1314

Yesterday I went around to all the non-English rooms and asked if the room should be kept open, and if there is some other place where that language community has congregated. I'll report back. (But I will wait a minimum of one full month before removing any rooms.)

And once done, I will update the list on https://scala-lang.org/community/

(And for any rooms that we delete as inactive, I'll attach an archive of messages here. I do not think we should keep a room open if it isn't being used and nobody asks for it to be kept open; in those cases, I think it's fine to encourage people to move to whatever alternative platform they self-organize onto.)

SethTisue:
So far the following rooms have requested to remain open:

* scala/cn
* scala/fr

SethTisue:
I went around and issued "second/final warning" on shutdowns to the other rooms. (Intended shutdown date, though I wasn't so specific: Jan 20.)

SethTisue:
scala/ar deleted; archive: [gitter-room-messages-2022-1-23.ndjson.zip](https://github.com/scala/scala-lang/files/7921743/gitter-room-messages-2022-1-23.ndjson.zip)

(I can only do one room at a time, have to wait 3 hours between, because rate limiting)

SethTisue:
although scala/cn and scala/fr requested to remain open, little to no actual traffic has ensued, and scala/scala itself has pretty much ground to a halt as well, so it's a bit tempting to just shut it all down

(otoh, it doesn't cost me much to drop by every couple days and check for spam, CoC, lost newbies)

PR #1331, opened 2022-01-24
remove dead Gitter rooms

SethTisue:
context: #1318


SethTisue:
merging on the fast track, since the dead links are making CI fail (well, only one is actually dead so far, but the rest will be dead soon)

PR #1332, opened 2022-01-24
fix dead ScalaUA links

SethTisue:
their website is dead but their Twitter still exists
so I guess let's use that


SethTisue:
fast-track merge because the dead links are failing CI

PR #1335, opened 2022-02-03
Update Community Resources

adamgfraser:
Add ZIO and Zymposium to community resources.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -76,6 +76,7 @@ Scala-oriented Discord servers operated by the community include:
 * **[Scalameta](https://discord.gg/RFpSVth)**: Scalameta-based tooling: Metals, Scalameta, Scalafix, Scalafmt, and Mdoc
 * **[Play Framework](https://discord.gg/g5s2vtZ4Fa)**: the Play web framework for Scala and Java
 * **[Typelevel](https://discord.gg/XF3CXcMzqD)**: the Typelevel ecosystem for pure-functional programming in Scala
+* **[ZIO](https://discord.gg/3F3xKppyHU)**: the ZIO ecosystem for high performance asynchronous and concurrent programming in Scala
 
 English-language Scala rooms on other chat platforms besides Discord include:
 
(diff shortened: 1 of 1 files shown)


SethTisue:
Of the three changes here, 1 and 3 look mergeable to me. I don't think the second change fits; it's a list of learning resources that are focused on the Scala language itself, and I don't think we want to broaden the scope beyond that.

adamgfraser:
I'm not sure what it means for a learning resource to be "focused on the Scala language itself".

Zymposium covers a variety of topics in Scala including use of specific language features as well how to use different libraries to solve various problems in Scala. For example:

* [Zymposium - Variance](https://www.youtube.com/watch?v=2uEjXuFBsBk&list=PLvdARMfvom9C8ss18he1P5vOcogawm5uC&index=12)
* [Zymposium — Path Dependent Types](https://www.youtube.com/watch?v=w2rcHCqdn-o&list=PLvdARMfvom9C8ss18he1P5vOcogawm5uC&index=19)
* [Zymposium — Type-Level Sets](https://www.youtube.com/watch?v=N_it0qwk7_U&list=PLvdARMfvom9C8ss18he1P5vOcogawm5uC&index=18)
* [Declarative vs Executable Encodings](https://www.youtube.com/watch?v=OD1Yr48-0Js&list=PLvdARMfvom9C8ss18he1P5vOcogawm5uC&index=27)
* [Zymposium - Caliban](https://www.youtube.com/watch?v=mzqsXklbmfM&list=PLvdARMfvom9C8ss18he1P5vOcogawm5uC&index=11)

I suppose it would be possible to have a principle that these learning resources should be focused exclusively on language features and not rely on any libraries outside the standard library. If so would be happy to update the PR to delete the reference to Zymposium as well as other learning references that are not focused on the Scala language itself, such as Scala Exercises, which includes materials on a variety of libraries outside the Scala standard library such as Cats, Shapeless, Doobie, ScalaCheck, Circe, Fetch, and Monocle.

Alternatively, I would tend to think that one of the great things about Scala is that there are a lot of awesome libraries written in Scala that allow you to solve a variety of problems that cannot and should not be expected to be solved with the standard library alone, such as standing up a web service. In this case learning resources would cover materials that focus on specific language features as well as how to solve various problems in Scala, which would necessarily include discussion of libraries appropriate to solve those problems.

Let me know which direction you would like to go.

SethTisue:
I'd suggest you drop the Zymposium link from this PR so then the rest can be promptly merged without further discussion.

Then the Zymposium thing could go in a separate PR and we can see what somebody from the Scala Center thinks, perhaps.

SethTisue:
(closing for inactivity, as is customary here; we can reopen if activity resumes)

PR #1344, opened 2022-02-22
Converge on one get

bishabosha:
This is a rework of


bishabosha:
Idea for follow up would be to prettify the layout of the get-started page on the docs site, as we have access to translations there

_layouts/frontpage.html
```diff
@@ -74,7 +74,7 @@ <h1>{{page.headerTitle}}</h1>
             <div class="circle-solid">
               <img src="/resources/img/frontpage/scala-spiral.png" alt="">
               <div class="scala-version">
-                <span>Scala</span>
+                <!-- <span>Scala</span> -->
               </div>
```
sjrd, line 78: I guess this should be removed entirely, at this point?

download/index.md
```diff
@@ -2,17 +2,17 @@
...
 - title: "Scala 3 documentation"
   url: "https://docs.scala-lang.org/scala3/"
 - title: "Migrating from Scala 2 to Scala 3"
   url: "https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html"
+
+newcomertutorials:
+- title: "Tour of Scala"
+  url: "https://docs.scala-lang.org/tour/tour-of-scala.html"
+- title: "Scala Book"
+  url: "https://docs.scala-lang.org/scala3/book/introduction.html"
```
sjrd, line 15: Perhaps we should put the Scala Book first?

index.md
```diff
@@ -9,33 +9,29 @@ headerButtonUrl: "/what-is-scala/"
...
+apiDocs:
+  mainTitle: "Learn Scala"
+  mainUrl: "https://docs.scala-lang.org"
+  subtitle: "Scala Book"
+  subtitleLink: "https://docs.scala-lang.org/scala3/book/introduction.html"
+  links:
+    - title: "New in Scala 3"
+      url: "https://docs.scala-lang.org/scala3/new-in-scala3.html"
+    - title: "Migrate From Scala 2"
+      url: "https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html"
```
sjrd, line 32: With the focus of the front page shifting away from the Scala 2/3 distinction, I feel like those two items don't really make much sense anymore. Not sure what to replace them with, though.

_layouts/frontpage.html
bishabosha: done

index.md
bishabosha: Screenshot 2022-02-24 at 00 26 38

updated as above

sjrd (approved):

index.md
```diff
@@ -9,33 +9,29 @@ headerButtonUrl: "/what-is-scala/"
...
+  mainUrl: "https://docs.scala-lang.org"
+  subtitle: "Scala Book"
+  subtitleLink: "https://docs.scala-lang.org/scala3/book/introduction.html"
+  links:
+    - title: "Tour of Scala"
+      url: "https://docs.scala-lang.org/tour/tour-of-scala.html"
+    - title: "Online Courses"
+      url: "https://docs.scala-lang.org/online-courses.html"
+    - title: "Scala 3 Reference"
+      url: "https://docs.scala-lang.org/scala3/reference/overview.html"
```
julienrf, line 34: I find it a bit confusing that “Language Specification” is on the left side (and it is only about Scala 2), but “Scala 3 Reference” is on the right side, although both links point to a similar type of content (the language specification, for Scala 2 and Scala 3). I wonder if we could have just one link “Language Specification” that would point to the Scala 3 reference, and in that page we should add a link to the Scala 2 specification. Or, maybe just put both links (Scala 2 spec, and Scala 3 spec) on the left side (just below the “Current API Docs” and “API Docs (All Releases)” links)?

download/index.md
```diff
@@ -2,17 +2,17 @@
...
 - title: "Scala 3 documentation"
   url: "https://docs.scala-lang.org/scala3/"
 - title: "Migrating from Scala 2 to Scala 3"
   url: "https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html"
+
+newcomertutorials:
+- title: "Scala Book"
+  url: "https://docs.scala-lang.org/scala3/book/introduction.html"
+- title: "Tour of Scala"
+  url: "https://docs.scala-lang.org/tour/tour-of-scala.html"
```
julienrf, line 15: It is a bit confusing that here we mix links to Scala 3 resources (book, online courses) and Scala 2 resources (tour). I believe we should plan to update the tour to Scala 3 soon, maybe this is already what you wanted to do?

_layouts/download.html
```diff
@@ -20,41 +20,36 @@ <h1>{{page.title}}</h1>
   <div class="wrap">
```
julienrf, line 20: I think most of the content on this page is redundant with the index page (the main two links point to getting-started.html and download/all.html, which are already provided on the index page. The other links are also provided on the index page, and they are not really related with the action of “downloading” Scala). I wonder if we should just remove this page? (To do it properly, the URL https://scala-lang.org/download/ should still work, but it should redirect to https://docs.scala-lang.org/getting-started.html)

_includes/downloads-scala2.html
```diff
@@ -1,70 +1,38 @@
...
-    <div class="download-left">
-      <span class="or">or</span>
-      <div class="description">
-        <img src="/resources/img/download/arrow-left.png" alt="">
-        <p>Best if you prefer a full-featured IDE (recommended for beginners)</p>
+      <h2>Recommended Download</h2>
+      <div class="download-options">
+        <a href="https://docs.scala-lang.org/getting-started/index.html" class="btn-download" id="get-started-link">
+          <i class="fa fa-download"></i>
+          <span>Get Started With Scala</span>
```
julienrf, line 15: Users came here by following the link “pick a specific release”, which is labeled “best if you have an advanced use case”. I don’t think we should point them back to the general getting started instructions.

bishabosha:
potential blocker - need a way to ensure that the "other releases" show in full - 
i.e. latest 2.12 / 2.11/ 2.10 maintenance releases - these are in metadata of the download/scala2.md page, need a way to source this once in all download pages

index.md
julienrf: Resolved after internal discussion: let’s keep this change for now, and revisit later when the Scala 3 reference is more mature.

_layouts/download.html
julienrf: Resolved after internal discussion: let’s keep it for now and see if we can go further in a separate PR.

bishabosha:
the new commit removes the old `/download/scala2.md` and `/download/scala3.md` pages, with a new source for the "current releases" metadata - `_data/scala-releases.yml`. Also remove the link to "get started with scala" from the release pages.

_includes/downloads-scala2.html
bishabosha: removed in next commit

download/index.md
bishabosha: I think it will be a bit of a mess until the structure of the documentation is finalised, so we could revisit at that point

index.md
bishabosha: my comment from offline is that this would be ideal, but these two documents have different purposes - i.e. the scala 3 reference is an addition to the spec, with more discussion added - would be good to update those pages together

download/index.md
```diff
@@ -2,17 +2,17 @@
 layout: download
 title: Download
```
julienrf, line 3: I suggest adding:

~~~ yaml
redirect_from:
  - /download/scala2.html
  - /download/scala3.html
~~~

To not break the possible existing links that point to these URLs.

_data/scala-releases.yml
```diff
@@ -0,0 +1,20 @@
+- category: current_version
```
julienrf, line 1: Do you use the category somewhere?

_data/scala-releases.yml
```diff
@@ -0,0 +1,20 @@
...
+  version: 2.12.15
+  release_date: September 14, 2021
+- category: "maintenance_version"
+  title: "Last 2.11.x maintenance release"
+  version: 2.11.12
+  release_date: "November 9, 2017"
+- category: "maintenance_version"
+  title: "Last 2.10.x maintenance release"
+  version: 2.10.7
+  release_date: "November 9, 2017"
```
julienrf, line 20: We should probably update the Scala 2 and Scala 3 release instructions to point them to this file, which needs to be updated on every new release.

_data/scala-releases.yml
bishabosha: the category is not used, as far as I can see but this "field" existed in the current `/download/scala2.md` so I included it here

_data/scala-releases.yml
bishabosha: @SethTisue @Kordyjan after this PR is merged this page will need updating on each Scala release, instead of `/download/scala2.md`

download/index.md
bishabosha: thanks!

julienrf (approved):
:rocket:

SethTisue:
LGTM. (In a general way — I didn't try to really pick this apart, but in principle the changes seem good to me).

I left a breadcrumb in the Scala 2 release steps so we remember to make any needed adjustments the next time we do a Scala 2 release: scala/scala-dev@193930578e84689e04ee3d907f68a34c40883b00

SethTisue (approved):

PR #1352, opened 2022-03-14
link api docs from release page

bishabosha:
I've added API docs link to the "ribbon" of a release if there is a known API docs for it.

Screenshot 2022-03-14 at 15 01 09

open questions:
- will old API docs get deleted eventually, making dead links (I guess a CI check may fail here)
- is the ribbon appropriate, alternatively we could have a section in the main body to link also the compiler and reflect APIs


julienrf (approved):
Nice!

SethTisue:
nice! I'll take care of making the corresponding change over in https://github.com/scala/make-release-notes: https://github.com/scala/make-release-notes/issues/437

PR #1355, opened 2022-03-24
Fix typos in the GSoC 2022 post

anatoliykmetyuk:
As per @Philippus review

PR #1356, opened 2022-03-24
Apply suggestion

Philippus:
as per https://github.com/scala/scala-lang/pull/1354#discussion_r833138228


Philippus:
duplicate of https://github.com/scala/scala-lang/pull/1355.

PR #1365, opened 2022-04-06
add SwissBorg to supporters

bishabosha:
Screenshot 2022-04-06 at 11 21 28


_sass/layout/maintenance.scss
```diff
@@ -70,6 +70,10 @@
             margin-right: 10px;
             margin-bottom: 8px;
 
+            img.small-logo {
```
sjrd, line 73: We should give a more "semantic" name to this class, that is more about they *why* it's special than *how* it's special. Consider calling it `logo-with-wide-aspect-ratio`, for example.

sjrd (changes requested):

bishabosha:
I changed it a bit

sjrd (approved):

PR #1366, opened 2022-04-06
Cleanup homepage

bishabosha:
remove the sections from homepage:
- IDEs
- run scala in the browser
- events
- courses/trainings
- talk to us
- new blog
- twitter feed
- scala staircase

rework:
- Scala-ecosystem is now more space efficient, promotes awesome-scala
- scala runs on... is after in a nutshell
- in a nutshell links have buttons telling you to click them
- community page now displays events and trainings (resources/json/trainings.json now actually is populated)

Screenshot 2022-04-07 at 10 06 51

Screenshot 2022-04-06 at 15 17 11

Screenshot 2022-04-06 at 15 18 24

Events with an example training (not committed)
Screenshot 2022-04-06 at 15 21 20


_layouts/frontpage.html
```diff
@@ -125,6 +109,7 @@ <h2><span>Scala in a Nutshell</span></h2>
         <div class="scala-item">
           <h3>{{scalaItem.shortTitle}}</h3>
           <p>{{scalaItem.shortDescription}}</p>
+          <div class="button">click to expand</div>
```
julienrf, line 112: I find “Click to expand” a bit long and repetitive. Do you think we could replace it with “…” for instance? (not sure “…” is self-explanatory enough, though)

_layouts/frontpage.html
```diff
@@ -136,227 +121,75 @@ <h3>{{scalaItem.shortTitle}}</h3>
...
-          <li></li>
-        {% endunless %}
+      {% unless forloop.last %}<li></li>{% endunless %}
       {% endfor %}
     </ul>
   </div>
 </section>
 
-<!-- Run Scala in your browser -->
-{% include scastie.html %}
```
julienrf, line 174: I think it was nice to have this one. At least, the ability to have executable code snippets from the front page is cool. Maybe the “Scala in a Nutshell” examples could be made executable? (that would probably be a big effort though)

julienrf (approved):
I am a bit hesitant to remove all this information, but it’s true that it’s still available on more specific pages…

Should we completely remove the “trainings”?

_layouts/frontpage.html
SethTisue: I think "More..." would be sufficient

_layouts/frontpage.html
SethTisue: Note that there is now a "Playground" link all the way at the top of page.

I think it's only worth spending the space on this on the front page if we actually give people sample snippets they can run, as you suggest. So I'd be in favor of bringing this back in that form in the future. At present, it just dumps visitors into a thing that isn't useful unless you already know Scala, so I think it should go.

SethTisue (approved):

_layouts/frontpage.html
bishabosha: I have updated the PR description with a screenshot of using "More..."

julienrf (approved):

PR #1369, opened 2022-04-17
add Scala Symposium 2022 to events page

SethTisue:


_events/2022-06-06-scala-symposium.md
```diff
@@ -0,0 +1,10 @@
+---
+category: event
+title: Scala Symposium
+logo: /resources/img/scala-symposium.png
+location: Berlin, Germany
+description: "the leading forum for researchers and practitioners related to the Scala programming language"
+start: 6 June 2022
+end: 6 June 2022
```
Philippus, line 8: Ecoop runs until 10 June 2022. Is the Scala Symposium on the first day?

SethTisue: It's difficult to be 100% sure, but that's a plausible way to read https://2022.ecoop.org/home/scala-ecoop-2022 (it says that June 6 is the "symposium date")

Philippus: Ah indeed! I completely missed the `new` entry.

PR #1371, opened 2022-04-25
compress sublinks to single line

bishabosha:
Screenshot 2022-04-25 at 14 18 00

make the table of links more compact


julienrf:
I think the medium size does not look well compared to the current state:

Screenshot from 2022-04-25 16-13-08

bishabosha:
I changed the top margin when in the smallest size, now it looks like:
Screenshot 2022-04-25 at 16 14 29

bishabosha:
now I center the content using columns in the medium size:
Screenshot 2022-04-25 at 16 18 22

Issue #1374, opened 2022-04-30
Htmlproofer failing in CI runs


SethTisue:
at e.g. https://github.com/scala/scala-lang/runs/6236625743 we see:

```
htmlproofer 3.10.2 | Error:  HTML-Proofer found 3 failures!
[12](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:12)
- ./_site/blog/2017/08/28/gsoc-connecting-contributors-with-projects.html
[13](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:13)
  *  External link https://developer.github.com/v3/ failed: 403 No error
[14](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:14)
  *  External link https://developer.github.com/v4/ failed: 403 No error
[15](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:15)
- ./_site/blog/2018/06/04/scalac-profiling.html
[16](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:16)
  *  External link https://docs.github.com/en/authentication/connecting-to-github-with-ssh failed: 403 No error
[17](https://github.com/scala/scala-lang/runs/6236625743?check_suite_focus=true#step:6:17)
Error: Process completed with exit code 1.
```

Philippus:
I think it may have been a temporary thing. I've updated two of those links in https://github.com/scala/scala-lang/pull/1375, because they changed.

SethTisue:
it's strange, I don't know what to make of it

after #1375, list of failures is:

```
- ./_site/blog/2017/08/28/gsoc-connecting-contributors-with-projects.html
  *  External link https://docs.github.com/en/graphql failed: 403 No error
  *  External link https://docs.github.com/en/rest failed: 403 No error
htmlproofer 3.10.2 | Error:  HTML-Proofer found 3 failures!
- ./_site/blog/2018/06/04/scalac-profiling.html
  *  External link https://docs.github.com/en/authentication/connecting-to-github-with-ssh failed: 403 No error
```

SethTisue:
there is some previous history on 403s at #945

griggt:
How odd.  I tried to `curl https://docs.github.com/en/rest` and I also got a 403.

[...]

PR #1376, opened 2022-05-01
Add Accept-Encoding header to html-proofer

Philippus:
with values "gzip, deflate". I think this can help solve #1374. Unfortunately I couldn't test this out locally.


SethTisue:
> Unfortunately I couldn't test this out locally.

perhaps #1377 helps?

SethTisue:
followup PR: #1378

PR #1377, opened 2022-05-02
run `bundle update --minor`

SethTisue:
I was unable to preview the site locally without doing this first

I used `--minor` because it's more conservative and seems sufficient to fix the errors I was getting

I was seeing:

> Your bundle is locked to minitest (5.14.1) from rubygems repository https://rubygems.org/ or
installed locally, but that version can no longer be found in that source. That means the author of
minitest (5.14.1) has removed it. You'll need to update your bundle to a version other than minitest
(5.14.1) that hasn't been removed in order to install.

and

> Your bundle is locked to listen (3.2.1) from rubygems repository https://rubygems.org/ or installed
locally, but that version can no longer be found in that source. That means the author of listen
(3.2.1) has removed it. You'll need to update your bundle to a version other than listen (3.2.1)
that hasn't been removed in order to install.


SethTisue:
as usual with Ruby stuff I have no real idea what I'm doing, but nobody else here seems to either, so, merging...

PR #1378, opened 2022-05-02
move fix for #1376 to right part of CI config

SethTisue:


SethTisue:
PR runs don't test the change, only merge runs. so, merging in order to test.

SethTisue:
oh, merging isn't enough, we'd have to wait for the next cron run

I'm not that patient, so I've submitted #1379 which should allow us to manually trigger the same stuff the cron runs do

PR #1379, opened 2022-05-02
allow running external link-checking manually

SethTisue:


SethTisue:
this worked — I was able to manually trigger a run at https://github.com/scala/scala-lang/actions/workflows/ci.yaml

SethTisue:
and the local-only vs external switch seems to be working:

Screen Shot 2022-05-01 at 5 37 00 PM

PR #1380, opened 2022-05-04
Add default HTML-Proofer Typhoeus configuration to the CI command line

griggt:
Apparently when specifying `--typhoeus-config` on the command line, all the defaults are discarded, and must be re-specified if desired.

The default values were taken from:
https://github.com/gjtorikian/html-proofer/blob/1bab3a1a18e95a10378371ddf000df9bea01740e/lib/html-proofer/configuration.rb#L36-L44

This should hopefully fix the rash of 403s since #1376 and #1378 were merged.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,4 +30,4 @@ jobs:
       - name: Proofing (all links)
         if: github.event_name != 'pull_request'
         run: |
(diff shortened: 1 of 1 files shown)


SethTisue:
The PR run doesn't test this, so let's just merge it and then wait and see what happens with the next cron run.

Thanks Tom!

PR #1381, opened 2022-05-11
alt-details does not need javascript

bishabosha:
using some css tricks to implement the accordion behaviour without JS:
Screenshot 2022-05-11 at 14 22 06


julienrf (approved):
Nice trick :wink:

PR #1382, opened 2022-05-11
tabs can operate without JS

bishabosha:
Support tab switching without JavaScript

based on https://levelup.gitconnected.com/tabbed-interfaces-without-javascript-661bab1eaec8

because all states must be known there are explicitly up to 9 dynamic tabs supported, we could add more in the future

### DIFF
--- a/resources/js/functions.js
+++ b/resources/js/functions.js
@@ -313,41 +313,13 @@ $(document).ready(function () {
 });
 
 $(document).ready(function () {
-  $('.tabsection').each(function () {
-    var tabsection = this;
-    $(tabsection).find('.nav-tab > .item-tab > .item-tab-link').each(function () {
-      var tabLink = this;
-      var targetTab = $(tabLink).attr('data-target');
-      $(tabLink).click(function () {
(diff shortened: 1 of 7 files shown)


resources/js/functions.js
```diff
@@ -312,40 +312,17 @@ $(document).ready(function () {
...
       os = 'linux';
     }
-    $("#install-cs-setup-tabs").find('.nav-tab > .item-tab > .item-tab-link').each(function () {
+    $("#install-cs-setup-tabs").find('input').each(function () {
       var targetTab = $(this).attr("data-target");
       if (targetTab === os) {
-        $(this).click();
+        $(this).prop("checked", true);
       }
     });
```
julienrf, line 327: The loop could probably be replaced by more precise selector:

```suggestion
    $("#install-cs-setup-tabs").find('input[data-target=' + os + ']').prop("checked", true);
```

julienrf (approved):
OK, why not!

You will probably have to update the content of the tab `4-other` in the other languages (I can help with the french one).

PR #1386, opened 2022-05-25
Add an instruction to install bundler in the README

anatoliykmetyuk:


README.md
```diff
@@ -31,6 +31,14 @@ docker-compose up --build
 
 ### With Bundler:
 
+Install bundler first:
+
+```
+sudo gem install bundler
```
julienrf, line 37: Don’t you need to install both the `bundler` and `jekyll`?

julienrf: (maybe `jekyll` is install when you `bundle install`?)

anatoliykmetyuk: No idea. To my memory, I didn't use `jekyll` before, so it was likely not installed on my machine. I suppose installing only `bundler` did the trick for me.

julienrf (approved):

PR #1387, opened 2022-05-27
Long-term compatibility plans for Scala 3

Kordyjan:
The post with communication about our decision to drop `-scala-output-version` and introduce the new Scala LTS / Scala Next scheme.


bjornregnell:
Regarding the roadmap picture: What does the minor release in the black track between 3.3 and 3.4 mean? If it represents 3.3.1 etc. then it seems to be a duplicate of the first minor release on the green track of 3.3... So it seems to that there should be no black minor releases in the picture as all LTS releases should be on the green LTS track. Or am I missing something here?  @Kordyjan

Kordyjan:
@bjornregnell Good catch! The 3.3.1 patch release (3.3 & 3.4 are minors) should only be present on the LTS track. The picture is fixed now.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,85 @@
...
+
+More importantly, during work on 3.2, we realized that maintaining this flag may be more challenging than anticipated. With time, it will be increasingly harder to be sure that our handling of it is correct. **Thus we have decided to remove the possibility of configuring the output version altogether in Scala 3.2.**
+
+## Our solution: Scala LTS & Scala Next
+
+Finally, we have reached the agreement that the best course of action would be to split Scala development into two lines, called “Scala LTS” and “Scala Next”. Both would live as separate branches in the compiler repository. Both will have separate, but possibly synchronized releases. Porting changes from one to the other should be easy and common.
+
+Scala Next will be the line that the development of the language is taking place on. It would receive frequent minor updates and all experimental language features will leave here.
```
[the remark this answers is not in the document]
Kordyjan: Thanks! It is fixed now.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,85 @@
...
+
+The new model doesn’t change our compatibility guarantees. We are keeping forward and backward compatibility within a single minor version and backward compatibility across different minor versions. That means that the code compiled with Scala LTS can be used as a dependency by all Scala Next versions newer than said LTS, but not the other way around.
+
+## The (future) best practices
+
+Once the LTS & Next model is implemented, we suggest the following practices for different groups of users:
+
+### Owners of commercial projects
+
+You should stick to the LTS version. It will give you the best tooling support and a stable compiler. If you are migrating a bigger project from Scala 2, you should migrate it straight to the newest LTS.
```
Jasper-M, line 75: Could you explain more what the risks for bigger or commercial projects would be if they choose not to stick with an LTS release?

sjrd: IMO this is worded a bit too strongly. I think there are two valid choices for commercial projects:

* Stick to the LTS version -> update less often (from LTS to LTS), with potentially more changes to deal with at once, or
* Be up-to-date to the latest minor version -> update more often, but with smaller deltas

In both cases, you will get the best tooling support and a stable compiler.

The catch is that you take the option of the latest minor version, you will *have to* update more often in order to keep having the best tooling support, performance improvements and bug fixes (including potential security fixes).

If you are prepared to keep up in that sense, it is perfectly fine to use the latest version in a commercial project.

Jasper-M: > The catch is that you take the option of the latest minor version, you will have to update more often in order to keep having the best tooling support, performance improvements and bug fixes (including potential security fixes).

I suppose this was the only option before LTS releases were introduced?

bjornregnell: > If you are prepared to keep up in that sense, it is perfectly fine to use the latest version in a commercial project.

Yes that is indeed important information so that commercial projects, are not "scared away" from being up-to-date if they care about frequent updates. It is good to know that following the frequent release train also provides a stable compiler. (As I understand it, it's the RC:s that are less stable, and nightly even less stable, in terms of how much real-world testing has been made so far.)

sjrd: > I suppose this was the only option before LTS releases were introduced?

Yes, indeed.

> As I understand it, it's the RC:s that are less stable, and nightly even less stable, in terms of how much real-world testing has been made so far.

Also yes, indeed. ;)

Kordyjan: > * Be up-to-date to the latest minor version -> update more often, but with smaller deltas

This is impossible for the large majority of the codebases in the industry. Nevertheless, it is true. I will add a sentence about it being possible if the code maintainers are willing to commit to the constant updates.

Kordyjan:
``` md
You may also choose to be up to date with the latest minor version. That will allow you to use new features as soon as they are added to the language. However, you will need to be ready for frequent compiler updates. Otherwise, your project may end up with an unsupported version of the language.
```
was added.

bjornregnell: I guess you want to update this blog as well, and remove some stuff:
https://www.scala-lang.org/blog/2022/04/12/scala-3.1.2-released.html
if scalaOutputVersion is cancelled, as well as sbt docs etc...
Otherwise there is a risk that people get hold of this and miss the updated policy on LTS etc.

Kordyjan: `-scala-output-version` was supported only in one pre-release version of the sbt, and it is being removed right now. Regarding the previous blog post, it may indeed be a good idea to add an update there.

SethTisue:
Here comes the usual feedback that I always give on blog posts :-)

I would suggest beginning the entire blog post with a short "Summary" section that says as concisely as possible what the takeaways are from the entire thing.

Then go into the full version, including all the material about how the conclusions were arrived at.

Doing it this way serves multiple purposes. It helps the reader decide whether they need or want to read the rest. And even for people who do end up reading the whole thing, it's useful to have the summary to check their understanding against, and to consult again when returning to the blog post later. (This is one that library maintainers are likely to return to multiple times.)

Sorry I can't be in Krakow this weekend like I'd planned :-/

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
...
+
+Once the LTS & Next model is implemented, we suggest the following practices for different groups of users:
+
+### Owners of commercial projects
+
+We suggest you stick to the LTS version. It will give you the best tooling support and a stable compiler. If you are migrating a bigger project from Scala 2, you should migrate it straight to the newest LTS.
+
+You may also choose to be up to date with the latest minor version. That will allow you to use new features as soon as they are added to the language. However, you will need to be ready for frequent compiler updates. Otherwise, your project may end up with an unsupported version of the language.
+
+### Library maintainers
```
rossabaker, line 79: I like the strategy and these "future" best practices, but as a library maintainer, we would benefit from best practices now.  The most aggressive dependency decides for all of us, and it has been chaotic downstream.

Until the first LTS (3.3?), is the recommendation that we stick to 3.1 (where most of us are now) or jump on 3.2?  Can we say so here?

armanbilge: Right, if the implication of this post is that when 3.2 lands, 3.1 will stop getting (security) patches and thus is effectively EOL, seems to me that libraries may as well move on. OTOH, if 3.1 or something was blessed an "STS" (short-term-support?) to hold us over until the real LTS lands, that would at least get security patches or something, then it could make sense to pin.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
...
+
+## The problem
+
+While this sounds good, it can be problematic for an intertwined ecosystem. Even though we do what we can to make the transition between minor versions of the compiler as smooth as possible, there will always be projects stuck on older versions for various reasons. That means that they may not be able to update some of their dependencies if said libraries updated themselves to the newer compiler. That, in order, encourages libraries to stay on the oldest possible version of the compiler, which bars them from potentially beneficial improvements in the newer compilers.
+
+Let’s examine the simple example. We have a project that for some reason, is stuck on Scala 3.1. We depend on version 1.3.5 of some library that was compiled with Scala 3.1. The library publishes version 1.4.0 with the compiler updated to 3.2. We cannot update our dependency because of output incompatibility. That also means that all patches for security issues won’t be available unless the library author decides to backport them and release them as version 1.3.6. Knowing all this, the library author would never migrate from Scala 3.1 to 3.2.
+
```
bjornregnell, line 29: "would never" is a bold statement and "never" is a very long time...
Perhaps phrase it as "is not not likely to want to" or something.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
+---
+layout: blog-detail
+post-type: blog
+by: Paweł Marks, VirtusLab
+title: Long-term compatibility plans for Scala 3
+---
+
+Scala 3 has stronger compatibility guarantees than Scala 2 ever had. That allows us to improve the language at a much faster pace than it was possible before 3.0. However, with the current scheme, when you bump the dependencies of your project, you may also have to bump the version of Scala used by the project. After an extended evaluation, we have decided to solve them by creating a new kind of Scala release - Scala LTS, that will be supported and receive updates for at least three years after the initial release.
```
felher, line 8: "we have decided to solve them" - sorry, maybe I'm just blind, but what does "them" refer to?

SethTisue:
If people have feedback not about the clarity of the blog post (which is what the comments on this PR are for), but about the versioning scheme itself, where do they go to lodge that feedback? I think the blog post should be explicit on that point.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
Kordyjan: 😅 Refactoring leftover

blog/_posts/2022-05-30-long-term-compatibility-plans.md
Kordyjan: I wonder if and how I should put that recommendation in this post.

SethTisue:
> Do you have something other in mind when speaking about "Summary"?

Imagine I am a library author reading this post. What do you actually want me to _do_ after reading?

This might be in two parts: what do you want me to do _right now_, and what do you currently expect or propose that I will need to do once the new scheme has been fully rolled out?

What is the upshot, the take-home? Do you want me (the library author) to carefully read the entire blog post and give feedback? Do you want me to drop something from my current CI configuration (perhaps `-scala-output-version`) or add something?

More specifically: what is the expected status of 3.0? What is the expected status of 3.1? What is the expected status of 3.2?

Imagine I haven't been following all this discussion and debate and I just want to do know what you, the maintainers of Scala 3, _want_ from me, a library author. Tell me what to do.

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Configurable output version
+
+This was the solution that we invested the most into. It was discussed in detail in [the previous blog post]({{ site.baseurl }}/blog/2022/04/12/scala-3.1.2-released.html#configurable-scala-output-version). To summarize: in 3.1.2, we have shipped a new experimental compiler flag that allowed developers to make the compiler generate output that is usable by older versions of the compiler. While this would allow all projects to receive important fixes, it would probably result in a strange state where all the libraries are released with output version 3.1, similar to how nearly all java libraries are still released with output version 8, even though Java 8 is eight years old.
+
+More importantly, during work on 3.2, we realized that maintaining this flag may be more challenging than anticipated. With time, it will be increasingly harder to be sure that our handling of it is correct. **Thus we have decided to remove the possibility of configuring the output version altogether in Scala 3.2.**
+
+## Our solution: Scala LTS & Scala Next
```
lrytz, line 55: I feel like this section only tells me what the solution is without saying how it solves the problem.

IIUC, the goal is to encourage library maintainers to stick to LTS versions as much as they can. In return, we promise to maintain these versions so that they don't need to use some old, non-supported Scala release. Does that sound about right?

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
...
+
+Scala Next will be the line that the development of the language is taking place on. It would receive frequent minor updates and all experimental language features will live here.
+
+Scala LTS on the other hand will be the stable long-term support line. One LTS will only receive patch updates which means that all releases of the same LTS line will be forward and backward compatible in terms of output. The only changes between releases will be bug fixes, non-language changes (doctool, semanticDB, reporting), and small quality-of-life enhancements (only if we are sure we are not breaking any compatibility guarantees. After more than two years we may decide to nominate one of the releases from the Next line as the new LTS. After that, we guarantee that the old LTS will be supported for at least one additional year. That means that the time of support for each LTS release is at least three years.
```
lrytz, line 61: There's a missing closing parenthesis in here

blog/_posts/2022-05-30-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,87 @@
...
+
+The new model doesn’t change our compatibility guarantees. We are keeping forward and backward compatibility within a single minor version and backward compatibility across different minor versions. That means that the code compiled with Scala LTS can be used as a dependency by all Scala Next versions newer than said LTS, but not the other way around.
+
+## The (future) best practices
+
+Once the LTS & Next model is implemented, we suggest the following practices for different groups of users:
+
+### Owners of commercial projects
+
+We suggest you stick to the LTS version. It will give you the best tooling support and a stable compiler. If you are migrating a bigger project from Scala 2, you should migrate it straight to the newest LTS.
```
sideeffffect, line 75: My humble 2 cents:
The OpenJDK authors recommend something along the lines of
 * using the newest Java version when the project is in the state of active development
 * once the project moves to the maintenance phase of its life, update up to the at-that-time newest LTS version and stay there

https://news.ycombinator.com/item?id=28821316

Maybe we could recommend something similar here about Scala version?

blog/_posts/2022-05-30-long-term-compatibility-plans.md
armanbilge: > I wonder if and how I should put that recommendation in this post.

You definitely should :) it should go in the summary section: "if you are a library author, here's what you should to today".

SethTisue:
I hope I didn't stall this PR with my well-meaning advice :-)

blog/_posts/2022-05-30-long-term-compatibility-plans.md
Kordyjan: I've added `We think this will allow projects to retain a more conservative approach to updating the compiler versions and still receive the bug fixes and some improvements`. Do you think it needs to be elaborated further?

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Output compatibility (sometimes called *binary compatibility*, especially in Scala 2 contexts) is much more tricky and essential for our long-term plans for Scala. In the rest of this post, the term *compatibility* will always refer to output compatibility. We can say that compiler `A` is output compatible with compiler `B` when compiler `B` can use the output (binaries and TASTy files) generated by compiler `A` and understands it correctly.
+
+We can further subdivide compatibility to forward and backward. We say that the two versions are forward compatible when the older compiler can depend on the output of the newer one. Conversely, backward compatibility means that the newer compiler can use the output of the older version. Scala 3 guarantees backward compatibility between all releases and forward compatibility between patch releases in the same minor line. That means that Scala `3.a.x` can consume the output of Scala `3.b.y` only if `a' is greater or equal to`b`. For example, the output of the Scala 3.0.2 compiler can be used by Scala 3.0.1 or 3.1.0. On the other hand, code compiled with Scala 3.1.0 cannot be a dependency of any project compiled with 3.0.2.
```
bishabosha, line 29:
```suggestion
We can further subdivide compatibility to forward and backward. We say that the two versions are forward compatible when the older compiler can depend on the output of the newer one. Conversely, backward compatibility means that the newer compiler can use the output of the older version. Scala 3 guarantees backward compatibility between all releases and forward compatibility between patch releases in the same minor line. That means that Scala `3.a.x` can consume the output of Scala `3.b.y` only if `a` is greater or equal to `b`. For example, the output of the Scala 3.0.2 compiler can be used by Scala 3.0.1 or 3.1.0. On the other hand, code compiled with Scala 3.1.0 cannot be a dependency of any project compiled with 3.0.2.
```

Kordyjan:
So, unfortunately, I wasn't able to find time to work on the announcement (or many other things, sadly) lately. But I'm back at it, and I believe we should be able to have it in a satisfying state soon.

I've added a TL;DR-ish section at the top. I don't know if this is the right approach, so I'm open to suggestions.

Also, this post was initially meant to be only an announcement about dropping `--scala-output-version` and starting the LTS branch. After talking with many people, I'm convinced there is a lot of misunderstanding of our compatibility guarantees and what they mean in practice, so I've tried to add at least some info about that to this post. Let me know if it fits as it is now or if I should elaborate further on some aspects or reshuffle the text in any way.

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+We say that two compiler versions are source compatible if every single project that can be compiled with one version can also be compiled with the other, and the resulting programs behave the same way. We put a lot of effort into ensuring that source compatibility is preserved between the patch releases, and every infringement between minor versions is easy to fix. To enforce that, we are running builds of hundred of open-source Scala 3 projects every night. Moreover, after every release candidate, we are testing the compiler on practically all open-source projects that were ever published for Scala 3.
+
+Also, it is worth noting that braking source compatibility is usually not a big deal, as it can be easily detected at compile time. Fixing those problems in the past usually only involved specifying explicit types in some places where the programmer used type inference. If your project is affected by breakage in the source compatibility, you often need to make only small adjustments. In the worst-case scenario, you will be locked on one version of the compiler for some time. That, however, will not affect projects depending on your code.
```
bishabosha, line 25:
```suggestion
Also, it is worth noting that breaking source compatibility is usually not a big deal, as it can be easily detected at compile time. Fixing those problems in the past usually only involved specifying explicit types in some places where the programmer used type inference. If your project is affected by breakage in the source compatibility, you often need to make only small adjustments. In the worst-case scenario, you will be locked on one version of the compiler for some time. That, however, will not affect projects depending on your code.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Output compatibility (sometimes called *binary compatibility*, especially in Scala 2 contexts) is much more tricky and essential for our long-term plans for Scala. In the rest of this post, the term *compatibility* will always refer to output compatibility. We can say that compiler `A` is output compatible with compiler `B` when compiler `B` can use the output (binaries and TASTy files) generated by compiler `A` and understands it correctly.
```
bishabosha, line 27:
```suggestion
Output compatibility (sometimes called *binary compatibility*, especially in Scala 2 contexts) is much more tricky and essential for our long-term plans for Scala. In the rest of this post, the term *compatibility* will always refer to output compatibility. We can say that compiler `A` is output compatible with compiler `B` when compiler `B` can use the output (binaries and TASTy files) generated by compiler `A` and understand it correctly.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
+---
+layout: blog-detail
+post-type: blog
+by: Paweł Marks, VirtusLab
+title: Long-term compatibility plans for Scala 3
+---
+
+## Key takeaways
+
+- Scala 3 compiler that you are using to compile your code should be treated the same way as any library dependency in your project. You shouldn't be afraid of updating it. Bumping the patch version (e.g. 3.1.2 -> 3.1.3) is a no-brainer. Bumping the minor version (e.g. 3.1.3 -> 3.2.0) requires the same considerations as changing the minor version of any other dependency but is generally preferable.
```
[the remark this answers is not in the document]
sjrd: preferable to what? (preferable is a comparative that requires something to be compared to)

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+- You do not need to cross-compile between Scala 3 versions.
```
sjrd, line 11: I think it's stronger than "we don't need to". It's really "we shouldn't". (or at least we shouldn't cross-*publish*)
```suggestion
- You should not cross-compile between Scala 3 versions.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+- If you are maintaining a library, you should drop Scala 3.0. Also, you should upgrade to Scala 3.2 in your next minor release.
+- We are testing that the code passing the compilation with previous versions of the compiler will keep compiling in future releases unless the fact that the code was compiling was a bug in the compiler. We are using hundreds of real-life projects to test that.
```
sjrd, line 13:
```suggestion
- We are testing that the code passing the compilation with previous versions of the compiler will keep compiling in future releases, unless the fact that the code was compiling was a bug in the compiler. We are using hundreds of real-life projects to test that.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+- After evaluation, we have dropped the experimental `-scala-output-version` flag, as it turned out to be hard to maintain and potentially harmful to the ecosystem.
```
sjrd, line 14:
```suggestion
- After evaluation, we have dropped the experimental `-scala-output-version` flag, as it turned out to be too hard to maintain and potentially harmful to the ecosystem.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+- Soon, we will start releasing Long-Term Support versions of the compiler for the users that require enhanced stability.
+
+## About compatibility
+
+Compilers, as all software really, are constantly evolving. Since the initial release of Scala 3, there have been seven stable versions of the compiler. Each has brought performance and stability enhancements, fixed bugs, improved reporting or introduced some new experimental features to the language. Even though we still treat all those versions as an implementation of the same language. What allows us to think in such a way is the compatibility guarantees.
```
sjrd, line 19: "Even though" introduces a subordinate sentence. It can't be used as the main sentence. Consider
```suggestion
Compilers, as all software really, are constantly evolving. Since the initial release of Scala 3, there have been seven stable versions of the compiler. Each has brought performance and stability enhancements, fixed bugs, improved reporting or introduced some new experimental features to the language. Nevertheless, we still treat all those versions as an implementation of the same language. What allows us to think in such a way is the compatibility guarantees.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+The term *compatibility* is usually used quite vaguely. In the Scala world, the first intuition for this term could be phrased as follows: a new version of Scala is compatible with a previous version of Scala if a program that used to work with the earlier version still works with the latest version. Let me examine the term *compatibility* closer right now. For Scala 3 compiler, two things are named *compatibility*: source and output compatibility.
```
sjrd, line 21:
```suggestion
The term *compatibility* is often used quite vaguely. In the Scala world, the first intuition for this term could be phrased as follows: a new version of Scala is compatible with a previous version of Scala if a program that used to work with the earlier version still works with the latest version. Let us examine the term *compatibility* closer right now. For Scala 3 compiler, *compatibility* involves two concepts: source and output compatibility.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+We say that two compiler versions are source compatible if every single project that can be compiled with one version can also be compiled with the other, and the resulting programs behave the same way. We put a lot of effort into ensuring that source compatibility is preserved between the patch releases, and every infringement between minor versions is easy to fix. To enforce that, we are running builds of hundred of open-source Scala 3 projects every night. Moreover, after every release candidate, we are testing the compiler on practically all open-source projects that were ever published for Scala 3.
```
sjrd, line 23:
```suggestion
We say that two compiler versions are source compatible if every single project that can be compiled with one version can also be compiled with the other, and the resulting programs behave the same way. We put a lot of effort into ensuring that source compatibility is preserved between patch releases, and that every infringement between minor versions is easy to fix. To enforce that, we are running builds of hundreds of open-source Scala 3 projects every night. Moreover, after every release candidate, we are testing the compiler on practically all open-source projects that were ever published for Scala 3.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Also, it is worth noting that breaking source compatibility is usually not a big deal, as it can be easily detected at compile time. Fixing those problems in the past usually only involved specifying explicit types in some places where the programmer used type inference. If your project is affected by breakage in the source compatibility, you often need to make only small adjustments. In the worst-case scenario, you will be locked on one version of the compiler for some time. That, however, will not affect projects depending on your code.
```
sjrd, line 25:
```suggestion
In addition, breaking source compatibility is usually not a big deal, as it can easily be detected at compile-time. In the past, fixing those problems have usually been limited to specifying explicit types in some places where the programmer relied on type inference. If your project is affected by a source-incompatible change, you often need to make only small adjustments. In the worst-case scenario, you will be locked on one version of the compiler for some time. That, however, will not affect projects depending on your code.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Output compatibility (sometimes called *binary compatibility*, especially in Scala 2 contexts) is much more tricky and essential for our long-term plans for Scala. In the rest of this post, the term *compatibility* will always refer to output compatibility. We can say that compiler `A` is output compatible with compiler `B` when compiler `B` can use the output (binaries and TASTy files) generated by compiler `A` and understand it correctly.
```
sjrd, line 27:
```suggestion
Output compatibility (sometimes called *binary compatibility*, especially in Scala 2 contexts) is much trickier to understand, yet essential for our long-term plans for Scala. In the rest of this post, the term *compatibility* will always refer to output compatibility. We say that a compiler `A` is output compatible with compiler `B` when compiler `B` can use the output (binaries and TASTy files) generated by compiler `A` and understand it correctly.
```

sjrd: The last sentence here is nonsensical. It actually defines *forward* output compatibility, but describes it as a (supposedly) bidirectional notion of compatibility.

The following paragraph defines these things correctly. Consider dropping the last sentence here.

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+We can further subdivide compatibility to forward and backward. We say that the two versions are forward compatible when the older compiler can depend on the output of the newer one. Conversely, backward compatibility means that the newer compiler can use the output of the older version. Scala 3 guarantees backward compatibility between all releases and forward compatibility between patch releases in the same minor line. That means that Scala `3.a.x` can consume the output of Scala `3.b.y` only if `a` is greater or equal to `b`. For example, the output of the Scala 3.0.2 compiler can be used by Scala 3.0.1 or 3.1.0. On the other hand, code compiled with Scala 3.1.0 cannot be a dependency of any project compiled with 3.0.2.
```
sjrd, line 29: Consider swapping the role of a/b and x/y, as intuitively I expect b to come after a:
```suggestion
We can further subdivide compatibility to forward and backward. We say that the two versions are forward compatible when the older compiler can depend on the output of the newer one. Conversely, backward compatibility means that the newer compiler can use the output of the older version. Scala 3 guarantees backward compatibility between all releases and forward compatibility between patch releases in the same minor line. That means that Scala `3.b.y` can consume the output of Scala `3.a.x` only if `b` is greater or equal to `a`. For example, the output of the Scala 3.0.2 compiler can be used by Scala 3.0.1 or 3.1.0. On the other hand, code compiled with Scala 3.1.0 cannot be a dependency of any project compiled with 3.0.2.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+## What Scala should I use?
```
sjrd, line 31:
```suggestion
## What version of Scala should I use?
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+So, we know what output compatibility is and what Scala 3 is guaranteeing. But what does it means in practice? What version of the compiler should I choose for my project? The compiler team believes that the good answer for this question is always *the newest one*.
```
sjrd, line 33:
```suggestion
So, we know what output compatibility is and what Scala 3 guarantees. What does it mean in practice? What version of the compiler should I choose for my project? The compiler team believes that the good answer for this question is always *the newest one*.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+We believe that you should treat the compiler as one of your dependencies. If you are ok with updating the versions of your library dependencies, there should be no reason not to update the compiler version. In most cases, you wouldn't even notice that the Scala version has changed. Still, you can get better compilation performance, improved error messages, and most importantly, you may be able to use newly stabilized features of the language and the standard library.
```
sjrd, line 35:
```suggestion
We believe that you should treat the compiler as one of your dependencies. If you are fine with updating the versions of your library dependencies, there should be no reason not to update the compiler version. In most cases, you wouldn't even notice that the Scala version has changed. Meanwhile, you can get better compilation performance, improved error messages, and most importantly, you may be able to use newly stabilized features of the language and of the standard library.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+## The problem
+
+While this sounds good, there is one problem with intertwined ecosystems. Even though we do what we can to make the transition between minor versions of the compiler as smooth as possible, there will always be projects stuck on older versions for various reasons. That means they may be unable to update some of their dependencies if said libraries updated themselves to the newer compiler. That, in order, encourages libraries to stay on the oldest possible version of the compiler, which bars them from potentially beneficial improvements in the newer compilers.
+
+Let's examine the simple example. For some reason, we have a project that is stuck on Scala 3.1. We depend on version 1.3.5 of a library compiled with Scala 3.1. The library publishes version 1.4.0 with the compiler updated to 3.2. We cannot update our dependency because of output incompatibility. That also means that all patches for security issues won't be available unless the library author decides to backport them and release them as version 1.3.6. With that knowledge, the library author would be very reluctant to migrate from Scala 3.1 to 3.2.
```
sjrd, line 46:
```suggestion
For example, let us consider a project that is stuck on Scala 3.1, for some reason. We depend on version 1.3.5 of a library compiled with Scala 3.1. The library publishes version 1.4.0 with the compiler updated to 3.2. We cannot update our dependency because of output incompatibility. That also means that all patches for security issues won't be available unless the library author decides to backport them and release them as version 1.3.6. With that knowledge, the library author would be very reluctant to migrate from Scala 3.1 to 3.2.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Searching for a solution
+
+We have been discussing and testing various potential solutions in the past few months. Today we want to tell more about the attempts that failed and present the solution that we believe is the best for the future of Scala.
```
sjrd, line 50:
```suggestion
We have been discussing and testing various potential solutions in the past few months. Today we want to tell you more about the attempts that failed and present the solution that we believe is the best for the future of Scala.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+More importantly, during work on 3.2, we realized that maintaining this flag may be more challenging than anticipated. With time, it will be increasingly harder to be sure that our handling of it is correct. **Thus, we have decided to remove the possibility of configuring the output version altogether in Scala 3.2.**
+
+## Our solution: Scala LTS & Scala Next
+
+Finally, we have agreed that the best course of action would be to split Scala development into two lines, called *Scala LTS* and *Scala Next*. Both would live as separate branches in the compiler repository. Both will have separate but possibly synchronized releases. Porting changes from one to the other should be easy and common.
+
+Scala Next will be the line on which the language development is taking place. It will receive frequent minor updates, and all experimental language features will live here.
```
sjrd, line 77:
```suggestion
Scala Next will be the line on which the language development is taking place. It will receive frequent minor updates, and all experimental language features will live there.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Scala LTS, on the other hand, will be the stable long-term support line. One LTS will only receive patch updates which means that all releases of the same LTS line will be forward and backward compatible in terms of output. The only changes between releases will be bug fixes, non-language changes (doctool, semanticDB, reporting), and minor quality-of-life enhancements (only if we are sure we are not breaking any compatibility guarantees). After more than two years, we may decide to nominate one of the releases from the Next line as the new LTS. After that, we guarantee that the compiler team will support the previous LTS for at least one additional year. That means the support time for each LTS release is at least three years.
```
sjrd, line 79:
```suggestion
Scala LTS, on the other hand, will be the stable long-term support line. An LTS version will only receive patch updates, which means that all releases of the same LTS line will be forward and backward compatible in terms of output. The only changes between releases will be bug fixes, non-language changes (doctool, semanticDB, reporting), and minor quality-of-life enhancements (only if we are sure we are not breaking any compatibility guarantees). After more than two years, we may decide to nominate one of the releases from the Next line as the new LTS. After that, we guarantee that the compiler team will support the previous LTS for at least one additional year. That means the support time for each LTS release is at least three years.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Owners of commercial projects
+
+We suggest you stick to the LTS version. It will give you the best tooling support and a stable compiler. If you are migrating a bigger project from Scala 2, you should migrate it straight to the newest LTS.
+
+You may also choose to be up to date with the latest minor version. That will allow you to use new features as soon as they are added to the language. However, you will need to be ready for frequent compiler updates. Otherwise, your project may end up with an unsupported language version.
+
+### Library maintainers
+
+Unless your library is built around some new language feature available only on Next, you should stick to LTS to have the broadest possible user base. If your library contains multiple modules, some of them may require Scala Next. It is perfectly fine as long as you keep in mind that modules compiled with Scala LTS cannot depend on them (the other way around is ok). Also, users using your library from Scala LTS won't be able to access modules built with Next.
```
sjrd, line 101:
```suggestion
Unless your library is built around some new language feature available only on Next, you should stick to the latest LTS version to have the broadest possible user base. If your library contains multiple modules, some of them may require Scala Next. It is perfectly fine as long as you keep in mind that modules compiled with Scala LTS cannot depend on them (the other way around is fine). Also, users using your library from Scala LTS won't be able to access modules built with Next.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+Remember that when you bump a minor version of the compiler (e.g., moving from the previous Scala LTS line to the next), you also need to bump the minor version of your library.
```
sjrd, line 103:
```suggestion
Remember that when you bump the minor version of the compiler (e.g., moving from the previous Scala LTS line to the next), you also need to bump the minor version of your library.
```

blog/_posts/2022-08-05-long-term-compatibility-plans.md
Kordyjan: Is it better now?

blog/_posts/2022-08-05-long-term-compatibility-plans.md
```diff
@@ -0,0 +1,107 @@
...
+
+If you are a library author following the semantic versioning, we advise you to introduce new minor versions of the compiler only in minor releases of your library. New patch versions of the compiler can be introduced anytime, as they are forward and backward compatible.
+
+That may be surprising if you have experience developing libraries for Scala 2. For example, migrating any codebase from 2.11 to 2.12 was a huge undertaking. Since they were not output-compatible, most libraries needed to resort to cross-compilation to support both versions. Thanks to the guarantees of Scala 3, there is no longer a need for cross-compilation. For example, the migration from Scala 3.1 to Scala 3.2 for the large majority of the projects would only require changing the version in the build definition.
+// TODO: PUT HERE THE EXACT PERCENTAGE OF PROJECTS THAT WILL BE AFFECTED BY THE SOURCE INCOMPATIBILITIES BETWEEN 3.1 AND 3.2
```
SethTisue, line 40: something needs to happen here before publication :-)

Kordyjan: Yeah, I remember. I'm waiting for the exact results.

SethTisue (approved):

PR #1388, opened 2022-06-02
fix CI by removing dead link

SethTisue:


Philippus:
Maybe make a reference to https://docs.scala-lang.org/scala3/getting-started.html instead.
The IDE support page before it was deleted only contained a link to that page: https://github.com/lampepfl/dotty/blob/7c446ce8b2bf7031a6cf223b5f6a2782c1543016/docs/_docs/usage/ide-support.md.

PR #1391, opened 2022-06-08
add Scala 2.12.16

SethTisue:


_posts/2022-06-07-release-notes-2.12.16.md
```diff
@@ -0,0 +1,12 @@
+---
```
julienrf, line 1: Note that date is 7th of June. By the way, we also plan to publish a new post today (8th of June), see #1389, is this an issue?

PR #1409, opened 2022-08-17
tweak new blog post slightly

SethTisue:
review by @Kordyjan ?


Kordyjan (approved):
Thank you. It definitely reads better.

PR #1411, opened 2022-08-19
Added myself to teachers list (at USMA)

ryandougherty:


julienrf:
Thank you Ryan!

I think we need to update the model to show a list of teachers per point on the map. Currently, there is already Chris Okasaki at USMA.

I thought we could have a list of institutions, where every institution would have a location and a list of teachers. What do you think?

julienrf:
I have just updated the data source to include multiple teachers from the same institution in abe8e4050f47cd30c58c0fdd0989502dba92467f, and I added you in c99c4c259e1cfd2c15a5f90e877aad87df1752e0

PR #1415, opened 2022-09-08
podcasts: add Tooling Talks, remove Scala Love

SethTisue:
Scala Love hasn't done an episode since 2020

I considered also removing The Scala Logs since it's been about a year since the last episode, but maybe a year isn't so long


SethTisue:
review by @ckipp01 ?

ckipp01 (approved):
👍🏼 thanks for submitting this! LGTM from me.

PR #1419, opened 2022-09-23
update user group links

SethTisue:
scala.space seems dead, scala.meetup.com does too


julienrf (approved):
Sounds good to me, thank you.

Philippus (approved):

PR #1423, opened 2022-10-12
fix broken archival Scala Days links

SethTisue:
as noticed by failing CI

### DIFF
--- a/_events/2014-06-16-scaladays.md
+++ b/_events/2014-06-16-scaladays.md
@@ -6,5 +6,5 @@ location: Berlin, Germany
 description: ""
 start: 16 June 2014
 end: 18 June 2014
-link-out: https://scaladays.org/assets/archive/berlin2014/index.html
+link-out: https://scaladays.org/
 ---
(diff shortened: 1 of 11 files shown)


_events/2014-06-16-scaladays.md
```diff
@@ -6,5 +6,5 @@ location: Berlin, Germany
 description: ""
 start: 16 June 2014
 end: 18 June 2014
-link-out: https://scaladays.org/assets/archive/berlin2014/index.html
+link-out: https://scaladays.org/
```
eed3si9n, line 9: Did we lose historical Scala Days sites? If so could we bring them back?

SethTisue: hmm, not sure who to ask, maybe @MaureenElsberry ?

I think we should merge the PR regardless to fix CI, but I'd be happy to revert it again later if the archives were resurrected.

MaureenElsberry:
@SethTisue @eed3si9n - we will be bringing them back, but we only had time to create a splash page when Lightbend transferred the domain to us. Israel is working on additions soon. I think it's fine to archive it for now and I can let you know when the pages are back up!

eed3si9n (approved):

SethTisue:
thanks Maureen!

PR #1435, opened 2022-11-10
add versions back to front page

bishabosha:
I replaced the redundant "learn more" button with three buttons of current versions and all versions

Screenshot 2022-11-10 at 11 53 22

### DIFF
--- a/_sass/components/buttons.scss
+++ b/_sass/components/buttons.scss
@@ -3,7 +3,7 @@
 //------------------------------------------------
 .button {
     padding: 8px 18px;
-    font-size: $font-size-small;
+    font-size: $font-size-large;
     font-weight: $font-bold;
     text-transform: uppercase;
     color: #fff;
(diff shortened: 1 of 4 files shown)


bishabosha:
I thought about highlighting each version in different colours, but maybe its best to have the same

_sass/components/buttons.scss
```diff
@@ -3,7 +3,7 @@
 //------------------------------------------------
 .button {
     padding: 8px 18px;
-    font-size: $font-size-small;
+    font-size: $font-size-large;
```
julienrf, line 6: Isn’t that going to have an impact on a lot of other places?

julienrf (approved):
Great idea!

_sass/components/buttons.scss
bishabosha: no its only used as a class on the front page

julienrf (approved):

SethTisue:
Excellent!

PR #1450, opened 2022-12-16
add spacing between buttons on small screens

bishabosha:
before:
Screenshot 2022-12-16 at 17 20 42

after:
Screenshot 2022-12-16 at 17 20 13


julienrf (approved):
Thank you for the fix!

PR #1451, opened 2022-12-20
change "Get Started" button text

bishabosha:
Change to "Setup Scala"

people seem to be confused at the choice of "Get Started" and "Learn Scala", here's one option to try:

Screenshot 2022-12-20 at 12 58 12

the other option could be to change the text of "Learn Scala" to something else

### DIFF
--- a/index.md
+++ b/index.md
@@ -9,7 +9,7 @@ headerButtonUrl: "/what-is-scala/"
 
 # Links of the Download / API Docs sections
 gettingStarted:
-  mainTitle: "Get Started"
+  mainTitle: "Setup Scala"
   mainUrl: "https://docs.scala-lang.org/getting-started.html"
   subtitle: "API Docs"
   subtitleLink: "https://docs.scala-lang.org/api/all.html"


sjrd:
TBH I think "Get Started" was the better one, and "Learn Scala" should be changed instead. Perhaps to "Reference documentation" or something like that.

julienrf:
> people seem to be confused at the choice of "Get Started" and "Learn Scala"

Where? Do you have examples?

I am not sure about the proposed change. If someone needs to be changed, maybe “learn Scala” could be replaced with “documentation”?

bishabosha:
> Where? Do you have examples?

every time I ask someone in person what they think this happens to me :) but that is still anecdotal

bishabosha:
something like this?

Screenshot 2022-12-20 at 15 23 10

julienrf:
It would be good to ask the same people who were confused about the current structure.

SethTisue:
"setup" isn't a verb — it's “set up”

bishabosha:
superseded by #1518

PR #1463, opened 2023-01-31
Add announcement for 3.2.2

Kordyjan:
This post announces the 3.2.2 patch release that was made public yesterday.

As Scala 3 is becoming mature, and patch releases contain mostly uninteresting bug fixes, I thought it might be a good idea not to release a separate blog post every time there is a new patch but rather announce it in the same style as Scala 2.13 does.

Of course, we will still have bigger blog posts for minor releases.

### DIFF
--- a/_data/scala-releases.yml
+++ b/_data/scala-releases.yml
@@ -1,7 +1,7 @@
 - category: current_version
   title: Current 3.2.x release
-  version: 3.2.1
-  release_date: November 7, 2022
+  version: 3.2.2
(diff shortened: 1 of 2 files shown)


Kordyjan:
@julienrf? @SethTisue? Can you take a look?

julienrf (approved):

SethTisue:
You can probably guess that I like the way we do it in Scala 2. So yes, I like it for Scala 3 too :-)

PR #1464, opened 2023-02-01
don't recommend the obsolete dotty homebrew formula anymore

SethTisue:
a Scala 2 user on Discord, looking to move to Scala 3, was bitten by this

### DIFF
--- a/_includes/downloads-scala3.html
+++ b/_includes/downloads-scala3.html
@@ -29,7 +29,7 @@ For a summary of important changes, see the <a
         <code>sdk install scala</code>
       </li>
       <li>On macOS you can also use <a href="https://brew.sh/">Homebrew</a> and run the following commands:<br>
-        <code>brew update</code><br /><code>brew install lampepfl/brew/dotty</code>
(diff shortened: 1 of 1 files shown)


bishabosha (approved):
We could also update the scala 2 download page to use scala@2.13 but this would need more logic otherwise it will be invalid for 2.12 🥲

SethTisue:
Yeah :-(

_includes/downloads-scala3.html
```diff
@@ -29,7 +29,7 @@ <h2>Ways to Install This Release</h2>
         <code>sdk install scala</code>
       </li>
       <li>On macOS you can also use <a href="https://brew.sh/">Homebrew</a> and run the following commands:<br>
-        <code>brew update</code><br /><code>brew install lampepfl/brew/dotty</code>
+        <code>brew update</code><br /><code>brew install scala</code>
```
julienrf, line 32: Where is this formula defined?

bishabosha: It's in homebrew/homebrew-core

PR #1469, opened 2023-02-14
Fix setup instructions

julienrf:
Consistent with https://github.com/scala/docs.scala-lang/pull/2698

### DIFF
--- a/_install_tabs/1-macos.html
+++ b/_install_tabs/1-macos.html
@@ -6,7 +6,7 @@ tabLabel: macOS
   <div class="wrap">
     <p>Run the following command in your terminal, following the on-screen instructions:</p>
     {% include code-snippet.html language='bash' codeSnippet=site.data.setup-scala.macOS-brew %}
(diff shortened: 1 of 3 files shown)


_install_tabs/1-macos.html
```diff
@@ -9,7 +9,10 @@
     {% assign homebrewAlt = "Alternatively, if you don't use Homebrew:" %}
```
bishabosha, line 9:
```suggestion
    {% assign homebrewAlt = "Alternatively for M1 Macs, or if you don't use Homebrew:" %}
```

bishabosha (approved):
generally I approve, but the home-brew link installs the x86 version so I think the alt-details label text should reflect this

julienrf:
Ah, good point. But I wonder if we could fix the homebrew formula instead to download the correct binary.

bishabosha:
> Ah, good point. But I wonder if we could fix the homebrew formula instead to download the correct binary.

then at that point, revert the title :)

PR #1497, opened 2023-04-29
Install page - Remove Scastie references.

diesalbla:
Scastie is not a way for _installing_ or downloading Scala into a local machine, as the title of the page says. There are links to Scastie directly from main page.

Regarding Scalafiddle, the link https://scalafiddle.io/  seem not to be working. If it is another playground like Scastie, should it be linked from the scala-lang page?


julienrf (approved):
Good point, thank you for the PR

bishabosha (approved):

PR #1505, opened 2023-05-23
Add 3.3.0 announcement

Kordyjan:


blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+class TestSuite extends munit.FunSuite:
+  test("the job doesn't fail"):
+    val job = canFail("some data")
+    assert(job.isSuccess)
+```
+
+### `boundary` and `break`
+
+Two new methods were added to the standard library in the place: `boundary` and `brake`. They are safer and more expressive replacements for recently deprecated non-local returns.
```
bishabosha, line 75:
```suggestion
Two new methods were added to the standard library in the place: `boundary` and `break`. They are safer and more expressive replacements for recently deprecated non-local returns.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
+---
+layout: blog-detail
+post-type: blog
+by: Paweł Marks, VirtusLab
+title: Scala 3.3.0 released!
+---
+
+![Scala 3.3]({{ site.baseurl }}/resources/img/scala-3.3-launch.png)
```
julienrf, line 8: The image is missing.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+LTS might be a great choice for library authors who can now receive constant bug fixes and developer experience improvements without forcing the users of the library to update the compiler version in their project.
+
+You can read more about our compatibility guarantees in [the recent blog post](https://virtuslab.com/blog/the-scala-3-compatibility-story/).
+
+## What's new in Scala 3.3.0
+
+### Linting
+
+Scala 3.3.0 brings back the linting to the Scala compiler.  Right now, you can enable checking for unused symbols and discarded values. More linting options will come soon in the following Scala 3.3.x releases.
```
julienrf, line 28: I am not sure about “the linting”. Should we just say “linting”?

Also, I think it would be good to mention that our goal with this work is to ease the migration to Scala 3.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+In the following simplified example
+
+```Scala
+def failure: Either[String, Nothing] = Left("something broke")
+
+def failedComputation: Either[String, Unit] =
+  Right(()).map(_ => failure)
+```
+
+the programmer, by mistake, used the `map` method instead of `flatMap`. The code still compiles, but due to value discarding, behaves unexpectedly, returning `Right(())` from the `failedComputation` method. If the `-Wvalue-discard` flag is enabled, the compiler will report the warning, saving the user from a potential bug.
```
julienrf, line 55: Would it be possible to show the compiler output in such a case?

blog/_posts/2023-05-26-scala-3.3.0-released.md
julienrf: It seems something is missing after “in the place”, did you mean “in place of `scala.util.control.Breaks` and `scala.util.control.NonLocalReturns`”?

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+def sumOfRoots(number: List[Double]): Option[Double] = boundary {
+  val roots = numbers.map { n =>
+    println(s" * calculating square root for $n*")
+    if n >= 0 then Math.sqrt(n) else break(None)
+  }
+  Some(roots.sum)
+}
+```
+
+When you run the above method, you will notice that it returns None when any of the input elements is negative. Moreover, thanks to `boundary`/`break`, you can see from the console output that it stops the entire process after encountering the first negative element.
```
julienrf, line 91:
```suggestion
When you run the above method, you will notice that it returns `None` when any of the input elements is negative. Moreover, thanks to `boundary`/`break`, you can see from the console output that it stops the entire process after encountering the first negative element.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+`break` can jump out of the`boundary` in the function deeper on the stack. To make it safe, only functions with a matching `Label` in their using clauses can break. This can be used to create isolated parts of an application with streamlined error handling.
```
julienrf, line 93:
```suggestion
`break` can jump out of the `boundary` in the function deeper on the stack. To make it safe, only functions with a matching `Label` in their using clauses can break. This can be used to create isolated parts of an application with streamlined error handling.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+### `boundary` and `break`
+
+Two new methods were added to the standard library in the place: `boundary` and `brake`. They are safer and more expressive replacements for recently deprecated non-local returns.
+
+`break` allows for a type-safe early escape from anywhere inside the block delimited by `boundary` to its end, returning the passed value from the entire block.
+
+```Scala
+import util.boundary, boundary.break
+
+def sumOfRoots(number: List[Double]): Option[Double] = boundary {
```
julienrf, line 82: Should we use the braceless syntax?

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+  boundary {
+    val user = getUser(id)
+    val address = getAddress(user)
+    Right((user, address))
+  }
+```
+
+### The new default implementation of lazy vals
+
+Last but not least, we have changed the default implementation of lazy vals. The new implementation has better performance and is safer under parallel access. This can lead to improvements in the effect systems performance.
```
julienrf, line 128:
```suggestion
Last but not least, we have changed the default implementation of lazy vals. The new implementation has better performance and is safer under parallel access. This can lead to performance improvements in the effect systems.
```

(but which effect systems do we talk about?)

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Should I update to Scala 3.3.0?
+
+If you are maintaining a standalone Scala 3 project without external projects depending on it, feel free to switch to Scala 3.3.0 at any moment. You will enjoy all the improvements in the newest version of the compiler.
+
+If you are a library author following semantic versioning, we advise you to update to Scala 3.3.0 in the next minor release of your project. Users of your library would also need to bump their compiler version to use the newest version of your library. If you have already published a minor version on 3.3.0 but learned about a critical vulnerability in your library, you can release a new patch release for the previous minor version of your library using Scala 3.2.2 (or any other version of the compiler you were using previously). This patch can be consumed by all the users of previous versions of the library, no matter if they have already switched to 3.3.0 or not.
```
julienrf, line 134:
```suggestion
If you are a library author following semantic versioning, we advise you to update to Scala 3.3.0 in the next minor release of your project. Users of your library would also need to bump their compiler version to use the newest version of your library. If you have already published a minor version on 3.3.0 but learned about a critical vulnerability in your library, you should release a new patch release for the previous minor version of your library using Scala 3.2.2 (or any other older version of the compiler). This patch can be consumed by all the users of previous versions of the library, no matter if they have already switched to 3.3.0 or not.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+def method(param: Outer): param.Inner = ???  // ok
+```
+
+There was a bug that resulted in the compiler assuming that all inline parameters are stable references. This could have led to unsound code being accepted by the compiler and potential runtime crashes.
+
+```Scala
+inline def method(inline param: Outer): param.Inner // error in Scala 3.3.0
+```
+
+is now not accepted by the compiler. Migration for that change is simple, as it only requires removing the `inline` modifier **from the parameter**. If the path-dependent type was used, it is safe to say that the intended behavior was to inline the method without inlining the parameter.
```
julienrf, line 162:
```suggestion
Such pieces of code are now rejected by the compiler. The migration for that change is simple, as it only requires removing the `inline` modifier **from the parameter**. If the path-dependent type was used, it is safe to say that the intended behavior was to inline the method without inlining the parameter.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue: > Should we just say “linting”?

Yes

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue, line 75:
```suggestion
Two new methods were added to the standard library in the place: `boundary` and `brake`. They are safer and more expressive replacements for non-local returns, which were deprecated recently.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue: yes please

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue, line 91:
```suggestion
When you run the above method, you will notice that it returns `None` when any of the input elements is negative. Moreover, thanks to `boundary`/`break`, you can see from the console output that it stops iterating after encountering the first negative element.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue: just "effect systems", not "the effect systems"

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+```Scala
+def failure: Either[String, Nothing] = Left("something broke")
+
+def failedComputation: Either[String, Unit] =
+  Right(()).map(_ => failure)
+```
+
+the programmer, by mistake, used the `map` method instead of `flatMap`. The code still compiles, but due to value discarding, behaves unexpectedly, returning `Right(())` from the `failedComputation` method. If the `-Wvalue-discard` flag is enabled, the compiler will report the warning, saving the user from a potential bug.
+
+### More consistent brace-less syntax
```
SethTisue, line 57:
```suggestion
### More consistent braceless syntax
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
som-snytt: Amelia Bedelia says to call it "delinting". It's like dusting the furniture when you intended to undust it.

blog/_posts/2023-05-26-scala-3.3.0-released.md
som-snytt: This may result in improvements in the performance of effect systems.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+![Scala 3.3]({{ site.baseurl }}/resources/img/scala-3.3-launch.png)
+
+We are thrilled to announce that after long months of work and careful testing, we have released Scala 3.3.0 LTS.
+
+## What does "LTS" mean?
+
+Scala 3.3.0 is the first Long Term Support (LTS) release. That means that it will be actively maintained for a period of at least three years. We have adopted a release model similar to the one that Java has successfully used for a long time.
+
+In the following years, there will be new minor releases (3.4, 3.5, and so on) that can bring new backward-compatible features. We are calling them Scala Next; they are equivalent to Java's feature releases. Bugfixes and usability improvements from those releases will be backported and released as 3.3.x patches. Apart from those forward and backward-compatible changes, LTS will be feature frozen.
```
adpi2, line 16:
```suggestion
In the following years, there will be new minor releases (3.4, 3.5, and so on) that can bring new backward-compatible features. We are calling them Scala Next; they are equivalent to Java's feature releases. Bug fixes and usability improvements from those releases will be back-ported and released as 3.3.x patches. Apart from those forward and backward-compatible changes, LTS will be feature frozen.
```

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+The LTS model doesn't change anything in our compatibility guarantees. Projects built with all future releases will be able to depend on any library compiled with Scala 3.3.x. This is just our standard guarantee that the newer compiler can always consume the output of the older version.
+
+LTS might be a great choice for library authors who can now receive constant bug fixes and developer experience improvements without forcing the users of the library to update the compiler version in their project.
+
+You can read more about our compatibility guarantees in [the recent blog post](https://virtuslab.com/blog/the-scala-3-compatibility-story/).
+
+## What's new in Scala 3.3.0
```
sjrd, line 24: Consider using the following explicit subsections for this section:

1. Language changes
2. Developer experience improvements
3. Standard library additions

We must make extra clear what is language and what is not language, as it confuses users too often.

blog/_posts/2023-05-26-scala-3.3.0-released.md
sjrd, line 93: From here onwards, IMO the content exceeds what is appropriate for release notes. It instead belongs to the documentation and/or a tutorial/blog post about `boundary/break`.

As it stands, these release notes contain a comparatively large amount of text for something that is way less important than language changes and linting.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+def getShippingData(id: String): Either[String, (User, Address)] =
+  boundary {
+    val user = getUser(id)
+    val address = getAddress(user)
+    Right((user, address))
+  }
+```
+
+### The new default implementation of lazy vals
```
sjrd, line 126: This should come earlier, as it is the only thing that may potentially have an impact on existing codebases during the upgrade process. Everything else is things you can profit from *after* you've upgraded.

In the Scala.js release notes, we have an explicit "Changes with compatibility concerns" section before a "What's new" section. See for example https://www.scala-js.org/news/2023/01/26/announcing-scalajs-1.13.0/. Consider doing the same here.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Should I update to Scala 3.3.0?
+
+If you are maintaining a standalone Scala 3 project without external projects depending on it, feel free to switch to Scala 3.3.0 at any moment. You will enjoy all the improvements in the newest version of the compiler.
+
+If you are a library author following semantic versioning, we advise you to update to Scala 3.3.0 in the next minor release of your project. Users of your library would also need to bump their compiler version to use the newest version of your library. If you have already published a minor version on 3.3.0 but learned about a critical vulnerability in your library, you can release a new patch release for the previous minor version of your library using Scala 3.2.2 (or any other version of the compiler you were using previously). This patch can be consumed by all the users of previous versions of the library, no matter if they have already switched to 3.3.0 or not.
+
+### Known incompatibility: Stability of inline parameters
```
sjrd, line 136: This is also for the "Changes with compatibility concerns" section.

You probably want to make it more explicit that this a bug fix (so OK to do in a minor version), but for which we have already noticed existing problems in the wild, and so we're helping users by proactively giving them more information on how to deal with it.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,162 @@
...
+
+In Scala, only stable paths can be used as a prefix in the path-dependent type.
```
sjrd, line 138:
```suggestion
In Scala, only stable paths can be used as prefix in path-dependent types.
```

sjrd (changes requested):

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,159 @@
...
+
+def failedComputation: Either[String, Unit] =
+  Right(()).map(_ => failure)
+```
+
+the programmer, by mistake, used the `map` method instead of `flatMap`. The code still compiles, but due to value discarding, behaves unexpectedly, returning `Right(())` from the `failedComputation` method. If the `-Wvalue-discard` flag is enabled, the compiler will report the warning, saving the user from a potential bug.
+
+### More consistent braceless syntax
+
+Braces around method parameters can now be replaced with a colon. This can lead to cleaner, shorter, and often more readable code in places like configuration DSLs or test case definitions.
```
eed3si9n, line 59: Given the controversial nature of Optional Brace, and now More Braceless syntax, it might be beneficial for both the maintainers and the readers to expand the context around this feature a bit, namely who proposed it, when, why, and how it was discussed and vetted (for example it was approved by SIP process in August 2022).

Kordyjan:
I've applied all the language suggestions. Now I'm thinking about the structural changes suggested by @sjrd. if I understand them correctly, we would end up with a structure like this:

```
- What does "LTS" mean?
- Changes in Scala 3.3.0
  ├── Changes with compatibility concerns
  │   ├── New lazy vals
  │   └── Stability of inline parameters
  ├── Language changes
  │   └── More consistent braceless syntax
  ├── Standard library changes
  │   └── `boundary` and `break`
  └── Usability Improvements
      └── Linting
          ├── Checking for unused values
          └── Checking for discarded values
- Should I update to Scala 3.3.0
```

I see a bunch of problems:
-  we are starting with things affecting only a very minor fraction of the users. New lazy vals are mostly implementation detail. The stability of inline parameters affects only a few libraries from the Open CB and probably even fewer end-users.
- The most important and most asked-for feature of the release - linting - is near the end of the text, and some readers may even miss it completely, especially since the sidebar is showing only two levels of headers.
- I think putting "new lazy vals" in the same category as known source incompatibility will be confusing. One of them always requires changes in the source code of affected projects, and the other should be practically transparent for the user and not require any action except for extraordinary cases that can be considered bugs.
- There is a split between  "Should I update..." and "Known incompatibilities", which should be grouped together as they are targeting the same group of readers, and reading both of them are required to have a full picture of migration.

I'm ok with removing the `boundary`/`break` example.

julienrf:
I think Sébastien’s suggestion makes sense. You could start with a section “Release Highlights” (just before “Changes with compatibility concerns”) with a summary of the high-impact changes for the end-users (like [sbt](https://eed3si9n.com/sbt-1.8.0) does).

blog/_posts/2023-05-26-scala-3.3.0-released.md
SethTisue: > IMO the content exceeds what is appropriate for release notes. It instead belongs to the documentation

fwiw, I was thinking similar thoughts and pondering whether to say something.

sjrd:
> * especially since the sidebar is showing only two levels of headers.

To address that specific concern, you can remove the heading "Changes in 3.3.0" and move everything under it one level up. It's the release notes for 3.3.0, so we already know that "changes" are about 3.3.0.

blog/_posts/2023-05-26-scala-3.3.0-released.md
```diff
@@ -0,0 +1,129 @@
...
+def method(param: Outer): param.Inner = ???  // ok
+```
+
+There was a bug that resulted in the compiler assuming that all inline parameters are stable references. This could have led to unsound code being accepted by the compiler and potential runtime crashes.
+
+```Scala
+inline def method(inline param: Outer): param.Inner // error in Scala 3.3.0
+```
+
+Such pieces of code are now rejected by the compiler. The migration for that change is simple, as it only requires removing the `inline` modifier **from the parameter**. If the path-dependent type was used, it is safe to say that the intended behavior was to inline the method without inlining the parameter.
```
prolativ, line 129: TODO: Contributors list

blog/_posts/2023-05-26-scala-3.3.0-released.md
tgodzik: On one hand it will be useful to have a list of things that might cause issues when updating, but maybe we can add it as a later section and let people know at the start?

"For compatibility concerns look to..."

This way we can first focus on things that are most interesting to users?

Also if lazy vals are an internal thing, so maybe it should on be in the same section? Or have a section with `Impactful changes for x.y.z release`? If the new lazy vals are not working then it's a bug and if we say that it's ` compatibility concerns` that might cause people to think that something changed that might make their code stop compiling, while it's actually very unlikely.

blog/_posts/2023-05-26-scala-3.3.0-released.md
tgodzik: We should make things clear what is a language feature for sure, but maybe we can have some tags instead? The sections might comprise of a single point, so that seems very formal for the sake of it being formal.

sjrd: I think having a section with a single point is good. It makes it very obvious that there is actually very little that changes in the language (contrary to what we can read left and right).

blog/_posts/2023-05-26-scala-3.3.0-released.md
sjrd: Yes, sure. At the end of the day, it's a matter of opinion. If you prefer to show the exciting stuff first given those considerations, go for it. ;)

blog/_posts/2023-05-30-scala-3.3.0-released.md
```diff
@@ -0,0 +1,185 @@
+---
+layout: blog-detail
+post-type: blog
+by: Paweł Marks, VirtusLab
+title: Scala 3.3.0 released!
+---
+
+![Scala 3.3]({{ site.baseurl }}/resources/img/scala-3.3_LTS-launch.png)
```
julienrf, line 8: Do you mind posting a screenshot of the rendering, please?

julienrf (approved):

Kordyjan:
image

Kordyjan:
image

Kordyjan:
Wow! Those screenshots are huge :D 
Sorry, I captured them automatically and didn't realise they are that long.

blog/_posts/2023-05-30-scala-3.3.0-released.md
```diff
@@ -70,6 +70,8 @@ class TestSuite extends munit.FunSuite:
     assert(job.isSuccess)
 ```
```
bishabosha, line 72: is it worth saying here that is can be disabled with `-source:3.2` / `import scala.language.3.2`?

PR #1506, opened 2023-05-30
Variable name mismatch in 2023-05-30-scala-3.3.0-released.md

ianlintner:
Fixed variable mismatch in example.

See https://scastie.scala-lang.org/PQCOjE0xTp6o1a3t2fm0cg


julienrf (approved):
Good catch, thank you!

PR #1524, opened 2023-07-13
index.md: grammar fix

nafg:
It might be accepted internet grammar but it isn't correct


nafg:
Hmmm, isn't there a redundancy between "some" and "include"? Maybe it should either be "some of the most popular domains are" or "popular domains include."

Also, I think the colon at the end is an older style, but I could be wrong

bishabosha:
>  or "popular domains include."

Thanks for the suggestion!

PR #1537, opened 2023-08-20
Don't call out Apple Silicon as needing manual installation

erbridge:
I have Apple Silicon and installed Scala fine for the first time via Homebrew.

See also https://github.com/scala/docs.scala-lang/pull/2889

### DIFF
--- a/_install_tabs/1-macos.html
+++ b/_install_tabs/1-macos.html
@@ -6,7 +6,7 @@ tabLabel: macOS
   <div class="wrap">
     <p>Run the following command in your terminal, following the on-screen instructions:</p>
     {% include code-snippet.html language='bash' codeSnippet=site.data.setup-scala.macOS-brew %}
-    {% assign homebrewAlt = "Alternatively for Apple Silicon, or if you don't use Homebrew:" %}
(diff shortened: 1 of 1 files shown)


bishabosha:
the homebrew formula installs the x86 version, so its under rosetta 2, for the best experience the curl download wins

SethTisue:
> the homebrew formula installs the x86-64 version

hmm. is there a ticket somewhere about changing that...?

bishabosha:
> > the homebrew formula installs the x86-64 version
> 
> hmm. is there a ticket somewhere about changing that...?

kind of https://github.com/coursier/coursier/issues/2770

SethTisue:
Closing on the assumption this isn't applicable anymore?

bishabosha:
If we changed the instructions to be the homebrew-core formula, rather than the coursier tap, then it will install a jar-only coursier, then `coursier setup` would install the native cs for apple silicon

SethTisue:
okay. well, I'm not going to dig deeper into this personally

michaelbilow:
I think this is related to #2999, coursier is complaining about lack of x86 machine instructions on my M1 mac now.

SethTisue:
@michaelbilow #2999 in what repo?

PR #1571, opened 2023-10-31
fix 404 reactconf link in past events page

SethTisue:
the old URL has been 404ing for a while now in our CI runs, e.g. https://github.com/scala/scala-lang/actions/runs/6700196688/job/18205723071#step:6:22

note that the old React conference is not to be confused with React.js


Philippus (commented):
We could also link to the videos instead: https://www.youtube.com/@instilco/videos

PR #1600, opened 2024-03-01
Update links to use scala/scala3 instead of lampepfl/dotty

hamzaremmal:
Part of https://github.com/lampepfl/dotty/pull/19839 and https://github.com/lampepfl/dotty/issues/19848


hamzaremmal:
@bishabosha This PR too :-)

PR #1616, opened 2024-03-26
Add David A Gil

dagmendez:
Add myself as ambassador.

I want to verify with you first if `linkedin` and  `meetup` work the same as `twitter`, `mastodon` or  `email`

### DIFF
--- a/_data/ambassadors.yml
+++ b/_data/ambassadors.yml
@@ -10,7 +10,7 @@
   description: Senior Scala Engineer (Scala 2 & 3), Organizer of AI&DG, Speaker. I'm "dagmendez" on the <a href="https://discord.com/invite/scala">Scala Discord</a>.
   image: https://avatars.githubusercontent.com/u/103842788
   city: Santiago de Compostela, Galicia, Spain
-  github: https://github.com/dagmendez
+  github: dagmendez
   linkedin: https://www.linkedin.com/groups/9545608/
   meetup: https://www.meetup.com/es-ES/big-data-e-inteligencia-artificial-galicia/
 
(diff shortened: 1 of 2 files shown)


SethTisue:
Oh good call on `linkedin` and `meetup`, I'll try to add support for those before merging this.

Not including GitHub was actually intentional. I think someone argued that GitHub didn't count as a means of contacting an ambassador, since GitHub doesn't have messaging.

But I thought about it and some more, and:

* Someone's GitHub page might have further contact information on it
* Although GitHub doesn't have messaging, it's still useful to know someone's GitHub handle
  * in order to @-mention them
  * or just to see what they're up to

So I think we should allow it, but we should just check when adding someone that the overall amount of contact information is sufficient.

I'll wait to merge this until I see if we can add support for `linkedin` and `meetup`.

I also want to see if we can do something for `discord` — I think it's not possible to link directly to someone's Discord account, but maybe we can finesse it somehow.

dagmendez:
I added `github` because even if it does not have messages, you can access the most active / most recent projects of that ambassador. For example, in the case of `Zainab`, we would see:

Screenshot 2024-03-27 at 10 15 16

As for my own `github`, I have the hacky landing page that is actually a repository. There I have the links to my blog posts and `linkedin` account. 

Screenshot 2024-03-27 at 10 21 37

On regard of `meetup`, it is the service I use for now, once I change to other, I will open a new PR here to update it. In that sense, is it worth it to add support?

SethTisue:
The Discord question is now an issue: https://github.com/scalacenter/ambassadors/issues/6

SethTisue:
`discord` too, now.

PR #1624, opened 2024-03-28
ambassadors page: add Zainab Ali

SethTisue:
@zainab-ali I hear you're busy at a conference today so I drafted for this you, is it okay?


zainab-ali:
Thanks @SethTisue !  I've opened #1629 with a more recent photo.

PR #1632, opened 2024-03-28
Add Jakub Kozłowski as an ambassador

kubukoz:
Thank you for the invitation :)

Also a couple related improvements:

- @zainab-ali photo looks strange as the browser tries to squash it into a square, I cropped it so it should display nicely now
- @zainab-ali's and @SethTisue's photos were a few MBs each, I resized them to 800x800 and exported to JPG at 90% quality.


SethTisue:
above and beyond, thank you!

PR #1634, opened 2024-03-29
Add 3.4.1 announcement

Kordyjan:


jchyb (approved):

Philippus (commented):
I think usually a list of contributors are added as well.

Kordyjan:
> I think usually a list of contributors are added as well.

Scala 2.13 and 2.12 don't do that for patch releases. There is a mention that contributors' credits can be found in the release notes. If this is a problem, I can add the contributors list.

Philippus:
No problem, just noticed it.

PR #1635, opened 2024-03-29
Add Nicolas to ambassadors

nrinaudo:
As requested by @SethTisue [here](https://fosstodon.org/@SethTisue/112178841696642149).


SethTisue:
another recruit at https://www.scalar-conf.com 👍

PR #1649, opened 2024-05-13
Add JetBrains as supporter.

sjrd:
I can't seem to build locally anymore, so hopefully this renders well.


SethTisue:
The hardest problems in computer science are caching, off-by-one errors, and rendering Jekyll sites locally.

PR #1661, opened 2024-06-19
Add a security policy under /security

lrytz:


security.md
```diff
@@ -0,0 +1,26 @@
...
+To set up email notifications for new security announcements, read [this post](https://users.scala-lang.org/t/about-the-security-announcements-category).
+
+## Reporting Vulnerabilities
+
+We strongly encourage reporting security issues in Scala to our private mailing list before disclosing them in public.
+
+The email address for security related communication is `security@scala-lang.org`.
+Messages are delivered to the Scala Security Team, which includes people from EPFL, the Scala Center, VirtusLab and Lightbend.
+
+We strive to acknowledge reports within 24 hours.
```
lrytz, line 22: Extend this, e.g. to 3 days?

security.md
```diff
@@ -0,0 +1,26 @@
...
+In case you don't receive a reply within a few days and would like to escalate, our advice is to ask for a contact person in a forum hosted by the Scala organization:
+  - [Meta category on Discourse](https://users.scala-lang.org/c/meta)
+  - [`#admin` channel on Discord](https://discord.com/channels/632150470000902164/632628729029328947) ([invite link](https://discord.com/invite/scala) for joining)
```
lrytz, line 25: Other options:
  - include a few direct contacts
  - open a ticket on scala/scala or scala/scala3

On a public forum, chat or also on a ticket, random people could jump in and give bad (or even malicious) advice.

security.md
sjrd: More importantly: business days ;)
May 2 business days?

security.md
sjrd: I think the current options are fine. This shouldn't happen anyway, since the mailing list should be enough.

community/index.md
```diff
@@ -31,6 +31,8 @@ The Scala Center focuses on education (especially online courses),
 documentation, open source community outreach, and tooling.  Community
 participation in all of these efforts is strongly encouraged.
 
+To receive security announcements or contact us about security issues, see our [security policy](/security/).
```
lrytz, line 34: The "Governance" section above would be a better place to add this (see https://www.scala-lang.org/community/), but there's currently no empty slot. Maybe "tooling summit" can move out...?

security.md
lrytz: I guess spam filtering is the possible concern here

SethTisue: regardless, I agree with Seb, this is already a good list of options

community/index.md
SethTisue: Hmm... no super strong opinion here, but I think I would suggest adding a Security section near the bottom of the page. I don't think it really fits in Governance.

security.md
```diff
@@ -0,0 +1,26 @@
...
+## Receiving Security Announcements
+
+Security announcements related to Scala are published to the ["Security Announcements" channel](https://users.scala-lang.org/c/security) on our discourse forum.
+
+Messages to this channel can only be posted by administrators, so it is very low traffic.
+To set up email notifications for new security announcements, read [this post](https://users.scala-lang.org/t/about-the-security-announcements-category).
+
+## Reporting Vulnerabilities
+
+We strongly encourage reporting security issues in Scala to our private mailing list before disclosing them in public.
```
SethTisue, line 17:
```suggestion
We strongly encourage reporting security issues in Scala to us privately before disclosing them in public.
```

SethTisue (commented):
This looks good and mergeable to me. But I'm wondering if you looked at the security pages for any other languages or technologies, and does this seem adequate to you based on that, modulo the fairly small size of our organization? (I think you might have said in conversation that you did look?)

Perhaps a security-savvy person at Lightbend could have a quick look (Michael N)?

lrytz:
> I'm wondering if you looked at the security pages for any other languages or technologies, and does this seem adequate to you based on that, modulo the fairly small size of our organization?

Yeah, I looked at rust, python, go, akka mainly. I kept it lightweight.

PR #1670, opened 2024-07-10
Add version to sdkman installation command

### LINKED ISSUE #1560: update sdk install instructions to include the actual version
On this page https://www.scala-lang.org/download/3.3.1.html
it says: `sdk install scala`

But I need to do
```
sdk install scala 3.3.1
sdk default scala 3.3.1
```
if I don't have a fresh install, but already have e.g. 3.3.0 installed.

If I do like in the instructions it for some reason downloads very old 3.1.1 and ask me to set that as default even if I also have 3.3.0 installed as (previous) default.

I'm not sure how sdkman thinks about the latest version of scala and the only safe thing to really get the actual version is to state it explicitly, hence I think the download page for each new version should also add the specific version after `sdk install scala`.

hamzaremmal:
Closes #1560


eed3si9n (approved):
@hamzaremmal Thanks!

PR #1674, opened 2024-07-31
fallback to 'monospace' for code font-family

cjbayliss:
Some people like myself disable custom fonts on the web. The reasons are often related to accesibility. In my case it is dyslexia, but there are probably other vision related reasons too.

### DIFF
--- a/_sass/components/code.scss
+++ b/_sass/components/code.scss
@@ -174,7 +174,7 @@
     code {
         // style that would be injected by hljs when JS is enabled
         display: block;
-        font-family: 'Consolas';
+        font-family: 'Consolas', monospace;
         border: none;
         display: block;
         overflow-x: auto;
--- a/_sass/layout/type-md.scss
+++ b/_sass/layout/type-md.scss
@@ -151,7 +151,7 @@
     li,
     p {
         code {
-            font-family: 'Consolas';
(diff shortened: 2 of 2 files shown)


SethTisue:
looks plausible, but I don't know CSS myself — were you able to test it?

cjbayliss:
I did test it, however there is no need to take my word for it. MDM has a very easy to follow document on `font-family`: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

Two relevant quotes:

"The `font-family` property specifies a list of fonts, from highest priority to lowest."

and:

"You should always include at least one generic family name in a font-family list, since there's no guarantee that any given font is available. This lets the browser select an acceptable fallback font when necessary."

I'm not a CSS person, I'm not even a front-end dev. I love Mozilla's MDM, and can highly recommend it whenever you need info on a CSS property or anything related to Javascript.

Another very useful resource if you need to check if a PR is safe to merge is caniuse.com: https://caniuse.com/mdn-css_properties_font-family

These two resources make it so that people like us who don't know CSS well can easily contribute and review CSS changes.

Sorry for the tangent. 😅 I hope it is helpful for future reference :)

SethTisue (approved):

SethTisue:
Thank you!

PR #1675, opened 2024-08-01
Blog post about given prioritization changes

bracevac:


_posts/2024-08-01-given-priority-change-3.5.md
```diff
@@ -0,0 +1,128 @@
...
+
+given userComponent: UserComponent = UserComponent()
+
+@main def run = printComponent 
+// Scala < 3.5: prints "user-defined"
+// Scala 3.5: prints "library-defined"
+```
+
+To detect such "silent" changes, we recommend compiling under Scala
+3.5 with the `-source:3.6-migration` flag:
```
WojciechMazur, line 86: I don't think we need to use `-source:3.6-migration` in 3.5 - the warnings would pop up automatically. 
However, we can point out that in Scala 3.6+ users can use `-source:3.5` to see the same warnings about changed priority in the future. 
Also combination of Scala 3.5 and `-source:3.6` might be useful to confirm that all implicit search results would not be ambiguous in the future, or to test application at runtime with the new rules being active.

bracevac: Resolved by [67fc9a0](https://github.com/scala/scala-lang/pull/1675/commits/67fc9a0abf514d44c5ed55b2c1ddf87a42c120c8)

_posts/2024-08-01-given-priority-change-3.5.md
```diff
@@ -0,0 +1,221 @@
...
+    override def msg = "user-defined"
+
+given userComponent: UserComponent = UserComponent()
+
+@main def run = printComponent 
+// Scala < 3.5: prints "user-defined"
+// Scala 3.5: prints "library-defined"
+```
+
+Scala 3.5 will automatically issue
```
KristianAN, line 91: Is it possible to turn off these warnings with a flag or is the only way through @nowarm or -source:3.6?

KristianAN: I ask because this breaks builds with fatal warnings.

bracevac: There is currently no flag to suppress those specific warnings and you'd have to use `@noWarn` or `-source:3.6`. As you point out. I've added that to the blog post.

_posts/2024-08-01-given-priority-change-3.5.md
```diff
@@ -0,0 +1,259 @@
...
+To assist with the migration to explicit `using` clauses, Scala 3.5
+provides an error message and offers automatic rewrites:
+
+```scala
+-- Error: bounds.scala:10:2 ----------------------------------------------
+10 |  f(eq, ord)  // error
+   |  ^
+   |Context bounds will map to context parameters.
+   |A `using` clause is needed to pass explicit arguments to them.
+   |This code can be rewritten automatically under -rewrite -source 3.4-migration.
```
bishabosha, line 258: does it actually print `3.4-migration` in the error? that seems surprising

bracevac: Yes, and it'll do the right thing.

bracevac:
Making this a draft for now, as we decided to delay the rollout of the new prioritization scheme.

_posts/2024-08-01-given-priority-change-3.5.md
```diff
@@ -0,0 +1,259 @@
+---
+layout: blog-detail
+post-type: blog
+by: Oliver Bračevac, EPFL
+title: "Changes to Givens in Scala 3.5"
+---
+
+## New Prioritization of Givens
+
+Starting with Scala 3.5, the prioritization of givens has been
```
lenguyenthanh, line 10: I think the correct version for prioritization of givens changes is 3.6, in 3.5 We just warming it up with warning.

bracevac: We were considering switching it on for 3.5 already. In the end, we decided against it to give people more time to test it.

lrytz:
It's good that the motivating example is something relevant in practice. However, starting the blog with Functor/Monad/Traverse could be overwhelming to Scala programmers who aren't using pure FP style / libraries. When they come across the new warning they would hopefully end up at the blog post. It might be good to put the simpler example (`LibComponent`) first.

_posts/2024-08-19-given-priority-change-3.7.md
```diff
@@ -0,0 +1,232 @@
...
+    // Scala 3.7: prints "library-defined"
+```
+
+What happened? In Scala 3.6 and earlier, the compiler prioritizes the
+`give`n with the _most specific_ compatible subtype
+(`userComponent`). However, in Scala 3.7, it selects the value with the
+_most general_ subtype instead (`libComponent`).
+
+
+### Motivation: Better Handling of Inheritance Triangles & Typeclasses
```
WojciechMazur, line 52: We might switch the order of `Motivation` and `New Prioritization of Givens`. 
By doing so we're starting with the overall language improvement (positive vibes), then pointing out the possible drawbacks (negativity), followed by migration guides (neutral, slightly positive).
So the overall structure might look like: 
```
# New Prioritization of Givens in Scala 3.7 (intro)

Scala 3.7 will introduce changes to how `givens` are resolved, which can
affect program behaviour when multiple givens are present. 
The goal of this change is to make a givens resolution more predictable, however, it can lead to problems when migrating to Scala 3.7 or later. 
In this article we'd like to explain the motivation, possible issues and migration guides to prepare developers for the upcoming changes.
...

### Motivation 
Why change the priority to the `given` .... 

### Source incompatibility of the new givens prioritization 
Unfortunately, this change might affect source compatibility of some Scala codebases. Let's consider a library that provides a default
... <here example from the 1st paragraph>

```

WojciechMazur, line 52: In most of [Scala documentation](https://docs.scala-lang.org/scala3/book/ca-type-classes.html) we're using `Type Class` in opposition to `Typeclass` that [seems to be used by Haskell community](https://serokell.io/blog/haskell-typeclasses)
```suggestion
### Motivation: Better Handling of Inheritance Triangles & Type classes
```
There's more of typeclass usages in the text

_posts/2024-08-19-given-priority-change-3.7.md
```diff
@@ -0,0 +1,232 @@
...
+**Scala 3.6**
+
+In Scala 3.6, these warnings will be on by default.
+
+**Scala 3.7**
+
+Scala 3.7 will finalize the transition, making the new given
+prioritization the standard behavior.
+
+#### Suppressing Warnings
```
WojciechMazur, line 142: We can consider adding a paragraph here that suppressing warnings is only a temporary solution. Eg. 

```
Consider that suppressing warnings should become only a temporary workaround. 
They can be used to resolve rare false-positive warnings produced by the compiler. 
However, they're only delaying the time when changes in either your codebase or libraries used by your project are required. Suppressing warnings can lead to unexpected problems when upgrading to the future versions of the Scala compiler.
```

WojciechMazur (commented):
In general looks good to me. I've found some typos and have idea how we can slightly change structure of the text to make it a bit more positive upfront

_posts/2024-08-19-given-priority-change-3.7.md
lenguyenthanh: `Type class` is also the correct terminology here as in [typeclassopedia](https://wiki.haskell.org/Typeclassopedia):


> The original spelling of “type class” is with two words, as evidenced by, for example, the [Haskell 2010 Language Report](http://www.haskell.org/onlinereport/haskell2010/), early papers on type classes like [Type classes in Haskell](https://dl.acm.org/doi/pdf/10.1145/227699.227700) and [Type classes: an exploration of the design space](https://citeseer.ist.psu.edu/viewdoc/download?doi=10.1.1.1085.8703&rep=rep1&type=pdf), and [Hudak et al.’s history of Haskell](http://citeseer.ist.psu.edu/viewdoc/summary?doi=10.1.1.168.4008). However, as often happens with two-word phrases that see a lot of use, it has started to show up as one word (“typeclass”) or, rarely, hyphenated (“type-class”).

SethTisue: fwiw (and at the risk of igniting a full-on holy war 😬), in Scala contexts I think "typeclass" is preferable because typeclasses aren't classes. Haskell doesn't have classes so there is no confusion in that context but in ours there is

bracevac: If [my boss](https://docs.scala-lang.net/sips/sips/typeclasses-syntax.html) uses ‘typeclass’, who am I to split words? 😄

_posts/2024-08-19-given-priority-change-3.7.md
bracevac: Done!

_posts/2024-08-19-given-priority-change-3.7.md
bracevac: Done!

bracevac: I don't feel too strongly about it, though I tend to agree with @SethTisue here. But if there's a majority that wants to have it changed to "Type Classes", I'll bend the knee.

WojciechMazur:
@SethTisue Seems like this one is ready to go. I don't have write access to proceed with merge. Can you please take care of it?

PR #1681, opened 2024-08-22
feat: use `<time>` tag

scarf005:
## Summary

applied [`<time>` tag](https://developer.mozilla.org/ko/docs/Web/HTML/Element/time) to following places:

| Blog | Blog posts |
| :-: | :-: |
| image | image |

- using [`<time>` tag](https://developer.mozilla.org/ko/docs/Web/HTML/Element/time) gives semantic meaning to time and helps accessibility for screen readers.
- the usage of tag doesn't affect styling.
- `blog.html` doesn't seem to be used anywhere else but included for completeness' sake.

### DIFF
--- a/_includes/blog-list.html
+++ b/_includes/blog-list.html
@@ -22,7 +22,7 @@
 								<ul class="tag-list">
 									{% for tag in post.tags %}
 										<li class="tag-item">{{tag}}</li>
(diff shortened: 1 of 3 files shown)


SethTisue:
Thank you!

PR #1689, opened 2024-09-03
disable dead links to scalamacros.org

SethTisue:
it appears someone bought the scalamacros.org domain and replaced it with some AI-generated junk... looks like a SEO scheme

doing the http->hxxp thing is kinda gross but the links are all in really ancient posts so I don't feel like putting more effort into e.g. finding archive.org links

as noticed by the link checker at e.g. https://github.com/scala/scala-lang/actions/runs/10674385404/job/29584796119


sjrd (approved):
I did not know about the `hxxp` convention. TIL.

Philippus (approved):

PR #1726, opened 2024-12-13
Scala 3.6.3 release announcement

WojciechMazur:


SethTisue:
@WojciechMazur is this supposed to be publicized on X, Bluesky, etc?

WojciechMazur:
> @WojciechMazur is this intended to be publicized on X, Bluesky, etc?

I think that previously the goal was to reserve these for minor releases of Scala 3 only , we might consider it for LTS though.

SethTisue:
okay. I'm not really sure what I think. part of me is eager to get the word out, but I can also understand not wanting to bother people too often

Issue #1738, opened 2025-01-23
Past events page should be reverse-chronological


SethTisue:
https://www.scala-lang.org/pastevents/

instead of 2010 at the top, 2024 should be at the top. this is a suggestion from @darjutak that I agree with. the past-events page isn't all that prominent of course, but it's going to get a visibility bump when we link to the events page from a forthcoming blog post

@Philippus you've given that page some love over the years, is this a task that would interest you? if not, I can handle it

Philippus:
Makes sense, I can have a look.

PR #1741, opened 2025-01-30
Reverse past events

Philippus:
Applying this PR reverses the events on the past events page. Fixes https://github.com/scala/scala-lang/issues/1738. I couldn't find a way to programmatically apply the "reversed" tag , that would be neater.


SethTisue:
looks to be working well at https://www.scala-lang.org/pastevents/ (and https://www.scala-lang.org/events/)

thank you, much appreciated by Darja and myself

PR #1769, opened 2025-03-03
Blog: 'Announcements' -> 'Releases'

SethTisue:
Screenshot 2025-03-03 at 2 51 21 PM

some blog posts were misfiled as release announcements so I recategorized them and added redirects


SethTisue:
One thing that I didn't clean up in this round is that it seems random whether things are in `/_posts` or `/blog/_posts`. In a general way it seems the former is used for release announcements and the latter for everything else, but there are posts in the "wrong" directory in both directions.

It isn't clear to me that it even matters which directory something goes in. If so many posts are in the "wrong" directory but we've never noticed, then maybe it doesn't matter and only the `category:` declarations matter?

SethTisue:
Another thing I didn't clean up except in a  couple places is that many files declare `post-type: announcement` or `post-type: blog`. As far as I can tell, this does nothing and only `category:` matters.

PR #1775, opened 2025-03-09
Scala 2 download pages: link to version-specific notes

SethTisue:
we've had this problem for yearrrrrrrrrs and I've always meant to fix it, and, well, today's the day

also remove the unneeded links to the archive of old release blog posts. it's distracting noise. the release notes have been on GitHub at least since 2.12.0 and we just don't care whether older versions are now a bit harder to find

### DIFF
--- a/_includes/downloads-scala2.html
+++ b/_includes/downloads-scala2.html
@@ -1,8 +1,7 @@
 <p>Released <b>{{page.release_date}}</b>. See <a href="{{ site.baseurl }}/download/all.html">all releases</a>.</p>
 <h3>Release Notes</h3>
-For a summary of important changes, see the <a href="https://github.com/scala/scala/releases">GitHub release notes</a>.
-<br/><small>(Or consult our archive of <a href="{{ site.baseurl }}/blog/releases/">older release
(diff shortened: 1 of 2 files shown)


SethTisue:
I built the site locally to verify that e.g. `/download/2.13.16.html` now links to https://github.com/scala/scala/releases/tag/v2.13.16 specifically instead of the more general https://github.com/scala/scala/releases

SethTisue:
@WojciechMazur I'm sort of guessing you'll be okay with the Scala 3 part of this, but. double-checking. The situation is a bit different in Scala 3 since (iirc) you have content in the release announcements on scala-lang.org that isn't in the GitHub release notes?

WojciechMazur:
Looks good to me. For the Scala 3 releases, especially new minors and it's the longer blogposts, we can add link in the GH changelog. In case of patches the scala-lang announcement  can typically be the same as GH changelog, in such case there would be no need to link between them, that's the case for the last one: https://scala-lang.org/news/3.6.4/ - https://github.com/scala/scala3/releases/tag/3.6.4

PR #1782, opened 2025-03-24
Update 2025-03-24-evolving-scala.md

gdemelo:
Fixing broken link for named tuples.


SethTisue:
thank you!

@odersky is this the best link? if not we can change it again

odersky:
I think it should be the reference page, not the SIP:

https://dotty.epfl.ch/docs/reference/other-new-features/named-tuples.html

The link was broken since the feature moved from experimental to standard in the last week when 3.7 was released.

SethTisue:
fixed: 1b88de709533062e431e015b6e8a72e6a2f63a6e

PR #1808, opened 2025-06-06
Update index.md (removed a link)

ColourGrey:
Removed the reference to scala-android.org because that site seems to have been hacked (just look at the Indonesian spam inserted on https://www.scala-android.org/dependencies).

Also, there is a huge difference between how the site looks today and how it looked in the past (see https://web.archive.org/web/20160706125519/http://scala-android.org/).


SethTisue:
thank you!

PR #1823, opened 2025-07-23
Update coursier link for Mac aarch64

### LINKED ISSUE #1821: scala instructions shouldn't point to ` VirtusLab/coursier-m1`
> (also, fair point that Scala install instructions should no longer be pointing here) 

 _Originally posted by @Gedochao in [VirtusLab/coursier-m1#156](https://github.com/VirtusLab/coursier-m1/issues/156#issuecomment-3061123029)_

https://github.com/scala/scala-lang/blob/9bfdd99f753bdae5e17b7fe7f97904db00ca9265/_data/setup-scala.yml#L2-L4

- #1814

Gedochao:
Fixes #1821


_data/setup-scala.yml
```diff
@@ -1,6 +1,6 @@
 linux-x86-64: curl -fL https://github.com/coursier/coursier/releases/latest/download/cs-x86_64-pc-linux.gz | gzip -d > cs && chmod +x cs && ./cs setup
 linux-arm64: curl -fL https://github.com/VirtusLab/coursier-m1/releases/latest/download/cs-aarch64-pc-linux.gz | gzip -d > cs && chmod +x cs && ./cs setup
```
Gedochao, line 2:
```suggestion
linux-arm64: curl -fL https://github.com/coursier/launchers/raw/master/cs-aarch64-pc-linux.gz | gzip -d > cs && chmod +x cs && ./cs setup
```
Now, we could do this, which at the point of typing this comment points to coursier 2.1.25-M16... as the new Linux aarch64 launchers are still locked behind milestone versions of 2.1.25, so they aren't available behind `latest`... but perhaps it's better to wait for stable?

`VirtusLab/coursier-m1` is otherwise unmaintained.

Either way, I opt to leave it like this for the time being, perhaps until 2.1.25 is fully out?
cc @alexarchambault

PR #1839, opened 2025-09-10
Remove Scala Days 2025 sponsors from the frontpage

anatoliykmetyuk:
The infrastructure to display sponsors is not deleted, as probably we'll need it again next year for the Scala Days 2026.


Philippus (approved):

_layouts/frontpage.html
```diff
@@ -16,7 +16,6 @@
         </p>
         <span class="hide"><i class="fa fa-close"></i></span>
       </div>
-      {% include scala-days-sponsors.html %}
```
sjrd, line 19: Leave a comment that this is where that include should be reintroduced in the future?

Issue #1854, opened 2025-11-05
Scastie snippets on homepage are missing


bishabosha:
The snippets in the "carousel" on the homepage are missing now in Scastie.

e.g.
- https://scastie.scala-lang.org/S0bkCiXkQiuOMXnlhWn46g
- https://scastie.scala-lang.org/jzYU3JWJTZW6731MigOqlA
- https://scastie.scala-lang.org/CQ70LFpbRJuWEDfGxbuYeQ
- https://scastie.scala-lang.org/VcczUlAVT8mHTl7tpGmiJg

<img width="621" height="382" alt="Image" src="https://github.com/user-attachments/assets/2fa79229-de07-419c-9b74-4ce66d98c0f0" />
<img width="832" height="253" alt="Image" src="https://github.com/user-attachments/assets/79fbcbef-7c70-499e-b2eb-903155bdd677" />

I don't remember if there are any backups of the original code so this could be unfortunate, or perhaps we can recreate something similar.

rochala:
We've changed policy with Scastie and all snippets which are anonymous were deleted. From now on they will persist for 30 days. If you want to have them assigned to organization there is an incoming PR

PR #1855, opened 2025-11-05
Update website links

rochala:
Fixes https://github.com/scala/scala-lang/issues/1854

This will be changed later on to links to scalacenter organisation that will not require updating the website, rather you will be able to just update the snippet on the scastie website itself and it will update here.


tgodzik (approved):
LGTM! Thanks!

PR #1864, opened 2025-11-26
chore: Switch to Inconsolata just in case

tgodzik:
Tried to make it as close as possible, but we don't have italic for Inconsolata


SethTisue (approved):

SethTisue:
good to have this oversight corrected

tgodzik:
> will docs.scala-lang need a corresponding PR?

Looks like it. I will take a look.

darjutak (approved):

SethTisue:
fwiw, I did not have Inconsolata on my system (MacOS, Safari), so with this merged, it's falling back to what looks like Courier

we could decide that's acceptable, or we could add a fallback to e.g. Monaco :shrug:

SethTisue:
here's the before and after in my browser

Screenshot 2025-11-26 at 8 33 02 AM
Screenshot 2025-11-26 at 8 39 09 AM

SethTisue:
https://scala-lang.org/resources/css/monospace.css now says Inconsolata, and I now see the right thing in Safari in both scala-lang and docs.scala-lang, so I think we're done unless someone reports a new problem

PR #1868, opened 2026-01-15
Add Scala 3.8.0 & 3.8.1 release blogpost

WojciechMazur:
Draft blogpost for release 3.8.0 

Planned publish date:  January 16 2026

### DIFF
--- a/_posts/2026-01-22-release-notes-3.8.md
+++ b/_posts/2026-01-22-release-notes-3.8.md
@@ -1,6 +1,6 @@
 ---
 category: release
-permalink: /news/3.8.0/
+permalink: /news/3.8/
 title: "Scala 3.8 released!"
 by: Wojciech Mazur, VirtusLab
 ---


_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+val ys = Array(3, 4)
+
+val out = concat(xs, ys)   // `ys` is implicitly converted because expected type is `into[IterableOnce[Int]]`
+```
+
+`into` is currently a **preview feature** and is planned to be stabilised in the minor version.
+It requires compilation with `-preview` compiler flag and can be activated using `import scala.language.experimental.into`.
+If you previously used experimental annotation-based preview implementation be aware of source incompatibilites when upgrading to Scala 3.8.
+
+More information about `into` modifiers can be found in [the dedicated reference](https://docs.scala-lang.org/scala3/reference/preview/into.html).
```
WojciechMazur: Reference docs are not yet updated to 3.8 changed. Link needs to be confirmed before publish

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
+---
+category: release
+permalink: /news/3.8.0/
+title: "Scala 3.8.0 released!"
+by: Wojciech Mazur, VirtusLab
+---
+
+![Scala 3.8]({{ site.baseurl }}/resources/img/scala-3.8-launch.png)
+
+We're thrilled to announce the release of Scala 3.8.0 — a milestone release that modernizes the Scala ecosystem and paves the way for Scala 3.9 LTS. This release brings the standard library into the Scala 3 era with native compilation, stabilizes highly-anticipated features like **Better Fors** (SIP-62) and **`runtimeChecked`** (SIP-57), and introduces groundbreaking experimental capabilities including **flexible varargs** and **strict equality pattern matching**.
```
SethTisue: "native compilation" risks being misinterpreted as being about Scala Native. I think the headline you chose below, "Compiled with Scala 3", is much clearer. "This release introduces a standard library compiled by Scala 3 itself", perhaps?

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+
+# What's new in Scala 3.8?
+
+## JDK 17 is now required
+
+> **Important:** This release requires **JDK 17 or later** for compilation and execution. If your project runs on an older JDK, you must upgrade before migrating to **Scala 3.8**.
+
+The plans for this change were [announced in early 2025](https://www.scala-lang.org/news/next-scala-lts-jdk.html) by the Scala team and are required to support future JDK 26+ which restricts access to `sun.misc.Unsafe` functionalities used by lazy-vals implementation. For most end users this might be just a small adjustment, but this change needs to be considered by the library authors and correctly communicated with its users. If your project still runs on an older JDK, you must first upgrade newer version of JDK before upgrading to Scala 3.8. Upgrading earlier will allow you to benefit from new language features and improved performance on the modern JVM.
```
SethTisue:
```suggestion
The plans for this change were [announced in early 2025](https://www.scala-lang.org/news/next-scala-lts-jdk.html) by the Scala team and are required to support future JDK 26+, which restricts access to `sun.misc.Unsafe` functionalities used by our existing lazy-vals implementation. For most end users this might be just a small adjustment, but this change needs to be considered by the library authors and communicated to their users. If your project still runs on an older JDK, you must first upgrade to at least 17 before upgrading to Scala 3.8. Upgrading will allow you to benefit from new language features and improved performance on the modern JVM.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+
+The long‑term support version Scala 3.3 will continue to produce JDK 8 compatible bytecode, but all future feature releases (including the 3.9 LTS) will target JDK 17.
```
SethTisue:
```suggestion
The LTS (long‑term support) version Scala 3.3 will continue to produce JDK 8 compatible bytecode, but all future Scala releases (including the 3.9 LTS) will require JDK 17 or higher.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+For library authors who need to keep compatibility with older JVM versions we recommend using upcoming Scala 3.3.8 LTS when it's released in the next months. That version will provide `-Yfuture-lazy-vals` flag which activates new implementation of Lazy Vals, yet, it still requires upgrade to at least JDK 9.
```
SethTisue:
```suggestion
For library authors who need to keep compatibility with at least some older JVM versions, we recommend using upcoming Scala 3.3.8 LTS when it's released in the next months. That version will provide a `-Yfuture-lazy-vals` flag which activates new implementation of lazy vals, which still requires upgrading to at least JDK 9.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+Compilation using Scala 3 allows to benefit from two Scala 3 optional features:
+
+1. **Null‑safety** - When compiled with the `-Yexplicit-nulls` flag, all reference types in the library become non‑nullable unless they are annotated with `| Null`. Although the default remains unchanged, library maintainers have annotated many APIs with explicit null‑return types. When you enable explicit nulls in your project the type checker will warn if you forget to handle a possible `Null`.
+
+2. **Capture checking** - The library code has been adjusted to interact properly with the experimental capture‑checking system. When you enable capture checking in your project (`-language:experimental.captureChecking`) the type checker tracks references to capabilities and ensures you do not capture local resources in closures.
+
+> **Note:** Both null‑safety and capture‑checking annotations remain opt‑in features. Your existing code will behave exactly as before unless you enable the appropriate language flag. The goal is to provide a smooth migration path towards safer Scala in the next LTS release.
```
SethTisue:
```suggestion
> **Note:** Both null‑safety and capture‑checking annotations remain experimental, opt‑in features. Your existing code will behave exactly as before unless you enable the appropriate language flag. The goal is to provide a smooth migration path towards safer Scala in the next LTS release.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+```scala
+object Array:
+  def empty[T: ClassTag]: Array[T]
+
+val fails = Array.empty(reflect.ClassTag.Int) // error: No ClassTag available for Any
+val works = Array.empty(using reflect.ClassTag.Int)
+val recommended = Array.empty[Int]
+```
+
+Most of these errors can be automatically fixed by compiler when using **Scala 3.7.4** with `-source:3.7-migration -rewrite` options, before upgrading to **Scala 3.8**.
```
SethTisue:
```suggestion
Most such errors can be automatically fixed by the compiler when using **Scala 3.7.4** with `-source:3.7-migration -rewrite` options, before upgrading to **Scala 3.8**.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+
```
SethTisue: Looking ahead to Scala’s future, compiling the standard library with Scala 3 itself paves the way for Scala 3 to eventually have its own standard library, free from the constraints of compatibility with Scala 2. However, this process will not actually begin until Scala 3.10.

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+### Pure functions and capture checking
+
+The `scala.caps.Pure` capability is no longer experimental and is used by the capture checker to model pure functions.
+Experimental separation checking ensures that closures do not capture resources, providing stronger guarantees about side‑effect isolation.
+To experiment with capture checking, import `scala.language.experimental.captureChecking` and annotate your functions with the appropriate capabilities.
```
SethTisue:
```suggestion
To experiment with capture checking, import `scala.language.experimental.captureChecking` and annotate your functions with the appropriate capabilities. Adding these annotations does not require enabling experimental features and it will not affect users who are not using capture checking.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+  sizes = List(1, 2, 3)
+  size <- sizes
+yield size * 2
+```
+
+* **Avoiding unnecessary `map` calls** - When the last generator yields a value of the same type as its input, the for‑comprehension now emits direct code without an intermediate `map` call. This simplifies the generated bytecode and avoids constructing unnecessary tuples.
+
+### [SIP-57: Replace non-sensical `@unchecked` annotations](https://docs.scala-lang.org/sips/57.html)
+
+The `runtimeChecked` extension is now a **standard feature**. It enables you to opt-out of certain static checks and defer them to runtime without resorting to the older `: @unchecked` type ascription.
```
SethTisue:
```suggestion
The `runtimeChecked` extension is now a **standard feature**. It enables you to opt-out of certain static checks and defer them to runtime without resorting to the older, syntactically awkward `: @unchecked` type ascription.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+def portFromConfig(name: String): Option[Int] = ???
+
+val Some(httpPort) = portFromConfig("http") // warning: right hand side expression is Option, not Some
+
+portFromConfig("https") match 
+    case Some(port) => println(s"HTTPS port: $port")  
+    // warning: match may not be exhaustive.
+```
+
+Compiler detects and warns that in some cases your program might fail at runtime with exception. The previous `: @unchecked` annotation could have been used to suppress these warnings, but its syntax made it difficult to apply.
```
SethTisue:
```suggestion
The compiler detects and warns that in some cases your program throw an exception at runtime. The previous `: @unchecked` annotation could have been used to suppress these warnings, but its syntax made it difficult to apply.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+    .map(_ + 1)
+    .runtimeChecked
+    .match 
+        case Some(port) => port
+```
+
+`runtimeChecked` only suppresses checks that can be performed at runtime. If you write an obviously wrong pattern (for instance, matching an `Int` against a `String`) the compiler still emits an error.
+
+## New preview features
+
```
SethTisue: Preview features are only enabled if the `-preview` flag is used. Features that have reached the preview stage are always eventually stabilized, but may change incompatibly before that time.

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+### [SIP-71: Allow fully implicit conversions in Scala 3 with `into`](https://docs.scala-lang.org/sips/71.html)
+
+Scala 3 normally requires `import scala.language.implicitConversions` anywhere you *use* `scala.Conversion` instances implicitly.
+The new `into` feature provides two opt-in ways to allow implicit conversions **without** that import:
```
SethTisue:
```suggestion
The new `into` keyword provides two opt-in ways to allow implicit conversions **without** that import:
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+
+`into` is currently a **preview feature** and is planned to be stabilised in the minor version.
+It requires compilation with `-preview` compiler flag and can be activated using `import scala.language.experimental.into`.
+If you previously used experimental annotation-based preview implementation be aware of source incompatibilites when upgrading to Scala 3.8.
+
+More information about `into` modifiers can be found in [the dedicated reference](https://docs.scala-lang.org/scala3/reference/preview/into.html).
+
+## New experimental features
+
+> The following features are provisional and subject to change. Do not rely on them in production code yet.
```
SethTisue:
```suggestion
> The following features are provisional and subject to change or outright removal. Do not rely on them in production code yet.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+      case _                             => "unsupported"
+    case _                               => "default"
+```
+
+* The sub-cases are tested only if the outer case matches (here: `Some(Document(_, version))`).
+* The sub-match scrutinee (`version`) can refer to variables bound by the outer pattern (`doc`).
+* Sub-matches do not have to be exhaustive. If none of the sub-cases match, the whole `case ... if <submatch>` is treated as not matched, and the outer match continues with the next case.
+* Sub-matches can be nested (sub-sub-matches), and you can interleave additional boolean guards between them when needed.
+
+This experimental feature has not yet entered the SIP, but we'd like to hear your feedback about this proposal.
```
SethTisue:
```suggestion
This experimental feature has not yet been considered by the SIP committee, but we'd like to hear your feedback about this proposal.
```

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+* **Metals**: Recommended: [Metals 1.6.4](https://scalameta.org/metals/blog/2025/11/25/osmium)
+* **scalafmt**: Minimum [scalafmt 3.10.1](https://github.com/scalameta/scalafmt/releases/tag/v3.10.1)
+* **Scalameta**: Minimum: [scalameta 4.14.1](https://github.com/scalameta/scalameta/releases/tag/v4.14.1) (used by Metals, scalafmt, scalafix, and more)
+
+## What's next?
+
+The compiler codebase itself has switched into a feature freeze mode in preparation for Scala 3.9 LTS - no new language features would be accepted (development and improvement of existing features is still allowed however). It allows us to focus on fixing existing bugs in both tooling and compiler to ensure a smooth transition into the Scala 3.9 LTS. The freeze would be lifted with the start of the Scala 3.10 development cycle.
+
+Effectively, **Scala 3.9 LTS would contain the same feature set as Scala 3.8** (we reserve the right to promote preview features into stable ones, but no new experimental or preview features would be introduced). As a result, any codebase using Scala 3.8 should not require any additional changes to use Scala 3.9 LTS.
```
SethTisue:
```suggestion
Effectively, **Scala 3.9 LTS would contain the same feature set as Scala 3.8**. We may promote preview features to stable, but no new experimental or preview features will be introduced. As a result, any codebase using Scala 3.8 should not require any additional changes to use Scala 3.9 LTS.
```

_posts/2026-01-16-release-notes-3.8.0.md
sjrd:
```suggestion
> **Note:** Both null‑safety and capture‑checking annotations remain opt‑in features. Your existing code will behave exactly as before unless you enable the appropriate language flag. The goal is to provide a smooth migration path towards safer Scala in the future.
```
I read that sentence as: null safety and capture will be made stable in the next LTS release. 😱

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+
+val Some(httpPort) = portFromConfig("http") // warning: right hand side expression is Option, not Some
+
+portFromConfig("https") match 
+    case Some(port) => println(s"HTTPS port: $port")  
+    // warning: match may not be exhaustive.
+```
+
+Compiler detects and warns that in some cases your program might fail at runtime with exception. The previous `: @unchecked` annotation could have been used to suppress these warnings, but its syntax made it difficult to apply.
+The new `runtimeChecked` clearly marks that given operation might fail, focusing your attention during code reviews. What's more, it's perfect for use in chained calls!
```
sjrd: This is backwards. `@unchecked` should give you the creeps during code review, like an `asInstanceOf`. `runtimeChecked` isn't any different than `.head` or `.get`.

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+`runtimeChecked` only suppresses checks that can be performed at runtime. If you write an obviously wrong pattern (for instance, matching an `Int` against a `String`) the compiler still emits an error.
+
+## New preview features
+
+### [SIP-71: Allow fully implicit conversions in Scala 3 with `into`](https://docs.scala-lang.org/sips/71.html)
+
+Scala 3 normally requires `import scala.language.implicitConversions` anywhere you *use* `scala.Conversion` instances implicitly.
+The new `into` feature provides two opt-in ways to allow implicit conversions **without** that import:
+
+* **`into[T]` as a type constructor**: marks specific parameter (or expected) types as “conversion allowed”, giving fine-grained control (only where you wrap with `into[...]`)
```
sjrd:
```suggestion
* **`into[T]` as a type constructor**: marks specific parameter types as “conversion allowed”, giving fine-grained control (only where you wrap with `into[...]`)
```
The "clarification" in parentheses is more obscure. Moreover, it is grammatically awkward because it likens an adjective (expected) with a noun (parameter), which adds to the cognitive load to make sense of the sentence.

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+### Match expressions with sub-cases [#23786](https://github.com/scala/scala3/pull/23786)
+
+This change adds an experimental extension to pattern matching: a `case` in a `match` can be followed by a **nested “sub-match”** introduced by the `if` keyword.
+It’s a small syntactic feature, but it improves readability in cases where you want to:
+
+* first match an “outer shape” (e.g., `Some(x)`), and then
+* immediately refine the result by matching on a field derived from the outer bindings (e.g., `x.version`),
+* without introducing an extra level of indentation or a `case ... => expr match ...` block.
```
sjrd: It's much more than that. If the nested pattern does not match, we still proceed to next enclosing `case`. This is fundamentally different than putting the sub-case within another level of nesting.

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,304 @@
...
+* **JDK 26 support** - Scala can now emit and consume JDK 26 bytecode. As mentioned [earlier](#jdk17is-now-required) internal implementation of the lazy vals has been adjusted to replace deprecated `sun.misc.Unsafe` with `VarHandles`, which starting with JDK 24 started to emit runtime warnings and is going to throw an exception in JDK 26. This change makes the new published code compatible with JDK 26, but mentioned problems might still be activated by previously published libraries or their transitive dependencies. We're working on creating tools to modify existing lazy-vals related code to make them future-proof and ease the migration to the upcoming JDK versions. For more details see the technical discussion in [latest Jakub Kozłowski's podcast with Łukasz Biały](https://www.youtube.com/watch?v=K_omndY1ifI) - VirtusLab's Scala Advocate, who prototyped this idea.
```
sjrd: There are very long sentences here. Break up sentences with full periods as much as possible.

_posts/2026-01-16-release-notes-3.8.0.md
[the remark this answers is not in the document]
bishabosha: Also maybe not the capabilities word

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,310 @@
...
+* **Metals**: Recommended: [Metals 1.6.4](https://scalameta.org/metals/blog/2025/11/25/osmium)
+* **scalafmt**: Minimum [scalafmt 3.10.1](https://github.com/scalameta/scalafmt/releases/tag/v3.10.1)
+* **Scalameta**: Minimum: [scalameta 4.14.1](https://github.com/scalameta/scalameta/releases/tag/v4.14.1) (used by Metals, scalafmt, scalafix, and more)
+
+## What's next?
+
+The language is now in feature freeze mode in preparation for Scala 3.9 LTS. No new language features will be accepted for 3.9. Development and improvement of existing features is still allowed. This allows us to focus on fixing existing bugs in both tooling and compiler to ensure a smooth transition into the Scala 3.9 LTS era. The freeze will be lifted with the start of the Scala 3.10 development cycle.
+
+Effectively, **Scala 3.9 LTS would contain the same feature set as Scala 3.8**. We may promote preview features to stable, but no new experimental or preview features will be introduced. As a result, any codebase using Scala 3.8 should not require any additional changes to use Scala 3.9 LTS.
```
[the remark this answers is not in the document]
natsukagami: Yes, capture checking is not expected to be enabled by default for the near future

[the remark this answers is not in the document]
WojciechMazur: Yes, capture checking would still be disabled by default in the LTS 3.9, similarly to explicit-nulls - both of these are experimental and opt-in. There is no timeline yet regarding when these can be made final and stable

_posts/2026-01-16-release-notes-3.8.0.md
WojciechMazur: Link confirmed

_posts/2026-01-16-release-notes-3.8.0.md
```diff
@@ -0,0 +1,321 @@
...
+For library authors who need to keep compatibility with at least some older JVM versions, we recommend using upcoming Scala 3.3.8 LTS when it's released in the next months. That version will provide a `-Yfuture-lazy-vals` flag which activates new implementation of lazy vals, which still requires upgrading to at least JDK 9.
+
+## Standard Library changes
+
+### Compiled with Scala 3
+
+The Scala standard library has historically been compiled using Scala 2.13 and used by Scala 3 as-is, thanks to binary compatibility. In 3.8.0 the library is now compiled with Scala 3. The change itself has been verified to be binary compatible and should not cause problems when migrating to Scala 3.8.
+
+> **Source Incompatibility:** Context bounds in standard library classes such as `scala.reflect.ClassTag` used in `Array.empty[T]` method, are now desugared to `given` instead of `implicit`. This change requires `using` modifiers when supplying explicit parameters. This changed behaviour was active since Scala 3.6, but was inactive while Scala Standard Library was compiled using Scala 2.13.
```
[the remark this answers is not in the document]
SethTisue: um, no strong opinion either way... the issue occurs only at build time, is a false positive (less pernicious than a false negative!), and should be very googleable, so I'd probably consider it sufficient to have a closed "not a bug" issue in the MiMa repo, but if you decide it's worth mentioning here, I don't object

Gedochao (approved):

